PR #386, opened 2020-02-24
DCS-383 Converting 500 response into correct 400 

andrewrlee:
This occurs when the user is an RO with global search,  has no active caseloads (as most won't) and provides no offender numbers

### DIFF
--- a/src/main/java/net/syscon/elite/service/BookingService.java
+++ b/src/main/java/net/syscon/elite/service/BookingService.java
@@ -651,7 +651,7 @@ public class BookingService {
     }
 
     private Set<String> getCaseLoadIdForUserIfRequired() {
-        return isViewAllBookings() ? null : caseLoadService.getCaseLoadIdsForUser(authenticationFacade.getCurrentUsername(), false);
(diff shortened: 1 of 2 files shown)


src/main/java/net/syscon/elite/api/model/OffenderSentenceDetail.java
```diff
@@ -17,7 +17,7 @@
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @AllArgsConstructor
 @NoArgsConstructor
-@EqualsAndHashCode(callSuper = true)
+@EqualsAndHashCode(callSuper = false)
```
petergphillips: not sure about this change - won't it then mean two sentence details can be equal for different offenders?

petergphillips:
Would be good to have a resource test that mimics the RO user with no active caseloads scenario

src/main/java/net/syscon/elite/api/model/OffenderSentenceDetail.java
andrewrlee: I missed the Data annotation and assumed that there wasn't an equals/hashcode on the super class
I've removed this change as it wasn't needed any more.

PR #389, opened 2020-02-26
DT-576 initial JPA model to support creation of court events.

matt-crowson-moj:
So this is the initial entity that will be created as part of the PECS secure prison to court movements.  

The first stage will be to create a court event which is linked to a court case and offender booking.

### DIFF
--- a/src/main/java/net/syscon/elite/repository/jpa/model/CourtEvent.java
+++ b/src/main/java/net/syscon/elite/repository/jpa/model/CourtEvent.java
@@ -1,0 +1,84 @@
+package net.syscon.elite.repository.jpa.model;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.NoArgsConstructor;
+import org.hibernate.annotations.JoinColumnOrFormula;
+import org.hibernate.annotations.JoinColumnsOrFormulas;
+import org.hibernate.annotations.JoinFormula;
+import org.hibernate.annotations.NotFound;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+import javax.persistence.Table;
+import java.time.LocalDate;
+import java.time.LocalTime;
+
+import static net.syscon.elite.repository.jpa.model.EventStatus.EVENT_STS;
+import static net.syscon.elite.repository.jpa.model.EventType.EVENT_TYPE;
+import static org.hibernate.annotations.NotFoundAction.IGNORE;
+
+@Data
+@Entity
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+@EqualsAndHashCode(of = "id", callSuper = false)
+@Table(name = "COURT_EVENTS")
+public class CourtEvent extends AuditableEntity {
+
+    @Id
+    @Column(name = "EVENT_ID", nullable = false)
+    private Long id;
+
+    @ManyToOne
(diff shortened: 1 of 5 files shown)


src/main/java/net/syscon/elite/repository/jpa/model/CourtEvent.java
```diff
@@ -0,0 +1,84 @@
...
+    private EventStatus eventStatus;
+
+    @ManyToOne
+    @JoinColumn(name = "AGY_LOC_ID", nullable = false)
+    private AgencyLocation courtLocation;
+
+    @Column(name = "COMMENT_TEXT", length = 240)
+    private String commentText;
+
+    @Column(name = "NEXT_EVENT_REQUEST_FLAG", length = 1)
```
matt-crowson-moj: I realise length is used during DDL generation (which we are not doing as NOMIS already exists and is very unlikely to change).

However do we think this is useful (documentation) as it should reflect what is in NOMIS?  Also we can potentially look at applying this to our models. e.g. `@Size(max = 240, message = "Blah must be a maximum of 240 characters.")`

src/main/java/net/syscon/elite/repository/jpa/model/CourtEvent.java
```diff
@@ -0,0 +1,84 @@
...
+    private OffenderBooking offenderBooking;
+
+    @Column(name = "EVENT_DATE", nullable = false)
+    private LocalDate eventDate;
+
+    @Column(name = "START_TIME", nullable = false)
+    private LocalTime startTime;
+
+    @ManyToOne
+    @NotFound(action = IGNORE)
```
andymarke: Any reason why these will not be found? Does the schema not have a foreign key to the reference data?

matt-crowson-moj: Hmm well this was originally based on the Reference Code stuff worked on by @petergphillips so it might be a carry over which isn't relevant...

src/test/java/net/syscon/elite/repository/jpa/repository/CourtEventRepositoryTest.java
```diff
@@ -0,0 +1,103 @@
...
+                .startTime(startTime)
+                .build());
+
+        entityManager.flush();
+
+        var persistedCourtEvent = courtEventRepository.findById(999L).orElseThrow();
+
+        assertThat(persistedCourtEvent)
+                .extracting(
+                        CourtEvent::getId,
```
andymarke: Nice, didn't know you could do this "extracting" passing a method ref

src/test/java/net/syscon/elite/repository/jpa/repository/CourtEventRepositoryTest.java
```diff
@@ -0,0 +1,103 @@
...
+                .eventDate(eventDate)
+                .eventStatus(eventStatus)
+                .nextEventRequestFlag(nextEventRequestFlag)
+                .offenderBooking(offenderBooking)
+                .offenderCourtCase(offenderCourtCase)
+                .orderRequestedFlag(orderRequestedFlag)
+                .startTime(startTime)
+                .build());
+
+        entityManager.flush();
```
andymarke: double checking; this does do the actual insert right? When not in a transaction?

matt-crowson-moj: Forcing it to commit (flush) the changes to the DB.  It's a bit of a pain really.  You can also use:
```
        TestTransaction.flagForCommit();
        TestTransaction.end();
        TestTransaction.start();
```
Pick your poison really...

It would be nice if the Spring JPA test framework made this a bit easier (there may well be of course, I tend to think I know Spring then someone says 'but didn't you know you could do this?'

PR #393, opened 2020-02-28
GDPR-85 Improving deletion scripts

brightonsbox:
* Adding missing deletion of work_flows / work_flow_logs
 * Adding missing cases (delete by offender_book_id) for addresses /
 phones / internet_addresses
 * Anonymising gl_transactions logs rather than deleting (advice from Syscon)
 * Tidying up / reorganising


src/main/resources/db/migration/nomis/ddl/V1_206__GL_TRANSACTIONS.sql
```diff
@@ -8,7 +8,7 @@ CREATE TABLE GL_TRANSACTIONS (
     TXN_TYPE VARCHAR2(6) NOT NULL,
     TXN_POST_USAGE VARCHAR2(12) NOT NULL,
     CASELOAD_ID VARCHAR2(6) NOT NULL,
-    OFFENDER_ID NUMBER(10,0) NOT NULL,
+    OFFENDER_ID NUMBER(10,0),
```
brightonsbox: Accidentally added this constraint that doesn't exist in production - can set offender_id to null (on instruction from Syscon)

PR #394, opened 2020-02-28
DT-586 initial commit prep work for exposing court events.

matt-crowson-moj:
Simple refactoring to reduce amount of boiler plate transformation code in the Bookings service for court cases in preparation for further changes for court events.


src/main/java/net/syscon/elite/service/transformers/CourtCaseTransformer.java
```diff
@@ -0,0 +1,39 @@
...
+        return CourtCase.builder()
+                .id(courtCase.getId())
+                .caseSeq(courtCase.getCaseSeq())
+                .beginDate(courtCase.getBeginDate())
+                .agency(Agency.builder()
+                        .agencyId(courtCase.getAgencyLocation().getId())
+                        .agencyType(courtCase.getAgencyLocation().getType())
+                        .active(courtCase.getAgencyLocation().getActiveFlag().isActive())
+                        .description(courtCase.getAgencyLocation().getDescription())
+                        .build())
```
Mjwillis: Shouldn't Agency have its own transformer?

matt-crowson-moj: No keeping people happy here ;)

Indeed it should this is just the first.  There be one and also one for court events....

src/test/java/net/syscon/elite/service/transformers/AgencyTransformerTest.java
```diff
@@ -0,0 +1,74 @@
...
+    void transform_active_and_description_capitalisation() {
+        var agency = builder.activeFlag(ActiveFlag.Y).build();
+
+        assertThat(transform(agency))
+                .extracting(
+                        Agency::getAgencyId,
+                        Agency::getAgencyType,
+                        Agency::getDescription,
+                        Agency::isActive)
+                .containsOnly(
```
petergphillips: This doesn't look right.  You're just testing that the four values are in the list.  So the builder could do
```java
        return Agency.builder()
                .agencyId(agency.getType())
                .agencyType(agency.getId())
```
and the test would still succeed.  Since these are lombok data classes would be much better to just do 
```java
        assertThat(transform(agency)).isEqualTo(new Agency(
...
```
instead

matt-crowson-moj: done

PR #395, opened 2020-03-02
Dt 370 jwk server

mikehalmamoj:
Use the JWK server from NomisOauth2.
Switch to Spring Security 5.


build.gradle
```diff
@@ -6,7 +6,7 @@ import java.time.LocalDate
 plugins {
     id 'java'
     id 'groovy'
-    id 'org.springframework.boot' version '2.2.4.RELEASE' // remove pinning of tomcat version to 9.0.30 when next upgrading
+    id 'org.springframework.boot' version '2.2.4.RELEASE'
```
brightonsbox: Can this be 2.2.5.RELEASE now?

src/main/java/net/syscon/elite/security/AuthenticationFacade.java
```diff
@@ -48,12 +48,9 @@ public boolean isIdentifiedAuthentication() {
         final var auth = getAuthentication();
         return StringUtils.isNotBlank(MDC.get(PROXY_USER))
                 && Optional.ofNullable(auth).
-                filter(OAuth2Authentication.class::isInstance).
```
andymarke: why was all this replaced?

src/main/java/net/syscon/elite/web/config/AuthAwareAuthenticationConverter.java
```diff
@@ -1,43 +1,38 @@
...
-            return AuthorityUtils.commaSeparatedStringToAuthorityList((String) authorities);
-        }
-        if (authorities instanceof Collection) {
-            return AuthorityUtils.commaSeparatedStringToAuthorityList(StringUtils
-                    .collectionToCommaDelimitedString((Collection<?>) authorities));
+    @SuppressWarnings("ConstantConditions")
+    private Collection<GrantedAuthority> extractAuthorities(final Jwt jwt) {
+        final var authorities = new HashSet<>(this.jwtGrantedAuthoritiesConverter.convert(jwt));
+        if (jwt.getClaims().containsKey("authorities")) {
+            authorities.addAll(((Collection<String>) jwt.getClaims().get("authorities")).stream().map(SimpleGrantedAuthority::new).collect(Collectors.toSet()));
```
andymarke: an thing we can do about this nasty cast?

src/main/resources/application.yml
```diff
@@ -17,6 +17,15 @@ spring:
     serialization:
       WRITE_DATES_AS_TIMESTAMPS: false
 
+  security:
```
andymarke: I guess this is unrelated to the Spring security move but this is what you started of doing?

mikehalmamoj:
> Am I right in thinking if I was running this locally I would also need an Auth running locally as well?
Yes, to verify the signed JWT it gets the public key from Oauth.

src/main/java/net/syscon/elite/security/AuthenticationFacade.java
mikehalmamoj: OAuth2Authentication is a class from Spring Security Oauth and isn't available in Spring Security 5.  The only line of interest is the call to `isAuthenticated` and we still do that.

src/main/java/net/syscon/elite/web/config/AuthAwareAuthenticationConverter.java
mikehalmamoj: I could extend the Spring Jwt class to codify our "authorities" claim as a Collection<String>?  Which essentially just moves the cast...  What do you think?

src/main/resources/application.yml
mikehalmamoj: Yes - started off getting the signing key from the jwk-set-uri location (well, Spring does that for you).  But then added issuer and moved to Spring Security 5 after feedback.

PR #399, opened 2020-03-03
Nn 2445 appointments endpoint

JamieDraperUK:
Endpoint to return all appointments at an agency for a given date.


src/main/java/net/syscon/elite/repository/jpa/model/ScheduledEvent.java
```diff
@@ -0,0 +1,62 @@
...
+import static net.syscon.elite.repository.jpa.model.ScheduledEventType.EVENT_TYPE;
+import static org.hibernate.annotations.NotFoundAction.IGNORE;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+@Entity
+@Table(name = "OFFENDER_IND_SCHEDULES")
+public class ScheduledEvent {
```
JamieDraperUK: Tempted to remove **ScheduledEvent** and **ScheduledEventType** because these files aren't actually being used. Initially I went down this route but abandoned it for an simpler (subquery) approach given my requirements. It could be useful and it adds to the JPA entity collection? Happy to remove if not.

src/test/resources/application-test.properties
```diff
@@ -15,3 +15,6 @@ saved-request.cookie.name=savedrequest
 saved-request.cookie.expiry-time=0
 
 data.compliance.db.enable.parallel.hints=false
+
+spring.jpa.show-sql=true
```
JamieDraperUK: Is this useful to have in by default?

src/main/java/net/syscon/elite/repository/jpa/model/ScheduledEvent.java
```diff
@@ -0,0 +1,62 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import org.hibernate.annotations.JoinColumnOrFormula;
+import org.hibernate.annotations.JoinColumnsOrFormulas;
+import org.hibernate.annotations.JoinFormula;
+import org.hibernate.annotations.NotFound;
+
+import javax.persistence.*;
```
matt-crowson-moj: A very meh comment but according to the GDS guidelines we shouldn't be using wildcard imports...

https://gds-way.cloudapps.digital/manuals/programming-languages/java.html#code-formatting

src/test/resources/application-test.properties
petergphillips: yes

src/test/java/net/syscon/elite/service/BookingServiceTest.java
```diff
@@ -1,6 +1,7 @@
 package net.syscon.elite.service;
 
 import net.syscon.elite.api.model.*;
+import net.syscon.elite.api.model.ScheduledEvent;
```
petergphillips: random import?

PR #405, opened 2020-03-04
DT-606 telemetry client

mikehalmamoj:
Retrieve the public key to verify JWTs from the JWK server.


src/main/java/net/syscon/elite/web/config/JwkClient.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.stereotype.Component;
+
+import java.net.URL;
+import java.security.PublicKey;
+import java.util.HashMap;
+import java.util.Map;
+
+@Component
+@Slf4j
+public class JwkClient {
```
JamieDraperUK: Looks to be in a different coding style to a lot of the previous code. Mainly the lack of final, var and streams.

src/main/java/net/syscon/elite/web/config/JwkClient.java
```diff
@@ -0,0 +1,53 @@
...
+    private String jwkUrl;
+    private Map<String, PublicKey> keyIdToPublicKeys = null;
+
+    public JwkClient(@Value("${spring.security.oauth2.resourceserver.jwt.jwk-set-uri}") String jwkUrl) {
+        this.jwkUrl = jwkUrl;
+    }
+
+    public PublicKey getPublicKeyForKeyId(String keyId) {
+        if (keyIdToPublicKeys == null) {
+            getJwkSets();
```
JamieDraperUK: Would it be worth having this function what it needs?

PR #410, opened 2020-03-06
DT-609 stub jwk server

mikehalmamoj:
The goal of this change is to allow for Elite2 to be started locally without any dependency on the Auth server.
In order to achieve this the property `spring.security.oauth2.resourceserver.jwt.issuer-uri` is not set locally in order to turn off JWT issuer verification. JWT issuer verification proved to be problematic due to conflicting requirements when running locally / running integration tests.  Issuer verification is turned on in all real environments (see https://github.com/ministryofjustice/nomis-api-terraform-azure/pull/218).

### DIFF
--- a/src/main/java/net/syscon/elite/web/config/ResourceServerConfiguration.java
+++ b/src/main/java/net/syscon/elite/web/config/ResourceServerConfiguration.java
@@ -32,7 +32,7 @@ public class ResourceServerConfiguration extends WebSecurityConfigurerAdapter {
                         "/health", "/info", "/ping", "/health/ping", "/h2-console/**",
                         "/v2/api-docs", "/api/swagger.json",
                         "/swagger-ui.html", "/swagger-resources", "/swagger-resources/configuration/ui",
-                        "/swagger-resources/configuration/security").permitAll()
+                        "/swagger-resources/configuration/security", "/auth/.well-known/jwks.json").permitAll()
                     .anyRequest().authenticated()
                 ).oauth2ResourceServer()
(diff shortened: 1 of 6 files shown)


src/main/java/net/syscon/elite/web/config/ResourceServerConfiguration.java
```diff
@@ -32,7 +32,7 @@ public void configure(final HttpSecurity http) throws Exception {
                         "/health", "/info", "/ping", "/health/ping", "/h2-console/**",
                         "/v2/api-docs", "/api/swagger.json",
                         "/swagger-ui.html", "/swagger-resources", "/swagger-resources/configuration/ui",
-                        "/swagger-resources/configuration/security").permitAll()
+                        "/swagger-resources/configuration/security", "/auth/.well-known/jwks.json").permitAll()
```
mikehalmamoj: I'm not happy about opening up the `/auth/.well-known/jwks.json` endpoint when it is only required for running locally.  Ideally I just want to permit all to this endpoint when in the `dev` profile, but I couldn't find a way to do that.

src/main/resources/application.yml
```diff
@@ -21,10 +21,7 @@ spring:
     oauth2:
       resourceserver:
         jwt:
-          issuer-uri: ${oauth.endpoint.url}/issuer
```
mikehalmamoj: Not providing an issuer when running locally means that any locally generated JWT will be valid.  If an issuer is specified then it will be validated which means that only tokens issued on a matching port are valid which is a hurdle I wanted to remove.

PR #411, opened 2020-03-09
DT-609 local public key

mikehalmamoj:
Do not use the JWK Set locally, load the public key instead.  This means that elite2 can run locally without worrying about an Auth server being up and running.


src/main/resources/application.yml
```diff
@@ -17,14 +17,7 @@ spring:
...
 
-  security:
-    oauth2:
-      resourceserver:
-        jwt:
-          issuer-uri: ${oauth.endpoint.url}/issuer
-          jwk-set-uri: ${oauth.endpoint.url}/.well-known/jwks.json
-
-oauth.endpoint.url: http://localhost:8080/auth
+oauth.endpoint.url: http://localhost:9090/auth
```
petergphillips: don't see why this is required?

mikehalmamoj: Just because if you're running locally against an Oauth server you don't want it to run on the same port as elite2.  That said, by default it's not used.  Shall I get rid?

PR #413, opened 2020-03-10
DT-370 Remove JWT issuer verification

mikehalmamoj:
Spring Security only applies the JWT issue validation if the config
property for issuer URI exists.  We want to turn it off as we have
multiple URIs for the Auth server but a multiple issuer scenario is
not supported.


mikehalmamoj:
Note that this ticket will not be merged until https://dsdmoj.atlassian.net/browse/DT-612 is done - then we can prove the issue in dev before rolling on this change to prove the fix.

PR #414, opened 2020-03-10
DT-603 changes to check court for hearing is currently active and hearing date is in future

matt-crowson-moj:
Small changes to the court hearings service to verify the date of the hearing i.e. in future and the (agency) court for the hearing is active.

### DIFF
--- a/src/test/java/net/syscon/elite/service/CourtHearingsServiceTest.java
+++ b/src/test/java/net/syscon/elite/service/CourtHearingsServiceTest.java
@@ -21,11 +21,14 @@ import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 
+import java.time.Clock;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
+import java.time.ZoneId;
 import java.util.List;
 import java.util.Optional;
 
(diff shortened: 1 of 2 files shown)


src/test/java/net/syscon/elite/service/CourtHearingsServiceTest.java
```diff
@@ -258,6 +264,38 @@ private void givenCourtNotFound() {
         when(agencyLocationRepository.findById("COURT")).thenReturn(Optional.empty());
     }
 
+    @Test
+    void scheduleHearing_errors_when_court_is_not_currently_active() {
+        givenProvidedCourtIsNotActive();
+
+        assertThatThrownBy(() -> courtHearingsService.scheduleHearing(OFFENDER_BOOKING_ID, PRISON_TO_COURT_HEARING))
+                .isInstanceOf(IllegalArgumentException.class)
+                .hasMessage("Supplied court location wih id %s is not active.", PRISON_TO_COURT_HEARING.getToCourtLocation());
```
petergphillips: not a great fan of testing error messages with dynamic strings in them, especially as text exactly the same as the java.  Would be better to just write
```java
                .hasMessage("Supplied court location wih id COURT is not active.")
```
especially since `COURT` is then used explicitly in the mockito when clause anyway

PR #415, opened 2020-03-11
DT-604 initial commit to wire in actual court hearing creation into API endpoint.

matt-crowson-moj:
Removing stubbed API return value from endpoint and wiring in call to service layer to schedule (create) court hearings and return via API endpoint.

This also fixes a bug I found during testing with the case status being active.

### DIFF
--- a/src/main/java/net/syscon/elite/api/resource/impl/OffenderMovementsResourceImpl.java
+++ b/src/main/java/net/syscon/elite/api/resource/impl/OffenderMovementsResourceImpl.java
@@ -1,36 +1,29 @@
 package net.syscon.elite.api.resource.impl;
 
-import net.syscon.elite.api.model.Agency;
 import net.syscon.elite.api.model.CourtHearing;
 import net.syscon.elite.api.model.PrisonToCourtHearing;
 import net.syscon.elite.api.resource.OffenderMovementsResource;
-import org.springframework.http.HttpStatus;
+import net.syscon.elite.service.CourtHearingsService;
 import org.springframework.http.ResponseEntity;
 import org.springframework.validation.annotation.Validated;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 
+import static org.springframework.http.HttpStatus.CREATED;
+
 @RestController
 @RequestMapping("${api.base.path}/bookings")
 @Validated
 public class OffenderMovementsResourceImpl implements OffenderMovementsResource {
-    //
-    // TODO - WIP what role(s) do we need to enforce here?
-    //
+
+    private final CourtHearingsService courtHearingsService;
+
+    public OffenderMovementsResourceImpl(final CourtHearingsService courtHearingsService) {
+        this.courtHearingsService = courtHearingsService;
+    }
+
     @Override
     public ResponseEntity<CourtHearing> prisonToCourtHearing(final Long bookingId, final PrisonToCourtHearing hearing) {
-        //
-        // TODO - WIP this is currently a stubbed response
-        //
-        return ResponseEntity.status(HttpStatus.CREATED).body(CourtHearing.builder()
-                .id(-1L)
-                .date(hearing.getCourtHearingDateTime().toLocalDate())
-                .time(hearing.getCourtHearingDateTime().toLocalTime())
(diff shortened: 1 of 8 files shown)


src/main/java/net/syscon/elite/api/resource/impl/OffenderMovementsResourceImpl.java
```diff
@@ -1,36 +1,29 @@
...
-                .id(-1L)
-                .date(hearing.getCourtHearingDateTime().toLocalDate())
-                .time(hearing.getCourtHearingDateTime().toLocalTime())
-                .location(Agency.builder()
-                        .agencyId(hearing.getToCourtLocation())
-                        .active(true)
-                        .agencyType("CRT")
-                        .build())
-                .build());
+        return ResponseEntity.status(CREATED).body(courtHearingsService.scheduleHearing(bookingId, hearing));
```
andymarke: Can you just return a CourtHearing? I guess there is an annotation to say 201 response code?

src/test/java/net/syscon/elite/api/resource/impl/OffenderMovementsResourceImplIntTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+public class OffenderMovementsResourceImplIntTest extends ResourceTest {
+
+    @Test
+    public void schedules_court_hearing() {
+        final var token = authTokenHelper.getToken(AuthTokenHelper.AuthToken.NORMAL_USER);
+
+        final var request = createHttpEntity(token, Map.of(
+                "courtCaseId", -1,
+                "fromPrisonLocation", "LEI",
```
andymarke: Sort on interesting that you need a prison location to create a court hearing, I wonder what happens if they are being held in the court and there are two hearings, would the fromPrisonLocation still be a prison? I guess this models what you have to input to NOMIS

src/main/java/net/syscon/elite/service/CourtHearingsService.java
```diff
@@ -56,6 +57,7 @@ public CourtHearingsService(final OffenderBookingRepository offenderBookingRepos
     }
 
     @Transactional
+    @VerifyBookingAccess
     public CourtHearing scheduleHearing(final Long bookingId, final PrisonToCourtHearing hearing) {
```
Mjwillis: Probably will want the `@HasWriteScope`

Mjwillis: On the resource you'll also want `@ProxyUser`

src/test/java/net/syscon/elite/api/resource/impl/OffenderMovementsResourceImplIntTest.java
matt-crowson-moj: Indeed it is really supporting what NOMIS does at this point in time.  Happy to talk about though.

src/main/java/net/syscon/elite/api/resource/impl/OffenderMovementsResourceImpl.java
matt-crowson-moj: The challenge in (some respects) is looking for consistency in the elite code base.  It isn't always consistent.

src/main/java/net/syscon/elite/service/CourtHearingsService.java
matt-crowson-moj: Done

src/test/java/net/syscon/elite/api/resource/impl/OffenderMovementsResourceImplIntTest.java
```diff
@@ -0,0 +1,135 @@
...
+                ErrorResponse.builder()
+                        .status(404)
+                        .userMessage("Resource with id [9999999] not found.")
+                        .developerMessage("Resource with id [9999999] not found.")
+                        .build());
+    }
+
+    @Test
+    public void schedule_court_hearing_fails_when_no_matching_prison() {
+        final var token = authTokenHelper.getToken(AuthTokenHelper.AuthToken.NORMAL_USER);
```
Mjwillis: I’d add a test for a user without permissions or write scope access

PR #418, opened 2020-03-12
DT-609 remote jwk set

mikehalmamoj:
Use the existing RemoteJWKSet to retrieve the JWK set as this supports caching out of the box.  And is standard.

### DIFF
--- a/src/main/java/net/syscon/elite/web/config/JwkClient.java
+++ b/src/main/java/net/syscon/elite/web/config/JwkClient.java
@@ -1,68 +1,64 @@
 package net.syscon.elite.web.config;
 
 import com.nimbusds.jose.JOSEException;
+import com.nimbusds.jose.RemoteKeySourceException;
 import com.nimbusds.jose.jwk.JWK;
-import com.nimbusds.jose.jwk.JWKSet;
+import com.nimbusds.jose.jwk.JWKMatcher;
+import com.nimbusds.jose.jwk.JWKSelector;
 import com.nimbusds.jose.jwk.RSAKey;
+import com.nimbusds.jose.jwk.source.DefaultJWKSetCache;
+import com.nimbusds.jose.jwk.source.RemoteJWKSet;
+import com.nimbusds.jose.proc.JWKSecurityContext;
+import com.nimbusds.jose.util.DefaultResourceRetriever;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.lang3.tuple.Pair;
 import org.springframework.beans.factory.annotation.Value;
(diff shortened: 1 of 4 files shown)


src/test/java/net/syscon/elite/web/config/ClientTrackingTelemetryModuleTest_JwkClient.java
```diff
@@ -0,0 +1,119 @@
...
+
+    @After
+    public void tearDown() {
+        ThreadContext.remove();
+        jwkServer.stop();
+        jwkServer.resetAll();
+    }
+
+    @Test
+    public void shouldAddClientIdAndUserNameToInsightTelemetry() {
```
matt-crowson-moj: Perhaps include clientIpAddress as well in test name.  As you are testing for the presence of 3 things. Meh.

src/main/java/net/syscon/elite/web/config/JwkClient.java
```diff
@@ -1,68 +1,64 @@
...
-    }
-
-    private Map<String, PublicKey> getPublicKeysById() {
-        return loadJwkKeys().stream()
-                .map(jwk -> Pair.of(jwk.getKeyID(), toPublicKey(jwk)))
-                .filter(pair -> pair.getRight().isPresent())
-                .collect(Collectors.toUnmodifiableMap(Pair::getLeft, pair -> pair.getRight().get()));
+        final var jwkSelector = new JWKSelector(new JWKMatcher.Builder().keyID(keyId).build());
+        final var jwk = getJwk(jwkSelector);
+        return jwk.map(this::toPublicKey)
```
matt-crowson-moj: I think you can just flatMap this e.g.  `return jwk.flatMap(this::toPublicKey).orElseGet(null);`

src/main/java/net/syscon/elite/web/config/JwkClient.java
```diff
@@ -1,68 +1,64 @@
...
         }
-        return emptyList();
     }
 
     private Optional<PublicKey> toPublicKey(final JWK jwk) {
         try {
             return Optional.of(((RSAKey) jwk).toPublicKey());
         } catch (JOSEException e) {
-            log.info(String.format("Failed to retrieve public key from JWK %s due to exception", jwk), e);
+            log.error(format("Failed to retrieve public key from JWK %s due to exception", jwk), e);
```
matt-crowson-moj: Do you need the format here? Why not just `log.error("Failed to retrieve public key from JWK {} due to exception", jwk, e);`

PR #421, opened 2020-03-13
Nn 2493 location casing update

Nimphal:
Currently the location formatter will titlecase most abbreviations, resulting in rooms showing up as "Vcc Room 1", for example. This change introduces a list of abbreviations to be excluded from titlecasing and adds a test for the specific method.

### DIFF
--- a/src/main/java/net/syscon/elite/service/support/LocationProcessor.java
+++ b/src/main/java/net/syscon/elite/service/support/LocationProcessor.java
@@ -69,6 +69,7 @@ public class LocationProcessor {
      * @throws {@code NullPointerException} if no location provided for processing.
      */
     public static List<Location> processLocations(final List<Location> locations, final boolean preferUserDescription) {
+        System.out.println(locations);
         Objects.requireNonNull(locations);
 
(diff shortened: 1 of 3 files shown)


src/main/java/net/syscon/elite/service/support/LocationProcessor.java
```diff
@@ -125,10 +132,35 @@ private static String newDescription(Location location, boolean preferUserDescri
...
+        // the letter sequence should it appear in the middle of a word
+        // e.g. this will not match 'mosaic' even though AIC is one of the abbreviations
+        Pattern pattern = Pattern.compile("\\b(" + String.join("|", ABBREVIATIONS) + ")\\b", Pattern.CASE_INSENSITIVE);
+        Matcher matcher = pattern.matcher(description);
+
+        // There could be more than one abbreviation in a string,
+        // e.g. HMP Moorland VCC Room 1
+        // By using the string buffer and the appendReplacement method
+        // we ensure that all the matching groups are replaced accordingly
+        StringBuffer stringBuffer = new StringBuffer();
```
matt-crowson-moj: Really minor point but nowadays unless you need synchronisation then favour `StringBuilder` over `StringBuffer`.

PR #422, opened 2020-03-16
NN-2489: Add the created by user id to appointment response

Nimphal:


Nimphal:
Endpoint is only used by us so should be good :)

PR #425, opened 2020-03-18
DT-651 wiring in call to concreate service method for retrieval of court hearings

matt-crowson-moj:
This still needs some filtering to be added though the number of hearings returned realistically will never be that much for a given booking.

### DIFF
--- a/src/main/java/net/syscon/elite/api/model/CourtHearings.java
+++ b/src/main/java/net/syscon/elite/api/model/CourtHearings.java
@@ -6,12 +6,17 @@ import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
+import lombok.Singular;
+
+import java.util.List;
 
 @ApiModel(description = "Represents court hearings for an offender booking.")
 @Builder
 @AllArgsConstructor
+@NoArgsConstructor
 @Data
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public class CourtHearings {
-    // TODO - WIP DT-651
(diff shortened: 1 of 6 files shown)


src/main/java/net/syscon/elite/service/CourtHearingsService.java
```diff
@@ -91,17 +91,36 @@ public CourtHearing scheduleHearing(final Long bookingId, final Long courtCaseId
...
-    // TODO - WIP DT-651
+    /**
+     * Returns all court hearings for a given booking ID.
+     */
     @VerifyBookingAccess
     public CourtHearings getCourtHearingsFor(final Long bookingId) {
-        return CourtHearings.builder().build();
+        final var courtHearingsBuilder = CourtHearings.builder();
+
+        courtEventRepository.findByOffenderBooking_BookingId(bookingId).stream()
```
matt-crowson-moj: Note: I am not restricting to active bookings.  It could but should it?  The same goes for court cases the court hearings will be linked to.

src/main/java/net/syscon/elite/api/model/CourtHearings.java
```diff
@@ -6,12 +6,17 @@
...
 
 @ApiModel(description = "Represents court hearings for an offender booking.")
 @Builder
 @AllArgsConstructor
+@NoArgsConstructor
 @Data
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public class CourtHearings {
-    // TODO - WIP DT-651
+    @Singular
```
andymarke: oooh, never used that before

PR #427, opened 2020-03-19
WIP DT-635 Audit updates from all users

mikehalmamoj:
All users with scope `proxy-user` will continue to have write access and
only Nomis users will connect to the DB as the user (and fully control
the audit columns on an update via the set_context stuff).

For non-Nomis users, the user id will still be stamped on the created by
field but will also appear in the AUDIT_CLIENT_USER_ID.

If any updates require a Nomis user in order to be allowed, these are
currently rejected with a 404 if the user is not a Nomis user.  I don't
see any reason to change this or beef it up with some kind of annotation
guarding these updates.  If I try to update with a Delius user and get a
`404 User not found` then I think that's self-explanatory.

### DIFF
--- a/src/main/java/net/syscon/elite/aop/connectionproxy/HsqlConnectionAspect.java
+++ b/src/main/java/net/syscon/elite/aop/connectionproxy/HsqlConnectionAspect.java
@@ -19,8 +19,9 @@ public class HsqlConnectionAspect extends AbstractConnectionAspect {
 
     @Override
     protected Connection openProxySessionIfIdentifiedAuthentication(final Connection pooledConnection) throws SQLException {
-        if (authenticationFacade.isIdentifiedAuthentication()) {
-            log.trace("Configuring Hsql Proxy Session.");
+        final var proxyUserAuthSource = authenticationFacade.getProxyUserAuthenticationSource();
+        if (proxyUserAuthSource.equals("nomis")) {
+            log.debug("Configuring Hsql Proxy Session.");
             return openAndConfigureProxySessionForConnection(pooledConnection);
         }
         return pooledConnection;
--- a/src/main/java/net/syscon/elite/security/AuthenticationFacade.java
+++ b/src/main/java/net/syscon/elite/security/AuthenticationFacade.java
@@ -44,14 +44,15 @@ public class AuthenticationFacade {
         return username;
     }
 
-    public boolean isIdentifiedAuthentication() {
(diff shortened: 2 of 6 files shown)


mikehalmamoj:
Sorry this was meant to be a draft PR (hence the WIP - but that's a GitLab thing).  Please do not merge!!!

src/main/java/net/syscon/elite/aop/connectionproxy/HsqlConnectionAspect.java
```diff
@@ -19,8 +19,9 @@ public HsqlConnectionAspect(final AuthenticationFacade authenticationFacade) {
 
     @Override
     protected Connection openProxySessionIfIdentifiedAuthentication(final Connection pooledConnection) throws SQLException {
-        if (authenticationFacade.isIdentifiedAuthentication()) {
-            log.trace("Configuring Hsql Proxy Session.");
+        final var proxyUserAuthSource = authenticationFacade.getProxyUserAuthenticationSource();
+        if (proxyUserAuthSource.equals("nomis")) {
+            log.debug("Configuring Hsql Proxy Session.");
```
petergphillips: better to be `trace` otherwise fills up the logs

src/main/resources/application-nomis.yml
```diff
@@ -8,7 +8,8 @@ server:
 
 spring:
   profiles:
-    include: nomis-local,logstash,ai
+#    include: nomis-local,logstash,ai
+    include: nomis-local,ai
```
petergphillips: did you mean to make this change?

src/main/java/net/syscon/elite/security/AuthenticationFacade.java
```diff
@@ -44,14 +44,15 @@ public String getCurrentUsername() {
         return username;
     }
 
-    public boolean isIdentifiedAuthentication() {
+    public String getProxyUserAuthenticationSource() {
```
petergphillips: wondering if this should be an enum as values are known?

src/main/resources/application-nomis.yml
mikehalmamoj: Nope, thanks.

src/main/java/net/syscon/elite/security/AuthenticationFacade.java
mikehalmamoj: Yeah, good idea.

PR #428, opened 2020-03-19
NN-2508: Abstract string formatting for strings with abbreviations, apply to appointment types

Nimphal:


src/main/java/net/syscon/elite/service/support/StringWithAbbreviationsProcessor.java
```diff
@@ -0,0 +1,84 @@
...
+        // Handle the possibility of the userDescription being empty
+        if (string == null) {
+            return null;
+        }
+        var description = WordUtils.capitalizeFully(string);
+        // Using word boundaries to find the right string ensures we catch the strings
+        // wherever they appear in the description, while also avoiding replacing
+        // the letter sequence should it appear in the middle of a word
+        // e.g. this will not match 'mosaic' even though AIC is one of the abbreviations
+        Pattern pattern = Pattern.compile("\\b(" + String.join("|", ABBREVIATIONS) + ")\\b", Pattern.CASE_INSENSITIVE);
```
andrewrlee: Generally good practice to extract out Patterns as constants where possible. 
They are generally quite resource intensive to create, not that it will matter much in this code path - but its a bit of a convention if possible

src/test/java/net/syscon/elite/service/support/StringWithAbbreviationsProcessorTest.java
```diff
@@ -0,0 +1,14 @@
...
+
+import org.junit.jupiter.api.Test;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class StringWithAbbreviationsProcessorTest {
+    @Test
+    public void format() {
+        // Assert that all abbreviations are unchanged, even when there a multiple ones in a string
+        final var abbreviationsString = String.join(", ", StringWithAbbreviationsProcessor.ABBREVIATIONS);
+        assertThat(StringWithAbbreviationsProcessor.format(abbreviationsString)).isEqualTo(abbreviationsString);
```
andrewrlee: could have explicit tests - empty string, null, no match, multiple matches, match at end of string

Nimphal: Because this tacks all the abbreviations in one place it tackles a few of the above, like end of string and multiple matches, but you're right that it will be better to make that explicit.

PR #437, opened 2020-03-24
DT-663 adding intial cut of offender jpa entity.

matt-crowson-moj:
PR includes some minor JPA annotation tidy ups as well.  Mainly around the optional = false attribute on ManyToOne annotations.  This is to aid lazy loading of non-collection attributes.

### DIFF
--- a/src/main/java/net/syscon/elite/repository/jpa/model/Offender.java
+++ b/src/main/java/net/syscon/elite/repository/jpa/model/Offender.java
@@ -1,0 +1,66 @@
+package net.syscon.elite.repository.jpa.model;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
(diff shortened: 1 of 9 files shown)


src/main/java/net/syscon/elite/repository/jpa/model/Offender.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @SequenceGenerator(name = "OFFENDER_ID", sequenceName = "OFFENDER_ID", allocationSize = 1)
+    @GeneratedValue(generator = "OFFENDER_ID")
+    @Id
+    @Column(name = "OFFENDER_ID", nullable = false)
+    private Long id;
+
+    @Column(name = "ID_SOURCE_CODE", nullable = false)
+    @Builder.Default
+    private String idSourceCode = "SEQ";
```
matt-crowson-moj: I think this is right @Mjwillis

PR #444, opened 2020-03-27
DT-685: Switch to quay for keyworker image and do not verify jwt for …

petergphillips:
…logging


build.gradle
```diff
@@ -53,10 +53,9 @@ def today = Instant.now()
 def todaysDate = LocalDate.now().format('yyyy-MM-dd')
 version = System.getenv('CI') ? "${todaysDate}.${System.getenv('CIRCLE_BUILD_NUM')}" : todaysDate
 
-jar {
-    baseName = rootProject.name
+bootJar {
```
petergphillips: fix for not printing out version during startup

docker-compose.yml
```diff
@@ -25,7 +25,7 @@ services:
       - SPRING_PROFILES_ACTIVE=nomis-hsqldb
 
   keyworker-api:
-    image: mojdigitalstudio/keyworker-api:latest
+    image: quay.io/hmpps/keyworker-api:latest
```
petergphillips: keyworker now using quay

PR #445, opened 2020-03-27
DT-692 adding activeOnly filter flag for court cases associated with a booking.

matt-crowson-moj:
Adding filter attribute to endpoint to allow inclusion of all court cases or only active active court cases.  Defaults to active only.

### DIFF
--- a/src/main/java/net/syscon/elite/service/BookingService.java
+++ b/src/main/java/net/syscon/elite/service/BookingService.java
@@ -6,7 +6,7 @@ import net.syscon.elite.api.support.Order;
 import net.syscon.elite.api.support.Page;
 import net.syscon.elite.repository.BookingRepository;
 import net.syscon.elite.repository.SentenceRepository;
-import net.syscon.elite.repository.jpa.model.OffenderBooking;
+import net.syscon.elite.repository.jpa.model.OffenderCourtCase;
 import net.syscon.elite.repository.jpa.model.ReferenceCode;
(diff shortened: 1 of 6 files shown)


src/main/java/net/syscon/elite/service/BookingService.java
```diff
@@ -626,13 +626,17 @@ public MilitaryRecords getMilitaryRecords(final Long bookingId) {
...
+    public List<CourtCase> getOffenderCourtCases(final Long bookingId, final boolean activeOnly) {
         return offenderBookingRepository.findById(bookingId)
-                .map(OffenderBooking::getCourtCases)
+                .map(booking -> activeOnly ? active(booking.getCourtCases()) : booking.getCourtCases())
                 .map(CourtCaseTransformer::transform)
                 .orElseThrow(EntityNotFoundException.withId(bookingId));
     }
 
+    private Collection<OffenderCourtCase> active(final Collection<OffenderCourtCase> courtCases) {
+        return courtCases.stream().filter(OffenderCourtCase::isActive).collect(Collectors.toUnmodifiableList());
```
mikehalmamoj: Maybe put the filter in the main stream in the calling method?  Less complex and no need for a second stream on the same thing...

mikehalmamoj: After discussion I see where you are coming from... ignore me :)

matt-crowson-moj: I will push this down in to the Domain ;)

PR #448, opened 2020-03-30
DT-696 court events relationship to court cases is optional.

matt-crowson-moj:
The relationship between court events and court cases should be optional.

A real world example of this would be a prisoner has been asked to go to court as a witness, so it is not linked to their specific court case(s).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -155,7 +155,7 @@ dependencies {
 
     compileOnly 'org.projectlombok:lombok:1.18.12'
 
-    runtime 'org.hsqldb:hsqldb:2.5.0'
+    runtime 'org.hsqldb:hsqldb:2.4.1'
     runtime 'org.flywaydb:flyway-core:6.3.2'
 
     testImplementation 'org.springframework.boot:spring-boot-starter-test'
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -155,7 +155,7 @@ dependencies {
 
     compileOnly 'org.projectlombok:lombok:1.18.12'
 
-    runtime 'org.hsqldb:hsqldb:2.5.0'
+    runtime 'org.hsqldb:hsqldb:2.4.1'
```
matt-crowson-moj: This change is not related to the underlying ticket.

It would appear due to the clock change at the weekend the CaseNoteRepositoryTest is now failing with date time related issues.  A ticket will be added to the backlog to investigate and revert back to the more recent version.

src/test/java/net/syscon/elite/repository/jpa/repository/CourtEventRepositoryTest.java
```diff
@@ -60,37 +64,51 @@ void setup() {
...
+        final var bookingWithoutCourtCase = offenderBookingRepository.findById(BOOKING_WITHOUT_COURT_CASE).orElseThrow();
+
+        assertThat(bookingWithoutCourtCase.getCourtCases()).isEmpty();
+
+        final var savedCourtEventWithoutCourtCase = courtEventRepository.save(builder
+                .offenderBooking(bookingWithoutCourtCase)
+                .offenderCourtCase(null)
+                .build());
+
+        assertThat(courtEventRepository.findById(savedCourtEventWithoutCourtCase.getId()).orElseThrow()).isEqualTo(savedCourtEventWithoutCourtCase);
```
brightonsbox: Meh.. could get rid of the orElseThrow() and do an `assertThat(...).contains()`

Could maybe also call the new method getOffenderCourtCase() to prove it returns empty?

PR #449, opened 2020-03-31
NN-2529: Return suspended flag as part of all activities endpoint

Nimphal:


src/test/java/net/syscon/elite/repository/ScheduleRepositoryTest.java
```diff
@@ -245,6 +245,9 @@ public void testGetAllActivitiesAtAgency() {
 
             // Get offender cell locations. -1L and -3L share a cell.
             assertThat(List.of("LEI-A-1-1", "LEI-A-1-2", "LEI-H-1-5", "LEI-A-1", "LEI-A-1-10", "MDI-1-1-001")).contains(result.getCellLocation());
+
+            // Assert it return both suspended and not suspended
+            assertThat(List.of(true, false)).contains(result.getSuspended());
```
brightonsbox: I was going to say you could maybe do a `assertThat(results).extracting(result -> result.getSuspended()).containsExactly(...)` to check that the right results have the expected values, but then I realised that this is returning 91 results so fair play! 😆

PR #452, opened 2020-04-02
DCS-198 Incidents by religion. DCS-199 Incidents by ethnic group.

bell-pa:
Add new fields `ethnicityCode` and `religionCode` to `/api/prisoners` and related resource responses. (`PrisonerDetail` class).
These primary key values are more stable than the description values used for `ethnicity` and `religion` and are easier to match. They will be used for new Use of Force reports - incidents by religious group and incidents by ethnic group. (DCS-198, DCS-199)

### DIFF
--- a/src/main/java/net/syscon/elite/api/model/PrisonerDetail.java
+++ b/src/main/java/net/syscon/elite/api/model/PrisonerDetail.java
@@ -3,10 +3,7 @@ package net.syscon.elite.api.model;
 import com.fasterxml.jackson.annotation.*;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
-import lombok.AllArgsConstructor;
-import lombok.Builder;
-import lombok.EqualsAndHashCode;
-import lombok.NoArgsConstructor;
+import lombok.*;
 
 import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.NotNull;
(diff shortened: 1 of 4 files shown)


src/main/java/net/syscon/elite/api/model/PrisonerDetail.java
```diff
@@ -24,73 +21,108 @@
...
     @NotNull
     private LocalDate dateOfBirth;
 
+    @ApiModelProperty(required = true, value = "The prisoner's gender.")
     @NotBlank
     private String gender;
 
+    @ApiModelProperty(required = true, value = "The prisoner's gender code.")
     @NotBlank
     private String sexCode;
```
andrewrlee, line 60: so they do it for this anyway which is good!

src/main/java/net/syscon/elite/api/model/PrisonerDetail.java
```diff
@@ -24,73 +21,108 @@
 @AllArgsConstructor
 @NoArgsConstructor
 @EqualsAndHashCode
+@Data
 public class PrisonerDetail {
     @JsonIgnore
     private Map<String, Object> additionalProperties;
 
+    @ApiModelProperty(required = true, value = "The prisoner's unique offender number (aka NOMS Number in the UK).")
```
andrewrlee, line 29: tend to put examples on these now..

src/main/java/net/syscon/elite/api/model/PrisonerDetail.java
Mjwillis: Nice to give an example

src/main/java/net/syscon/elite/api/model/PrisonerDetail.java
bell-pa: I've added examples for the religion, religionCode, ethnicity and ethnicityCode properties.

PR #455, opened 2020-04-03
GDPR-78 Improving integration test to prevent regression of lazy loading bug fix

brightonsbox:


src/main/java/uk/gov/justice/hmpps/nomis/datacompliance/repository/jpa/model/OffenderAliasPendingDeletion.java
```diff
@@ -45,6 +46,6 @@
     private LocalDate birthDate;
 
     @Singular
-    @OneToMany(mappedBy = "offenderAlias")
+    @OneToMany(mappedBy = "offenderAlias", fetch = FetchType.EAGER)
```
brightonsbox, line 49: Because the API that triggers the fetching of this object is async, JPA was losing track of the session and so couldn't lazily load this field.  Because we are transforming the entirety of this object anyway, there is no problem with eagerly loading the associated offenderBookings to resolve this issue.

src/test/java/uk/gov/justice/hmpps/nomis/datacompliance/controller/DataComplianceControllerTest.java
```diff
@@ -3,27 +3,34 @@
...
 public class DataComplianceControllerTest extends ResourceTest {
 
     private static final String REQUEST_ID = "123";
-    private static final LocalDateTime WINDOW_START = LocalDateTime.now();
-    private static final LocalDateTime WINDOW_END = WINDOW_START.plusDays(1);
 
-    @MockBean
-    private OffenderDataComplianceService offenderDataComplianceService;
+    // This date is 7 years after the SED_CALCULATED_DATE of the expected record
+    private static final LocalDateTime WINDOW_START = LocalDateTime.of(2027, 3, 24, 0, 0);
```
mikehalmamoj, line 29: You could call the variable something that indicates its purpose... like WINDOW_START_AFTER_DELETION_PERIOD.  And I guess that number 7 is defined somewhere?  You could use that when building the date.

src/main/java/uk/gov/justice/hmpps/nomis/datacompliance/repository/jpa/model/OffenderAliasPendingDeletion.java
mikehalmamoj: Bet that was fun working that out. :+1:

PR #460, opened 2020-04-06
DT-708 wiring in endpoint to service layer for court hearings without court case.

matt-crowson-moj:
Adding new point for court hearings with court case and extending the integration test coverage.


src/test/java/net/syscon/elite/api/resource/impl/OffenderMovementsResourceImplIntTest_scheduleCourtCaseHearing.java
```diff
@@ -25,17 +26,23 @@ public void schedules_court_hearing() {
...
-                "/api/bookings/-1/court-cases/-1/prison-to-court-hearings",
+                "/api/bookings/-2/court-cases/-2/prison-to-court-hearings",
                 HttpMethod.POST,
                 request,
                 new ParameterizedTypeReference<String>() {
                 });
 
-        assertThatJsonFileAndStatus(response, 201, "schedule_court_hearing.json");
+        assertThat(response.getStatusCode()).isEqualTo(CREATED);
+        assertThat(getBodyAsJsonContent(response)).extractingJsonPathNumberValue("$.id").isNotNull();
```
matt-crowson-moj, line 36: This is a DB sequence generated value so we cannot easily assert on the real value here due to other tests.

src/test/java/net/syscon/elite/api/resource/impl/ResourceTest.java
```diff
@@ -26,7 +25,6 @@
 import static org.springframework.boot.test.context.SpringBootTest.WebEnvironment.RANDOM_PORT;
 import static org.springframework.core.ResolvableType.forType;
 
-@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_CLASS)
```
matt-crowson-moj, line 29: So I have taken this off as it does not appear to have any impact/effect on the tests.  Shout if you disagree.

PR #464, opened 2020-04-07
DT-235 movetocell bycode

mikehalmamoj:
Accept the new cell's code rather than its id.

### DIFF
--- a/src/main/java/net/syscon/elite/service/MovementUpdateService.java
+++ b/src/main/java/net/syscon/elite/service/MovementUpdateService.java
@@ -3,6 +3,8 @@ package net.syscon.elite.service;
 import com.amazonaws.util.StringUtils;
 import net.syscon.elite.api.model.OffenderBooking;
 import net.syscon.elite.core.HasWriteScope;
+import net.syscon.elite.repository.jpa.model.AgencyInternalLocation;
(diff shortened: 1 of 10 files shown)


src/main/java/net/syscon/elite/service/MovementUpdateService.java
```diff
@@ -23,55 +25,77 @@
...
                 dateTime == null || dateTime.isBefore(LocalDateTime.now(clock)) || dateTime.isEqual(LocalDateTime.now(clock)),
                 "The date cannot be in the future"
         );
     }
 
+    private void checkReasonCode(String reasonCode) {
+        try {
+            referenceDomainService.getReferenceCodeByDomainAndCode(CELL_MOVE_REASON.getDomain(), reasonCode, false);
+        } catch(EntityNotFoundException e) {
+            throw new IllegalArgumentException(e.getMessage(), e);
```
brightonsbox, line 77: meh.. `throw new IllegalArgumentException(e);` would be equivalent

mikehalmamoj: That's what I thought... but it turns out the `public Throwable(Throwable cause)` does  `this.detailMessage = cause == null ? null : cause.toString();` and so you get more than just the message.

PR #467, opened 2020-04-08
DT-722 - Handle prisoners released on the /full-status endpoint

Mjwillis:


src/main/java/net/syscon/elite/api/model/PrisonerInformation.java
```diff
@@ -80,24 +80,26 @@
 
     public void deriveLegalStatus(String bandCodeStr) {
         if (bandCodeStr != null) {
-            final var bandCode = Integer.valueOf(bandCodeStr);
+            final var bandCode = Integer.parseInt(bandCodeStr);
```
andymarke, line 83: curious, why did you make this change?

Mjwillis: IntelliJ recommended.

PR #468, opened 2020-04-09
GDPR-51 Echoing back the referral ID in deletion complete message

brightonsbox:


src/main/java/uk/gov/justice/hmpps/nomis/datacompliance/events/dto/OffenderDeletionCompleteEvent.java
```diff
@@ -10,6 +11,11 @@
 @NoArgsConstructor
 @Data
 public class OffenderDeletionCompleteEvent {
+
+    @JsonProperty("offenderIdDisplay")
```
andymarke, line 15: Never understood this; but do you need this annotation when the name and field have the same name?

brightonsbox: No, this is just habit really, it just brings more attention to the fact that the name really matters and could break API contracts.

PR #473, opened 2020-04-22
CAT-586 Add other categories to the cat-tool service

steverendell:
Add extra fields that may be set by create / approve categorisation

### DIFF
--- a/src/main/java/net/syscon/elite/service/InmateService.java
+++ b/src/main/java/net/syscon/elite/service/InmateService.java
@@ -36,6 +36,8 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static net.syscon.elite.repository.support.StatusFilter.ACTIVE_ONLY;
+import static net.syscon.elite.repository.support.StatusFilter.ALL;
 import static net.syscon.elite.service.support.InmatesHelper.deriveClassification;
 import static net.syscon.elite.service.support.InmatesHelper.deriveClassificationCode;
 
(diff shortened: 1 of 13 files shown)


src/test/resources/features/booking_details.feature
```diff
@@ -1,3 +1,4 @@
+@sdar
```
brightonsbox, line 1: What is this doing?

src/main/java/net/syscon/elite/service/InmateService.java
```diff
@@ -593,6 +608,13 @@ private void validate(final CategoryApprovalDetail detail) {
         } catch (final EntityNotFoundException ex) {
             throw new HttpClientErrorException(HttpStatus.BAD_REQUEST, "Committee Code not recognised.");
         }
+        if (StringUtils.isNotBlank(detail.getApprovedPlacementAgencyId())) {
+            try {
+                agencyService.getAgency(detail.getApprovedPlacementAgencyId(), ACTIVE_ONLY, "INST");
+            } catch (final EntityNotFoundException ex) {
+                throw new HttpClientErrorException(HttpStatus.BAD_REQUEST, "Review placement agency id not recognised.");
```
brightonsbox, line 615: Maybe could exercise these error conditions in tests?

src/test/resources/features/booking_details.feature
steverendell: whoops good spot

src/main/java/net/syscon/elite/service/InmateService.java
steverendell: its covered in OffenderAssessmentResourceTest

PR #474, opened 2020-04-23
DT-704: Add validation to swagger docs and fix

petergphillips:


src/main/java/net/syscon/elite/web/config/SwaggerConfig.java
```diff
@@ -64,18 +63,17 @@ private Contact contactInfo() {
     }
 
     private ApiInfo apiInfo() {
-        final var vendorExtension = new StringVendorExtension("", "");
-        final Collection<VendorExtension> vendorExtensions = new ArrayList<>();
-        vendorExtensions.add(vendorExtension);
```
petergphillips, line 69: was causing parse errors

PR #477, opened 2020-04-23
DT-704: Fix issue with local date

petergphillips:


src/main/java/net/syscon/elite/api/model/v1/AvailableDates.java
```diff
@@ -15,6 +19,6 @@
 @ToString
 public class AvailableDates {
 
-    @ApiModelProperty(value = "Available Dates", allowEmptyValue = true)
+    @ApiModelProperty(value = "Available Dates", dataType = "[Ljava.sql.Date;", allowEmptyValue = true)
```
petergphillips, line 22: had to manually supply data type as wasn't inferred correctly leaving #definitions/LocalDate not defined error

src/main/java/net/syscon/elite/web/config/SwaggerConfig.java
```diff
@@ -36,18 +37,16 @@ public SwaggerConfig(@Autowired(required = false) final BuildProperties buildPro
...
-        docket.genericModelSubstitutes(Optional.class);
-        docket.directModelSubstitute(ZonedDateTime.class, Date.class);
-        docket.directModelSubstitute(LocalDateTime.class, Date.class);
-
-        return docket;
+                .apiInfo(apiInfo())
+                .genericModelSubstitutes(Optional.class)
+                .directModelSubstitute(ZonedDateTime.class, Date.class)
+                .directModelSubstitute(LocalDateTime.class, Date.class)
+                .directModelSubstitute(LocalDate.class, java.sql.Date.class);
```
petergphillips, line 49: this line should have been the fix, but didn't do anything.  Leaving in, in case it is fixed properly some time.

PR #479, opened 2020-04-24
DT-772: Convert junit to assertj

petergphillips:


src/test/java/net/syscon/elite/executablespecification/steps/AdjudicationSteps.java
```diff
@@ -53,19 +52,19 @@ protected void init() {
     }
 
     public void verifyNoAwards() {
-        assertTrue("There are " + summary.getAwards().size() + " awards", summary.getAwards().isEmpty());
+        assertThat(summary.getAwards().isEmpty()).as("There are " + summary.getAwards().size() + " awards").isTrue();
```
petergphillips, line 55: Could do with manually fixing up some of these, but this is a good start

PR #482, opened 2020-04-27
DT-772: Fix main offence sql parameters

petergphillips:


src/main/java/net/syscon/elite/repository/impl/SentenceRepositoryImpl.java
```diff
@@ -28,7 +28,7 @@
 
         return jdbcTemplate.query(
                 sql,
-                createParams("bookingId", bookingId, "mostSerious", "Y", "chargeStatus", "A", "severityRanking", 999),
+                createParams("bookingId", bookingId, "mostSerious", "Y", "chargeStatus", "A", "severityRanking", "999"),
```
petergphillips, line 31: turns out that severity ranking is stored in oracle as
```
	SEVERITY_RANKING VARCHAR2(12 char),
```
so needs to be sent in as a string instead (okay in hsqldb, but oracle moans)

PR #485, opened 2020-04-28
Nn 2589 new scheduled activity endpoint

JamieDraperUK:


src/main/java/net/syscon/elite/api/resource/ScheduleResource.java
```diff
@@ -96,15 +105,22 @@
...
                                          @ApiParam(value = "The required offender numbers (mandatory)", required = true) @RequestBody List<String> body,
                                          @ApiParam(value = "Date of whereabouts list, default today") @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) @RequestParam(value = "date", required = false) LocalDate date,
                                          @ApiParam(value = "AM, PM or ED", allowableValues = "AM,PM,ED") @RequestParam(value = "timeSlot", required = false) TimeSlot timeSlot,
                                          @ApiParam(value = "Whether to include 'excluded' activities in the results", defaultValue = "false") @RequestParam(value = "includeExcluded", required = false, defaultValue = "false") boolean includeExcluded);
 
+
+    @PostMapping("/activities")
+    @ApiOperation(value = "", nickname = "getActivities")
+    @ApiResponses(value = {
+            @ApiResponse(code = 200, message = "", response = PrisonerSchedule.class, responseContainer = "List")})
```
petergphillips, line 121: unnecessary 200 - will be generated automatically (esp. as aren't setting a custom message)

src/main/java/net/syscon/elite/api/resource/ScheduleResource.java
```diff
@@ -96,15 +105,22 @@
...
+    List<PrisonerSchedule> getActivities(@ApiParam(value = "Event ids(mandatory)", required = true) @RequestBody List<Long> body);
```
petergphillips, line 122: do you need a `NotEmpty` or other annotation here to enforce the mandatory nature?  Can just see it returning no results if the body isn't passed in.  Should it be a 400?  Could do with a resource test for that...

src/main/java/net/syscon/elite/api/resource/ScheduleResource.java
JamieDraperUK: The code seems to be a required parameter of ApiResponse. What did you have in mind? Happy to change this and the rest in this file.

src/main/java/net/syscon/elite/api/resource/ScheduleResource.java
JamieDraperUK: Will add one now.

src/main/java/net/syscon/elite/api/resource/impl/SchedulesResourceImpl.java
```diff
@@ -76,10 +76,15 @@ public SchedulesResourceImpl(final SchedulesService schedulesService, final Appo
...
 
     @Override
-    public List<PrisonerSchedule> getActivities(final String agencyId, final List<String> body, final LocalDate date, final TimeSlot timeSlot, final boolean includeExcluded) {
+    public List<PrisonerSchedule> getActivitiesForBookings(final String agencyId, final List<String> body, final LocalDate date, final TimeSlot timeSlot, final boolean includeExcluded) {
         return schedulesService.getActivities(agencyId, body, date, timeSlot, includeExcluded);
     }
 
+    @Override
+    public List<PrisonerSchedule> getActivities(final List<Long> eventIds) {
+        return schedulesService.getActivities(eventIds);
```
petergphillips, line 85: can't remember how does security work for this - will anyone with a token be able to access everything?

JamieDraperUK: Good catch! Not sure the schedule is too sensitive but will verify the agency like the other methods in the service. At least then it's consistent.

PR #486, opened 2020-04-29
DT-786: Document movement type and fix swagger docs

petergphillips:


src/main/java/net/syscon/elite/api/resource/MovementResource.java
```diff
@@ -64,7 +76,6 @@ MovementCount getRollcountMovements(
     @GetMapping("/{agencyId}/enroute")
     @ApiOperation(value = "Enroute prisoner movement details.", notes = "Enroute to reception", nickname = "getEnrouteOffenderMovements")
     @ApiResponses(value = {
-            @ApiResponse(code = 200, message = "OK", response = MovementCount.class, responseContainer = "List"),
```
petergphillips, line 67: quite a few of these - where the swagger docs are plain wrong and wrong type is returned

src/main/java/net/syscon/elite/api/resource/BookingResource.java
```diff
@@ -80,7 +79,7 @@
     @GetMapping
     @ApiOperation(value = "Summary information for all offenders accessible to current user.", notes = "Summary information for all offenders accessible to current user.", nickname = "getOffenderBookings")
     @ApiResponses(value = {
-            @ApiResponse(code = 200, message = "Summary information for all offenders accessible to current user.", response = OffenderBooking.class, responseContainer = "List"),
+            @ApiResponse(code = 200, message = "Summary information for all offenders accessible to current user."),
```
petergphillips, line 82: spring fox generates the type automatically so better not to specify as it then gets out of sync

PR #495, opened 2020-04-30
Nn 2591 add property dto

Nimphal:


src/main/java/net/syscon/elite/service/transformers/PropertyContainerTransformer.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Simple transform object to help reduce boiler plate mapping of entity {@link OffenderCourtCase} to model object {@link CourtCase}.
+ */
+public class PropertyContainerTransformer {
+
+    public static PropertyContainer transform(final OffenderPropertyContainer propertyContainer) {
+        return PropertyContainer.builder()
+                .location(Location.builder()
```
Mjwillis, line 23: Think there is already a transformer for locations ?

Nimphal: There is, just in a different place. I've updated :)

PR #496, opened 2020-04-30
DT-793 - Provide more information about an offender in one call

Mjwillis:
This is support enhanced data set for Elastic search - hide the extra data request behind a new param
New endpoint for offender by offender no


src/main/java/net/syscon/elite/api/model/Alias.java
```diff
@@ -1,203 +1,57 @@
...
 
-import com.fasterxml.jackson.annotation.*;
+import com.fasterxml.jackson.annotation.JsonInclude;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
-import lombok.AllArgsConstructor;
-import lombok.Builder;
-import lombok.EqualsAndHashCode;
-import lombok.NoArgsConstructor;
+import lombok.*;
```
andrewrlee, line 6: think people have started to use the non wildcards - if everyone standardises on that then it leads to less churn

src/main/java/net/syscon/elite/api/model/Alias.java
```diff
@@ -1,203 +1,57 @@
...
     @NotBlank
+    @ApiModelProperty(required = true, value = "Last name of offender alias")
     private String lastName;
 
     @NotNull
+    @ApiModelProperty(required = true, value = "Age of Offender")
     private Integer age;
 
     @NotNull
+    @ApiModelProperty(required = true, value = "Date of Birth of Offender")
```
andrewrlee, line 40: examples?

src/main/java/net/syscon/elite/repository/impl/InmateRepositoryImpl.java
```diff
@@ -888,6 +890,24 @@ public void updateActiveCategoryNextReviewDate(final long bookingId, final Local
...
 
+    @Override
+    public Optional<ImprisonmentStatus> getImprisonmentStatus(final long bookingId) {
+        ImprisonmentStatus imprisonmentStatus;
+        try {
+            imprisonmentStatus = jdbcTemplate.queryForObject(
+                    getQuery("GET_IMPRISONMENT_STATUS"),
+                    createParams("bookingId", bookingId),
+                    IMPRISONMENT_STATUS_MAPPER);
+            if (imprisonmentStatus != null) {
```
andrewrlee, line 901: this seems a bit weird - when would it return null and when would it throw `EmptyResultDataAccessException`? 
are these both trying to handle not getting back data?

src/main/java/net/syscon/elite/api/model/InmateDetail.java
```diff
@@ -126,6 +122,30 @@
     @ApiModelProperty(value = "Country of birth", example = "GBR")
     private String birthCountryCode;
 
+    @ApiModelProperty(value = "In/Out Status", required = true, example = "IN", allowableValues = "IN,OUT")
```
andrewrlee, line 125: Is there some way that the swagger doc can show that these are all not going to be returned unless extra details are specified?  

(In REST terms they would probably be defined as different media types but theres quite a lot of overhead getting that to work and more burden on the client)

src/main/java/net/syscon/elite/api/model/Alias.java
Nimphal: IntelliJ has become such a pain with that, it autowildcards now after a certain amount, so you have to be vigilant :D I am still on the hunt for a setting to stop it from doing that.

andrewrlee: It's annoying that it doesn't have a unlimited - but you can set it to 999 and its per package I think.

src/main/java/net/syscon/elite/api/model/InmateDetail.java
andrewrlee: I see there's some doc on the flags, could maybe add a `note` on affected fields?  🤷‍♂️  not sure how thats represented in the generated swagger documentation

src/main/java/net/syscon/elite/api/model/ImprisonmentStatus.java
```diff
@@ -0,0 +1,20 @@
...
+
+    private Long bookingId;
+    private String bandCode;
+    private String imprisonmentStatus;
+    private String legalStatus;
+
+    public void deriveLegalStatus() {
+        if (this.bandCode != null) {
+            final var bandCode = Integer.parseInt(this.bandCode);
+            this.legalStatus = (bandCode <= 8 || bandCode == 11) ? "Convicted" : "Remand";
```
andymarke, line 16: Not sure if you want to enhance this now, but I think months ago we talked about deportation status that might be recorded incorrectly as  Convicted. Probably something for the future

src/main/java/net/syscon/elite/repository/impl/InmateRepositoryImpl.java
```diff
@@ -872,7 +899,7 @@ public void updateActiveCategoryNextReviewDate(final long bookingId, final Local
         if (result != 1) {
             var message = String.format("Unable to update next review date, could not find latest, active categorisation for booking id %d, result count = %d", bookingId, result);
             log.error(message);
-            throw new EntityNotFoundException(String.format(message));
+            throw new EntityNotFoundException(message);
```
matt-crowson-moj, line 902: Meh there is a factory method `withMessage` on the exception class...

src/main/java/net/syscon/elite/api/model/ImprisonmentStatus.java
```diff
@@ -0,0 +1,20 @@
...
+
+@Data
+public class ImprisonmentStatus {
+
+    private Long bookingId;
+    private String bandCode;
+    private String imprisonmentStatus;
+    private String legalStatus;
+
+    public void deriveLegalStatus() {
```
matt-crowson-moj, line 13: Just a thought (as you are relying on this being called).  Can you just not put in the getter (so replacing th@Data genereted on) so it derives up calling?

src/main/java/net/syscon/elite/api/model/ImprisonmentStatus.java
Mjwillis: Indeed

src/main/java/net/syscon/elite/repository/impl/InmateRepositoryImpl.java
```diff
@@ -892,19 +917,17 @@ public void updateActiveCategoryNextReviewDate(final long bookingId, final Local
 
     @Override
     public Optional<ImprisonmentStatus> getImprisonmentStatus(final long bookingId) {
-        ImprisonmentStatus imprisonmentStatus;
+        Optional<ImprisonmentStatus> imprisonmentStatus;
         try {
-            imprisonmentStatus = jdbcTemplate.queryForObject(
+            imprisonmentStatus = Optional.ofNullable(jdbcTemplate.queryForObject(
```
andrewrlee, line 922: so it can return null AND it can throw EmptyResultDataAccessException?

andrewrlee: I would have thought it would be one or the other

PR #498, opened 2020-05-01
Nn 2593 add extra information offender address

Nimphal:
Add the start date for an offender address, as well as the associated phone number.


src/main/resources/sqls/OffenderAddressRepositoryImpl.sql
```diff
@@ -11,11 +11,16 @@ GET_OFFENDER_ADDRESSES {
...
+          ADDR.NO_FIXED_ADDRESS_FLAG AS NO_FIXED_ADDRESS,
+          ADDR.START_DATE,
+          PH.PHONE_TYPE,
+          PH.PHONE_NO,
+          PH.EXT_NO AS PHONE_EXT
    FROM ADDRESSES ADDR
      LEFT JOIN REFERENCE_CODES RC1 ON RC1.CODE = ADDR.COUNTY_CODE AND RC1.DOMAIN = 'COUNTY'
      LEFT JOIN REFERENCE_CODES RC2 ON RC2.CODE = ADDR.COUNTRY_CODE AND RC2.DOMAIN = 'COUNTRY'
      LEFT JOIN REFERENCE_CODES RC3 ON RC3.CODE = ADDR.CITY_CODE AND RC3.DOMAIN = 'CITY'
+     LEFT JOIN PHONES PH ON PH.OWNER_CLASS = 'ADDR' AND PH.OWNER_ID = ADDR.ADDRESS_ID
```
Mjwillis, line 23: Don't you get multiple phone types for each address?

andymarke: So looks like each address would be repeated for each number, not sure how you could this as JOIN, any ideas @Mjwillis

Nimphal: It looks like there may be a case for a separate call to collect the phone numbers, rather than part of this query. Will close this PR while working on it :)

PR #505, opened 2020-05-05
Nn 2594 get address for person

Nimphal:
Creates 3 new endpoints in the person namespace, returning a list of addresses and associated phone numbers, a list of phone numbers associated to the person and a list of emails associated to the person.

Also renames OffenderAddress to AddressDto as it is not offender specific and is being reused for the new endpoints.


src/main/java/net/syscon/elite/api/resource/OffenderResource.java
```diff
@@ -71,11 +71,11 @@
     @GetMapping("/{offenderNo}/addresses")
     @ApiOperation(value = "Return a list of addresses for a given offender, most recent first.")
     @ApiResponses(value = {
-            @ApiResponse(code = 200, message = "OK", response = OffenderAddress.class, responseContainer = "List"),
+            @ApiResponse(code = 200, message = "OK", response = AddressDto.class, responseContainer = "List"),
```
matt-crowson-moj, line 74: I might be wrong on this one but I don't think we need to add the 200 response docs anymore. @petergphillips ???

src/test/java/net/syscon/elite/service/PersonServiceTest.java
```diff
@@ -0,0 +1,206 @@
...
+    @Mock
+    private PhoneRepository phoneRepository;
+
+    @Mock
+    private InternetAddressRepository internetAddressRepository;
+
+    private PersonService personService;
+
+    @BeforeEach
+    public void setUp() {
```
matt-crowson-moj, line 47: If using junit 5 (Jupiter) no need for `public` methods anymore...

src/main/java/net/syscon/elite/repository/jpa/model/InternetAddress.java
```diff
@@ -0,0 +1,36 @@
...
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+@Entity
+@Table(name = "INTERNET_ADDRESSES")
+public class InternetAddress {
+    @Id
+    @Column(name = "INTERNET_ADDRESS_ID")
```
matt-crowson-moj, line 21: Probably pay to add `nullable = false` to the @Column annotations for those fields that are not nullable (defaults to true).  I appreciate we are not technically generating the schema as the NOMIS DB already exists.

src/main/java/net/syscon/elite/api/resource/OffenderResource.java
matt-crowson-moj: You could take it off and just start up the app and go the docs on localhost...

Nimphal: Good catch, missed this one in the rename!

src/main/java/net/syscon/elite/repository/jpa/model/InternetAddress.java
Nimphal: sounds like a good thing to document anyway, have added it to the new models

PR #510, opened 2020-05-11
NN-2596 expose secondary languages

JamieDraperUK:
- Expose secondary languages
- Move some of the language code to the JPA model


src/main/java/net/syscon/elite/service/InmateService.java
```diff
@@ -683,6 +687,24 @@ private void validate(final CategoryRejectionDetail detail) {
         return repository.findInmateAliases(bookingId, defaultOrderBy, sortOrder, offset, limit);
     }
 
+    @VerifyBookingAccess
+    public List<SecondaryLanguage> getSecondaryLanguages(final Long bookingId) {
+        return offenderLanguageRepository
+                .findByOffenderBookId(bookingId)
+                .stream()
+                .filter(lang -> lang.getType().equalsIgnoreCase("SEC"))
```
brightonsbox, line 695: Could `lang.getType()` be null?

matt-crowson-moj: Would suggest this is reversed perhaps...

`"SEC".equalsIgnoreCase(lang.getType())`

src/main/java/net/syscon/elite/service/InmateService.java
```diff
@@ -683,6 +687,24 @@ private void validate(final CategoryRejectionDetail detail) {
...
+                .map(lang -> SecondaryLanguage
+                        .builder()
+                        .bookingId(lang.getOffenderBookId())
+                        .code(lang.getCode())
+                        .description(lang.getReferenceCode() != null ? lang.getReferenceCode().getDescription() : null)
+                        .canRead(lang.getReadSkill() != null && lang.getReadSkill().equalsIgnoreCase("Y"))
```
matt-crowson-moj, line 701: As above perhaps to get rid of != null check...meh.

`"Y".equalsIgnoreCase(...)`

src/main/java/net/syscon/elite/repository/jpa/model/OffenderLanguage.java
```diff
@@ -0,0 +1,68 @@
...
+    @Id
+    private Long offenderBookId;
+
+    @Id
+    private String type;
+
+    @Id
+    private String code;
+
+    private String readSkill;
```
matt-crowson-moj, line 57: So kinda on the fence how to represent nullable fields on entities.  So one thing I have done is override getters and return optional's.   I accept this not ideal but just a bit cleaner when dealing with in other places.   If only this were 'Kotlin'....

src/test/java/net/syscon/elite/repository/jpa/repository/OffenderLanguageRepositoryTest.java
```diff
@@ -0,0 +1,39 @@
...
+@ActiveProfiles("test")
+@DataJpaTest
+@AutoConfigureTestDatabase(replace = NONE)
+public class OffenderLanguageRepositoryTest {
+
+    @Autowired
+    private OffenderLanguageRepository repository;
+
+    @Test
+    public void testGetLanguages() {
```
matt-crowson-moj, line 23: Just in case you didn't know no need for public methods with Junit 5.  Personally I am not sure what benefit this adds...

src/main/java/net/syscon/elite/service/InmateService.java
JamieDraperUK: Looks much cleaner, thanks!

src/test/java/net/syscon/elite/api/resource/impl/BookingResourceTest.java
```diff
@@ -372,4 +372,13 @@ public void getOffenceHistory() {
 
         assertThatJsonFileAndStatus(response, 200, "offender_main_offences.json");
     }
+
+    @Test
```
JamieDraperUK, line 376: @matt-crowson-moj

PR #512, opened 2020-05-11
correct syntax issue, move dnspolicy

mattops:


mattops:
> LGTM, although I can't see what the difference is (looks to be at the same indentation level?)

It's because of the `include` immediately after where it was, the `include` was a continuation of the container block.

PR #519, opened 2020-05-15
DT-820 new endpoint to enable cancellation of a scheduled prison to prison move

matt-crowson-moj:
New endpoint to enable the cancellation of a prison to prison move in NOMIS.


src/main/java/net/syscon/elite/service/PrisonToPrisonMoveSchedulingService.java
```diff
@@ -159,16 +159,27 @@ public void cancel(final Long bookingId, final Long scheduledMoveId, final Strin
...
+        move.setEventStatus(cancelled);
         move.setCancellationReason(transferCancellationReasonRepository.findById(TransferCancellationReason.pk(transferCancellationReasonCode)).orElseThrow(() -> EntityNotFoundException.withMessage("Cancellation reason %s not found.", transferCancellationReasonCode)));
 
         scheduleRepository.save(move);
 
         log.debug("Cancelled scheduled prison to prison move with id {} for offender {}", move.getId(), move.getOffenderBooking().getOffender().getNomsId());
     }
 
+    private boolean is(final OffenderIndividualSchedule move, final EventStatus status) {
+        return move.getEventStatus().equals(status);
```
brightonsbox, line 180: Will the move always have an event status? If these are enums, could use `==` instead?

matt-crowson-moj: Yes will always have an event status and not an enum.  Reference data.

PR #523, opened 2020-05-18
DCS-488 Upgrading dependencies

andrewrlee:


build.gradle
```diff
@@ -115,12 +114,12 @@ dependencies {
 
     implementation 'org.springframework.boot:spring-boot-starter-oauth2-resource-server'
 
-    implementation 'com.nimbusds:nimbus-jose-jwt:8.16' // pinned due to CVE-2019-17195 in 7.8.1 brought in by spring-boot-starter-oauth2-resource-server-2.2.5.RELEASE
+    implementation 'com.nimbusds:nimbus-jose-jwt:8.17'
```
petergphillips, line 117: does this still need to be pinned?

build.gradle
```diff
@@ -115,12 +114,12 @@ dependencies {
...
 
     implementation 'org.hibernate:hibernate-java8:5.4.10.Final'
```
petergphillips, line 119: can be removed now too - pinning to an old hibernate artifact.  Should be changed to `hibernate-core`, which can then be bumped - or even better removed as included in data-jpa

build.gradle
```diff
@@ -171,14 +170,14 @@ dependencies {
     testImplementation 'com.tngtech.java:junit-dataprovider:1.13.1'
     testImplementation 'net.javacrumbs.json-unit:json-unit-assertj:2.17.0'
 
-    testImplementation 'net.serenity-bdd:serenity-core:2.1.8'
-    testImplementation 'net.serenity-bdd:serenity-junit:2.1.8'
-    testImplementation 'net.serenity-bdd:serenity-spring:2.1.8'
+    testImplementation 'net.serenity-bdd:serenity-core:2.2.7'
```
petergphillips, line 173: I tried upgrading this and it broke everything

petergphillips: I even raised https://dsdmoj.atlassian.net/browse/DT-666 to fix

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.4.1-bin.zip
```
petergphillips, line 3: intellij prefers the `-all` and will moan if you use the `-bin` version I think

build.gradle
```diff
@@ -115,12 +114,11 @@ dependencies {
 
     implementation 'org.springframework.boot:spring-boot-starter-oauth2-resource-server'
 
-    implementation 'com.nimbusds:nimbus-jose-jwt:8.16' // pinned due to CVE-2019-17195 in 7.8.1 brought in by spring-boot-starter-oauth2-resource-server-2.2.5.RELEASE
+    implementation 'com.nimbusds:nimbus-jose-jwt'
```
petergphillips, line 117: sorry I more meant that nimbus will be included authomatically so don't think you need that line at all?

andrewrlee: removed it

build.gradle
andrewrlee: I've reverted this.. would be nice to get rid entirely but obviously a ton of work. 
We can't move to junit 5 until it goes

build.gradle
andrewrlee: removed

gradle/wrapper/gradle-wrapper.properties
andrewrlee: changed back to `all`. I don't think I've ever seen it moan apart from in plugin development

PR #530, opened 2020-05-21
DCS-488 Breaking the dependency cycle by excluding DataSourceAutoConfiguration

brightonsbox:


src/main/java/net/syscon/elite/web/config/PersistenceConfigs.java
```diff
@@ -16,6 +22,7 @@
 @Configuration
 @EnableAspectJAutoProxy
 @EnableTransactionManagement
+@EnableAutoConfiguration(exclude = { DataSourceAutoConfiguration.class })
```
brightonsbox, line 25: This is the only extra addition

PR #532, opened 2020-05-21
Nn 2599 cell search part 1

Nimphal:
Adds tables, data and repositories. Service and resource to follow in a separate PR.


src/main/java/net/syscon/elite/repository/jpa/model/LivingUnit.java
```diff
@@ -0,0 +1,157 @@
...
+    public static class PK implements Serializable {
+
+        @Column(name = "LIVING_UNIT_ID", insertable = false, updatable = false, nullable = false)
+        private Long livingUnitId;
+
+        @Column(name = "DESCRIPTION", nullable = false)
+        private String description;
+
+        @Column(name = "AGY_LOC_ID", nullable = false)
+        private String agencyLocationId;
```
Nimphal, line 43: No table primary key, so being a bit creative.

src/main/java/net/syscon/elite/repository/jpa/model/LivingUnitProfile.java
```diff
@@ -0,0 +1,74 @@
...
+
+        @Column(name = "DESCRIPTION")
+        private String description;
+
+        @Column(name = "AGY_LOC_ID")
+        private String agencyLocationId;
+
+        @Column(name = "PROFILE_ID")
+        private Long profileId;
+
```
Nimphal, line 47: Same as above.

src/main/java/net/syscon/elite/repository/jpa/model/LivingUnit.java
```diff
@@ -0,0 +1,157 @@
...
+    @AllArgsConstructor
+    public static class PK implements Serializable {
+
+        @Column(name = "LIVING_UNIT_ID", insertable = false, updatable = false, nullable = false)
+        private Long livingUnitId;
+
+        @Column(name = "DESCRIPTION", nullable = false)
+        private String description;
+
+        @Column(name = "AGY_LOC_ID", nullable = false)
```
matt-crowson-moj, line 42: We have the models in place to support this already:

```
    @ManyToOne(optional = false)
    @JoinColumn(name = "AGY_LOC_ID", nullable = false)
    private AgencyLocation agencyLocationId;

```

src/main/java/net/syscon/elite/repository/jpa/model/LivingUnit.java
```diff
@@ -0,0 +1,157 @@
...
+    private String lowestLevelFlag;
+
+    @Column(name = "REACH_OPER_CAPACITY_FLAG")
+    private String reachedOperationalCapacityFlag;
+
+    @Column(name = "NO_OF_OCCUPANT")
+    private Integer noOfOccupants;
+
+    public boolean isActive() {
+        return activeFlag.equals("Y");
```
matt-crowson-moj, line 142: Not really a big deal (as `activeFlag` should not be null) but I find it better to do something like: `"Y".equals(activeFlag)` this helps with preventing nullpointers.

src/main/java/net/syscon/elite/repository/jpa/model/LivingUnit.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    public boolean isActiveCell() {
+        return isActive() && livingUnitType != null && livingUnitType.equals("CELL");
```
matt-crowson-moj, line 146: Same hear as above (avoid the need for the null check):

`return isActive() && "CELL".equals(livingUnitType);`

src/main/java/net/syscon/elite/repository/jpa/model/LivingUnitProfile.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Id
+    @Column(name = "LIVING_UNIT_ID", nullable = false)
+    private Long livingUnitId;
+
+    @Id
+    @Column(name = "AGY_LOC_ID", nullable = false)
+    private String agencyLocationId;
```
matt-crowson-moj, line 56: See agency comment above.  We have an object to model this.

src/main/java/net/syscon/elite/repository/jpa/model/LivingUnit.java
```diff
@@ -0,0 +1,157 @@
...
+import static org.hibernate.annotations.NotFoundAction.IGNORE;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+@Entity
+@IdClass(LivingUnit.PK.class)
+@Table(name = "LIVING_UNITS")
+public class LivingUnit {
```
matt-crowson-moj, line 30: As a general comment do you need to map all of the fields here?   I tend to only map what is needed.

src/main/java/net/syscon/elite/repository/jpa/model/LivingUnit.java
Nimphal: In this case the agency location column is part of the composite primary key, so I don't know if it's possible to use the model instead. Also don't need the agency information as part of the object, purely used for filtering.

src/main/java/net/syscon/elite/repository/jpa/model/LivingUnitProfile.java
Nimphal: Same as above.

src/main/java/net/syscon/elite/repository/jpa/model/LivingUnit.java
matt-crowson-moj: Just realised I put this on the wrong line anyway. Meant for line 53 which I think would be OK.  Not matter.

PR #539, opened 2020-05-27
Nn 2635 return offender official contacts

Nimphal:
Extends current offender contacts API to additionally return non next of kin contacts (docs suggest that was the intention but does not seem to have been implemented)


src/test/java/net/syscon/elite/executablespecification/ContactStepDefinitions.java
```diff
@@ -23,6 +23,7 @@
 
     @When("^contact details with booking id ([0-9-]+) is requested$")
     public void sentenceWithBookingId(final Long id) {
+        System.out.println();
```
mikehalmamoj, line 26: Debugging? :)

Nimphal: :scream:

PR #540, opened 2020-05-27
NN-2642 expose more offender language data

JamieDraperUK:


src/main/java/net/syscon/elite/api/model/InmateDetail.java
```diff
@@ -69,6 +69,12 @@
     @ApiModelProperty(value = "The prisoner's preferred spoken language.")
     private String language;
 
+    @ApiModelProperty(value = "The prisoner's requires an interpreter")
```
Nimphal, line 72: meh, typo in doc

src/main/java/net/syscon/elite/api/model/InmateDetail.java
```diff
@@ -69,6 +69,12 @@
...
+    private Boolean interpreterRequired;
+
+    @ApiModelProperty(value = "The prisoner's preferred written language")
```
Nimphal, line 75: same as above

src/main/java/net/syscon/elite/service/InmateService.java
```diff
@@ -259,13 +268,22 @@ public InmateDetail findInmate(final Long bookingId, final boolean extraInfo) {
...
-    private Optional<String> getFirstPreferredSpokenLanguage(final Long bookingId) {
+    private Optional<OffenderLanguage> getFirstPreferredSpokenLanguage(final Long bookingId) {
         return offenderLanguageRepository
                 .findByOffenderBookId(bookingId)
                 .stream()
                 .filter(l -> "PREF_SPEAK".equals(l.getType()) && l.getReferenceCode() != null)
-                .map(l -> l.getReferenceCode().getDescription())
-                .max(Comparator.naturalOrder());
+                .sorted(Comparator.comparing(right -> right.getReferenceCode().getDescription()))
+                .reduce((first, second) -> second);
```
brightonsbox, line 277: Is the method name a bit misleading now?
As far as I understand, this filters for 'PREF_SPEAK', orders by description alphabetically and picks the last?

JamieDraperUK: The behaviour is the same as before, only now it's returning the whole object instead of the description. Not why why it was named like that previously or why we order then take the last but thought it best to keep it the same.

PR #549, opened 2020-06-01
GDPR-110 Adding check on duplicate LIDS numbers

brightonsbox:


src/main/java/uk/gov/justice/hmpps/nomis/datacompliance/repository/jpa/repository/DuplicateOffenderRepository.java
```diff
@@ -51,4 +51,22 @@
                     ")",
             nativeQuery = true)
     List<DuplicateOffender> getOffendersWithMatchingCroNumbers(String offenderNo, Set<String> formattedIds);
+
+    /**
+     * The following query finds offenders that share the same
+     * LIDS booking number.
+     */
```
mikehalmamoj, line 58: The method name is pretty good so maybe don't need the comment?

PR #554, opened 2020-06-03
NN-2649 Only return address information for active bookings

JamieDraperUK:


src/main/java/net/syscon/elite/service/OffenderAddressService.java
```diff
@@ -22,13 +22,17 @@
...
-    private final OffenderRepository offenderRepository;
+    private final OffenderBookingRepository offenderBookingRepository;
     private final AddressRepository addressRepository;
     private final PhoneRepository phoneRepository;
 
     @VerifyOffenderAccess
     public List<AddressDto> getAddressesByOffenderNo(@NotNull String offenderNo) {
-        final var offenderRootId = offenderRepository.findByNomsId(offenderNo).getRootOffenderId();
+        final var offenderBookings = offenderBookingRepository.findByOffenderNomsIdAndActiveFlag(offenderNo, "Y");
+        if(offenderBookings.size() > 1) throw new RuntimeException(String.format("More than one active booking was returned for offender number %s\n", offenderNo));
```
matt-crowson-moj, line 32: Kinda feels like this could be an `IllegalStateException` and not just plain old `Runtime`....meh

JamieDraperUK: Agreed. Raised other PR to fix it. Cheers Matt.

PR #559, opened 2020-06-04
NN-2663: Oracle doesn't like columns longer than 30 chars

Nimphal:


Nimphal:
The test database doesn't have this problem, which is why it wasn't caught. The endpoints themselves have tests.

PR #560, opened 2020-06-04
DT-893 new service to support rescheduling of court hearing dates.

matt-crowson-moj:
This service is support changes to court hearings dates and times.

It is quite common as court hearings get nearer the dates and/or times may need to be shuffled around.


src/main/java/net/syscon/elite/repository/jpa/model/CourtEvent.java
```diff
@@ -110,6 +110,11 @@ public LocalDateTime getEventDateTime() {
         return eventDate.atTime(startTime.toLocalTime());
     }
 
+    public void setEventDateTime(final LocalDateTime dateTime) {
```
matt-crowson-moj, line 113: Convenience method.  NOMIS has this as two fields on the DB.

PR #562, opened 2020-06-04
DT-894 changes for wiring endpoint to service layer for rescheduling court hearings.

matt-crowson-moj:


src/main/java/net/syscon/elite/api/resource/OffenderMovementsResource.java
```diff
@@ -99,4 +99,14 @@ ScheduledPrisonToPrisonMove prisonToPrison(@ApiParam(value = "The offender booki
     ResponseEntity<Void> cancelPrisonToPrisonMove(@ApiParam(value = "The offender booking to associate the prison to prison move with.", required = true) @PathVariable("bookingId") Long bookingId,
                                                   @ApiParam(value = "The scheduled event identifier for the prison to prison.", required = true) @PathVariable("eventId") Long eventId,
                                                   @ApiParam(value = "The cancellation details.", required = true) @RequestBody @Valid PrisonMoveCancellation cancellation);
+
+    @PutMapping("/{bookingId}/prison-to-court-hearings/{hearingId}/revised-hearing-date/{revisedDateTime}")
```
andymarke, line 103: I've never been an expert on Rest but would revisedDateTime be better as a body rather than as part of the path? Since you are "putting" a court hearing rather than a date?
Do we have any other examples in elite for this?

src/test/java/net/syscon/elite/api/resource/impl/OffenderMovementsResourceImplIntTest_rescheduleCourtHearing.java
```diff
@@ -0,0 +1,116 @@
...
+                ErrorResponse.class);
+
+        assertThat(response.getBody()).isEqualTo(
+                ErrorResponse.builder()
+                        .status(400)
+                        .userMessage("The existing court hearing '-209' must be in a scheduled state to reschedule.")
+                        .developerMessage("The existing court hearing '-209' must be in a scheduled state to reschedule.")
+                        .build());
+    }
+}
```
andymarke, line 116: Maybe add a test to create a court hearing then amend it (and do you have a cancel now? If so cancel it) Nice to see full CRUD in a single test

src/main/java/net/syscon/elite/api/resource/OffenderMovementsResource.java
andrewrlee: I think body is better - URIs are for unique addresses of a resource. 
The resource you are modifying is the date field not the date value

src/test/java/net/syscon/elite/api/resource/impl/OffenderMovementsResourceImplIntTest_rescheduleCourtHearing.java
matt-crowson-moj: No there isn't any ability to cancel a court hearing.  There is for scheduled movements but that is a different entity in NOMIS e.g. prison to prison move.  You would think the same for prison to court but alas no.

src/test/java/net/syscon/elite/api/resource/impl/OffenderMovementsResourceImplIntTest_rescheduleCourtHearing.java
```diff
@@ -0,0 +1,141 @@
...
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.in;
+import static org.springframework.http.HttpStatus.OK;
+
+public class OffenderMovementsResourceImplIntTest_rescheduleCourtHearing extends ResourceTest {
+
+    @Test
+    public void reschedule_court_hearing() {
```
matt-crowson-moj, line 21: FAO @andymarke expanded the test to schedule a hearing to be amended/revised on the the hearing date/

src/main/java/net/syscon/elite/api/resource/OffenderMovementsResource.java
```diff
@@ -99,4 +100,14 @@ ScheduledPrisonToPrisonMove prisonToPrison(@ApiParam(value = "The offender booki
     ResponseEntity<Void> cancelPrisonToPrisonMove(@ApiParam(value = "The offender booking to associate the prison to prison move with.", required = true) @PathVariable("bookingId") Long bookingId,
                                                   @ApiParam(value = "The scheduled event identifier for the prison to prison.", required = true) @PathVariable("eventId") Long eventId,
                                                   @ApiParam(value = "The cancellation details.", required = true) @RequestBody @Valid PrisonMoveCancellation cancellation);
+
+    @PutMapping("/{bookingId}/prison-to-court-hearings/{hearingId}/revised-hearing-date")
```
andrewrlee, line 104: `/{bookingId}/prison-to-court-hearings/{hearingId}/hearing-date` ?

PR #565, opened 2020-06-08
NN-2701 fix address service bug

JamieDraperUK:


src/main/java/net/syscon/elite/service/OffenderAddressService.java
```diff
@@ -29,9 +29,10 @@
     @VerifyOffenderAccess
     public List<AddressDto> getAddressesByOffenderNo(@NotNull String offenderNo) {
         final var offenderBookings = offenderBookingRepository.findByOffenderNomsIdAndActiveFlag(offenderNo, "Y");
-        if(offenderBookings.size() > 1) throw new IllegalStateException(String.format("More than one active booking was returned for offender number %s\n", offenderNo));
+        if (offenderBookings.size() > 1)
+            throw new IllegalStateException(String.format("More than one active booking was returned for offender number %s\n", offenderNo));
 
-        final var offenderBooking = offenderBookings.stream().findFirst().orElseThrow(EntityNotFoundException.withMessage(String.format("No active offender bookings found for offender number %s\n",offenderNo)));
+        final var offenderBooking = offenderBookings.stream().findFirst().orElseThrow(EntityNotFoundException.withMessage(String.format("No active offender bookings found for offender number %s\n", offenderNo)));
         final var offenderRootId = offenderBooking.getOffender().getRootOffenderId();
```
Nimphal, line 36: could the offender of the booking be null? probably not, but is it worth checking?

JamieDraperUK: No it can't. The mapping requires it.

JamieDraperUK: `    @ManyToOne(optional = false)
    @JoinColumn(name = "OFFENDER_ID", nullable = false)
    private Offender offender;
`

src/main/java/net/syscon/elite/service/OffenderAddressService.java
```diff
@@ -60,7 +61,7 @@
                                             .addressId(address.getAddressId())
                                             .activeFlag("Y".equalsIgnoreCase(addressUsage.getActiveFlag()))
                                             .addressUsage(addressUsage.getAddressUsage())
-                                            .addressUsageDescription(addressUsage.getAddressUsageType().getDescription())
+                                            .addressUsageDescription(addressUsage.getAddressUsageType() == null ? null : addressUsage.getAddressUsageType().getDescription())
```
matt-crowson-moj, line 64: I know Lombok helps to reduce boiler plate but I think it would really pay to use the Optional feature in Java more on some of our JPA mappings.  Of course this would mean adding a getter to override the Lombok generated one but worth I think.

PR #568, opened 2020-06-09
NN-2695: Add address usages to person address

Nimphal:


src/main/java/net/syscon/elite/service/PersonService.java
```diff
@@ -54,6 +56,14 @@
                     .startDate(address.getStartDate())
                     .endDate(address.getEndDate())
                     .street(address.getStreet())
+                    .addressUsages(address.getAddressUsages() == null ? null : address.getAddressUsages().stream()
```
matt-crowson-moj, line 59: I am not sure the modelling here is right.

The DB does not allow this value to be null so the null check seems wrong.  I'd also go as far to say Collection return types should return empty lists/sets etc so the null checks should not be needed.

I think Lombok is not entirely intuitive here also.  You can ensure the Collection is initialised in a couple of ways.  Declare the `@Singular` annotation on the field or just initialise with `private final List<X> x = new ArrayList();`

Nimphal: The value must have been null somewhere in live as this in another part of the codebase resulted in a bunch of 500 errors yesterday. I agree about initialising the collection, a good thing to fix in a future PR.

PR #569, opened 2020-06-09
NN-2695: Return address type description instead of code

Nimphal:


src/main/java/net/syscon/elite/service/OffenderAddressService.java
```diff
@@ -55,7 +56,7 @@
                     .startDate(address.getStartDate())
                     .endDate(address.getEndDate())
                     .street(address.getStreet())
-                    .addressUsages(address.getAddressUsages() == null ? null : address.getAddressUsages().stream()
+                    .addressUsages(address.getAddressUsages().stream()
```
Nimphal, line 59: This is initiated to an empty list, no need to check for null anymore.

src/main/java/net/syscon/elite/service/PersonService.java
```diff
@@ -56,7 +57,7 @@
                     .startDate(address.getStartDate())
                     .endDate(address.getEndDate())
                     .street(address.getStreet())
-                    .addressUsages(address.getAddressUsages() == null ? null : address.getAddressUsages().stream()
+                    .addressUsages(address.getAddressUsages().stream()
```
Nimphal, line 60: Same as above.

src/main/java/net/syscon/elite/repository/jpa/model/Address.java
```diff
@@ -43,8 +45,13 @@
 
     private String flat;
 
-    @Column(name = "ADDRESS_TYPE")
-    private String addressType;
+    @ManyToOne
```
mikehalmamoj, line 48: Should this be FetchType.LAZY?

Nimphal: Oooh, I thought it was lazy by default? We don't seem to specify it in most places.

mikehalmamoj: Apparently only for @OneToMany - mentioned in this thread on Slack https://mojdt.slack.com/archives/C012FNNQ4A3/p1591606149012500

Nimphal: that comment mentions about always resolving the data, I think that's appropriate for reference codes as we'd always want to have the description

mikehalmamoj: I think there's only a problem where some entity data does is not included on the DTO - queries running for no reason.

Mjwillis: Generally we don't want EAGER because JPA will always join to these tables and we might not be interested in them.  It's better to let the application load what it needs.  The caveat on that is that a big list might want to join to one extra table (say its parent) for example if we were selecting a 2000 bookings we might want to inner join to offenders table because we always need their name.  However JPA does allow you to explicitly join in its query language on specific repo calls so better to leave the default to LAZY

PR #575, opened 2020-06-12
DT-901 changes to support cancellation (removal) of court hearings.

matt-crowson-moj:


src/main/java/net/syscon/elite/service/CourtHearingCancellationService.java
```diff
@@ -0,0 +1,34 @@
...
+import net.syscon.elite.security.VerifyBookingAccess;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+import org.springframework.validation.annotation.Validated;
+
+@Service
+@Transactional(readOnly = true)
+@Validated
+@Slf4j
+public class CourtHearingCancellationService {
```
matt-crowson-moj, line 15: I realise this is refers to **cancellation** but we are actually deleting.  Happy to rename...

src/main/java/net/syscon/elite/repository/jpa/model/CourtEvent.java
```diff
@@ -115,19 +126,48 @@ public void setEventDateTime(final LocalDateTime dateTime) {
...
+    }
 
-            offenderCourtCase.getCharges().stream().filter(OffenderCharge::isActive).forEach(this::addPreExisting);
+    static class OnCreate {
+        /**
+         * On court event creation on NOMIS existing active charges are applied to the new court event if associated with a court case.
+         */
+        @PrePersist
+        void applyActiveCourtCaseChargesFor(final CourtEvent event) {
+            event.getOffenderCourtCase().ifPresent(courtCase -> {
```
andymarke, line 139: So interestingly this will probably cause a SELECT assuming this is lazy loaded - which I'm sure is OK

src/main/java/net/syscon/elite/repository/jpa/model/CourtEvent.java
```diff
@@ -115,19 +126,48 @@ public void setEventDateTime(final LocalDateTime dateTime) {
...
-                .offenderCharge(charge)
-                .courtEvent(this)
-                .build());
+    @Component
+    static class OnDelete {
+
+        private Clock clock;
+
+        @Autowired
+        void setClock(final Clock clock) {
```
andymarke, line 161: Nice that this works, the auto-wiring

src/main/java/net/syscon/elite/repository/jpa/model/CourtEvent.java
```diff
@@ -115,19 +126,48 @@ public void setEventDateTime(final LocalDateTime dateTime) {
...
+            this.clock = clock;
+        }
+
+        @PreRemove
+        void checkIsScheduledFutureEventPriorToRemovalOf(final CourtEvent event) {
+            Objects.requireNonNull(clock, "Clock not set.");
+
+            checkState(event.getEventDateTime().isAfter(LocalDateTime.now(clock)), "Court hearing '%s' cannot be deleted as its start date/time is in the past.", event.getId());
+            checkState(event.getEventStatus().isScheduled(), "Court hearing '%s' must be in a scheduled state to delete.", event.getId());
```
andymarke, line 170: What exception do you end up getting when the state isn't satisfied?

src/main/java/net/syscon/elite/service/CourtHearingCancellationService.java
mikehalmamoj: I think it's fine, in future cancellation might switch to e.g. a change of status but consumers of the service shouldn't care.

src/main/java/net/syscon/elite/repository/jpa/model/CourtEvent.java
matt-crowson-moj: You do get this exception but is wrapped in another exception on instance type DataAccessException.  We should be able to deal with this in higher layers.

src/test/java/net/syscon/elite/repository/jpa/repository/CourtEventRepositoryTest.java
```diff
@@ -149,4 +181,55 @@ void court_event_without_court_case_retrieved() {
...
+        final var savedCourtEventWithCourtCase = courtEventRepository.save(builder
+                .eventDate(LocalDate.now(clock).minusDays(1))
+                .build());
+
+        entityManager.flush();
+
+        final var id = savedCourtEventWithCourtCase.getId();
+
+        assertThatThrownBy(() -> courtEventRepository.deleteById(id))
+                .isInstanceOf(DataAccessException.class)
```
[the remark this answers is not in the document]
matt-crowson-moj: Ultimately I think so yes...

PR #586, opened 2020-06-18
DT-929 adding permissions/role check for create, update and deletion of court hearings.

matt-crowson-moj:
As part of discussions with the NART team it was decided the permissions to create. retrieve and update court hearings needs to be hardened.

### DIFF
--- a/src/main/java/net/syscon/elite/service/CourtHearingCancellationService.java
+++ b/src/main/java/net/syscon/elite/service/CourtHearingCancellationService.java
@@ -4,6 +4,7 @@ import lombok.extern.slf4j.Slf4j;
 import net.syscon.elite.core.HasWriteScope;
 import net.syscon.elite.repository.jpa.repository.CourtEventRepository;
 import net.syscon.elite.security.VerifyBookingAccess;
+import org.springframework.security.access.prepost.PreAuthorize;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.validation.annotation.Validated;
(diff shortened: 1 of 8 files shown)


src/main/java/net/syscon/elite/service/CourtHearingCancellationService.java
```diff
@@ -23,6 +24,7 @@ public CourtHearingCancellationService(final CourtEventRepository courtEventRepo
     @Transactional
     @VerifyBookingAccess
     @HasWriteScope
+    @PreAuthorize("hasRole('PECS_PRISON')")
```
andymarke, line 27: Very minor should this be something like 'COURT_HEARING_MAINTAINER' which is assigned to the PECS system credentials

matt-crowson-moj: Hmmm not sure as this what came about talking with @Mjwillis

matt-crowson-moj: Updated.

src/test/java/net/syscon/elite/api/resource/impl/OffenderMovementsResourceImplIntTest_cancelCourtHearing.java
```diff
@@ -42,7 +42,7 @@ public Clock clock() {
 
     @Before
     public void setup() {
-        token = authTokenHelper.getToken(AuthTokenHelper.AuthToken.NORMAL_USER);
+        token = authTokenHelper.getToken(AuthTokenHelper.AuthToken.COURT_HEARING_MAINTAINER);
```
petergphillips, line 45: would be good to test the negation too i.e. I could delete all of the `PreAuthorize` lines from the code and all the tests would still pass

matt-crowson-moj: Good point.  Will add.

PR #593, opened 2020-06-22
Nn 2415 list of visits endpoint

Nimphal:


src/main/java/net/syscon/elite/service/BookingService.java
```diff
@@ -305,7 +319,7 @@ private long daysSinceDetailBecameEffective(final PrivilegeDetail currentDetail)
     }
 
     @VerifyBookingAccess
-    public Page<ScheduledEvent> getBookingActivities(final Long bookingId, final LocalDate fromDate, final LocalDate toDate, final long offset, final long limit, final String orderByFields, final Order order) {
+    public net.syscon.elite.api.support.Page<ScheduledEvent> getBookingActivities(final Long bookingId, final LocalDate fromDate, final LocalDate toDate, final long offset, final long limit, final String orderByFields, final Order order) {
```
Nimphal, line 322: I'm not 100% sure this is the best way to tackle this, but the Page implementation above conflicts with the Page implementation used for Pageable and JPA repository. They can't both me imported and Java doesn't allow renaming of imports. Happy for suggestions on how to tackle this better.

src/main/java/net/syscon/elite/api/model/VisitWithVisitors.java
```diff
@@ -0,0 +1,32 @@
...
+@ApiModel(description = "List of visitors for a visit")
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@Data
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+public class VisitWithVisitors<V extends Visit> {
+    @ApiModelProperty(value = "List of visitors on visit", required = true)
+    @JsonProperty("visitors")
+    @NotBlank
```
brightonsbox, line 25: Is @NotBlank correct here? I thought it was only applicable to Strings?

Nimphal: Very good spot, too much copy pasting. Fixed :)

src/main/java/net/syscon/elite/api/model/Visitor.java
```diff
@@ -0,0 +1,54 @@
...
+    private String lastName;
+
+    @ApiModelProperty(value = "First name of visitor")
+    @JsonProperty("firstName")
+    @NotBlank
+    private String firstName;
+
+    @ApiModelProperty(value = "Date of birth of visitor")
+    @JsonProperty("dateOfBirth")
+    @NotBlank
```
brightonsbox, line 42: You've got some more in this class too that just need converting to `NotNull`

Nimphal: Not anymore :) Thank you!

src/main/java/net/syscon/elite/api/resource/BookingResource.java
```diff
@@ -507,6 +511,19 @@ ReasonableAdjustments getReasonableAdjustments(@ApiParam(value = "The offender b
                                                           @ApiParam(value = "Comma separated list of one or more of the following fields - <b>eventDate, startTime, endTime, eventLocation</b>") @RequestHeader(value = "Sort-Fields", required = false) String sortFields,
                                                           @ApiParam(value = "Sort order (ASC or DESC) - defaults to ASC.", defaultValue = "ASC") @RequestHeader(value = "Sort-Order", defaultValue = "ASC", required = false) Order sortOrder);
 
+    @GetMapping("/{bookingId}/visits-with-visitors")
+    @ApiOperation(value = "All visits for offender.", notes = "All visits for offender.", nickname = "getBookingVisitsWithVisitor")
+    @ApiResponses(value = {
+            @ApiResponse(code = 200, message = "OK", response = VisitWithVisitors.class, responseContainer = "List"),
```
brightonsbox, line 517: I think technically the 200 response is provided automatically so you don't need to add it explicitly.  No biggie though

Nimphal: Yeap, I've removed

src/main/java/net/syscon/elite/service/BookingService.java
brightonsbox: I don't think there's much that you can do about this, you just end up having to pick which one to use the full package name for 👍

src/test/java/net/syscon/elite/repository/jpa/repository/VisitInformationRepositoryTest.java
```diff
@@ -0,0 +1,57 @@
...
+        assertThat(visits).extracting(VisitInformation::getLocation).containsOnly("Visiting Room", "Visiting Room", "Visiting Room", "Chapel", "Visiting Room", "Visiting Room", "Visiting Room", "Visiting Room", "Visiting Room", "Chapel", "Visiting Room", "Chapel", "Visiting Room", "Visiting Room", "Visiting Room");
+        assertThat(visits).extracting(VisitInformation::getVisitType).containsOnly("SCON", "SCON", "SCON", "OFFI", "SCON", "SCON", "SCON", "SCON", "SCON", "OFFI", "SCON", "OFFI", "SCON", "SCON", "SCON");
```
brightonsbox, line 48: `containsOnly` ignores duplicates so this would still pass if you removed all but one of the `"SCON"` values. If you want to ensure that all duplicates are present, you can use `containsExactly` if the order matters or `containsExactlyInAnyOrder` if it doesn't

src/main/java/net/syscon/elite/service/BookingService.java
```diff
@@ -388,6 +402,53 @@ private void validateActivity(final Long activityId) {
         return bookingRepository.getBookingVisits(bookingId, fromDate, toDate, sortFields, sortOrder);
     }
 
+    @VerifyBookingAccess
+    public Page<VisitWithVisitors<Visit>> getBookingVisitsWithVisitor(final @NotNull Long bookingId, final LocalDate fromDate, final LocalDate toDate, final String visitType, final Pageable pageable) {
+        final var visits = visitRepository.findAllByBookingId(bookingId, pageable);
+
+        final var visitsWithVisitors = visits.stream()
+                .filter(visit -> fromDate == null || visit.getStartTime().isAfter(fromDate.atStartOfDay()))
+                .filter(visit -> toDate == null || visit.getStartTime().isBefore(toDate.atTime(LocalTime.of(23, 59))))
```
brightonsbox, line 411: Could maybe use before LocalTime.MAX here so you're not caught out by start times of 23:59 exactly?

Nimphal: Learn something new every day!

src/test/java/net/syscon/elite/repository/jpa/repository/VisitInformationRepositoryTest.java
Nimphal: That's a very good suggestion, updated :)

PR #595, opened 2020-06-22
DT-940 adding a specific for handling of prison to prison moves.

matt-crowson-moj:
Locking down prison to prison moves to a specific role.  Without this role you cannot do it.


src/main/java/net/syscon/elite/service/PrisonToPrisonMoveSchedulingService.java
```diff
@@ -71,6 +72,7 @@ public PrisonToPrisonMoveSchedulingService(final Clock clock,
     @Transactional
     @VerifyBookingAccess
```
Mjwillis, line 73: This this needs `@VerifyBookingAccess(overrideRoles = "PRISON_MOVE_MAINTAINER")`

PR #597, opened 2020-06-23
DT-944 fix to resolve role based access for court and prison moves.

matt-crowson-moj:


src/test/java/net/syscon/elite/executablespecification/steps/AuthTokenHelper.java
```diff
@@ -301,7 +301,7 @@ private String updateAlert() {
     private String courtHearingMaintainer() {
         return jwtAuthenticationHelper.createJwt(
                 JwtParameters.builder()
-                        .username("ITAG_USER")
+                        .username("DOES_NOT_EXIST")
```
matt-crowson-moj, line 304: Is there a better (clearer) way of doing this?

src/main/java/net/syscon/elite/service/CourtHearingsService.java
```diff
@@ -130,7 +130,7 @@ public CourtHearing scheduleHearing(final Long bookingId, final PrisonToCourtHea
     /**
      * Returns all court hearings for a given booking ID for the given date range.
      */
-    @VerifyBookingAccess
+    @VerifyBookingAccess(overrideRoles = "COURT_HEARING_MAINTAINER")
     public CourtHearings getCourtHearingsFor(final Long bookingId, final LocalDate fromDate, final LocalDate toDate) {
```
Mjwillis, line 134: is this missing a PreAuthorize?

Mjwillis: Do we need to restrict for reads?  Think this doesn't need the role, Ignore me, this is to allow unfettered access to court hearings if you have the role right?, maybe add the SYSTEM_USER role in to?

src/test/java/net/syscon/elite/executablespecification/steps/AuthTokenHelper.java
mikehalmamoj: Depends what you're trying to do.  Currently reads to me that you want client creds with a user that does not exist?

Mjwillis: Username could be the clientId.  @petergphillips maybe this should be `sub` now?

src/main/java/net/syscon/elite/service/CourtHearingsService.java
matt-crowson-moj: Done.

src/test/java/net/syscon/elite/executablespecification/steps/AuthTokenHelper.java
petergphillips: maybe, although  `username` is baked into spring security oauth2 so won't be coming out anytime soon.  I would have thought it would be better to not  pass in `username` at all, since that will be what pecs are doing.

PR #600, opened 2020-06-24
NN-2735 expose case-id

JamieDraperUK:


src/main/java/net/syscon/elite/api/model/OffenceHistoryDetail.java
```diff
@@ -64,4 +64,7 @@
 
     @ApiModelProperty(value = "Latest court date associated with the offence", example = "2018-02-10", position = 14)
     private LocalDate courtDate;
+
+    @ApiModelProperty(value = "Court case id", example = "100", position = 15)
+    private String caseId;
```
Nimphal, line 69: Just checking this is meant to be a string as ids are usually longs.

JamieDraperUK: Dam. Not had enough coffee yet!

Nimphal: Never enough coffee for this stuff :D

PR #614, opened 2020-07-01
DT-960 initial revision adding JPA entities for retrieving non-associations for offenders

matt-crowson-moj:
This is a first cut and may need some tweaking/additions.  This gets the initial structure in place (hopefully).


src/main/java/net/syscon/elite/repository/jpa/repository/OffenderNonAssociationDetailRepository.java
```diff
@@ -0,0 +1,10 @@
+package net.syscon.elite.repository.jpa.repository;
+
+import net.syscon.elite.repository.jpa.model.OffenderNonAssociationDetail;
+import org.springframework.data.repository.CrudRepository;
+
+import java.util.List;
+
+public interface OffenderNonAssociationDetailRepository extends CrudRepository<OffenderNonAssociationDetail, OffenderNonAssociationDetail.Pk> {
```
matt-crowson-moj, line 8: Not sure if both these repos are strictly needed.  Easy to remove if not though.

src/main/java/net/syscon/elite/repository/jpa/model/OffenderNonAssociationDetail.java
```diff
@@ -0,0 +1,110 @@
...
+    private Offender offender;
+
+    @Id
+    @ManyToOne(optional = false)
+    @JoinColumn(name = "NS_OFFENDER_ID", nullable = false, insertable = false, updatable = false, referencedColumnName = "OFFENDER_ID")
+    private Offender nsOffender;
+
+    @Id
+    @Column(name = "TYPE_SEQ", nullable = false, insertable = false, updatable = false)
+    private Integer typeSequence;
```
JamieDraperUK, line 61: Is it safe to use Integer here over Long?

matt-crowson-moj: Schema is a number(3) so yes.

src/main/java/net/syscon/elite/repository/jpa/repository/OffenderNonAssociationDetailRepository.java
JamieDraperUK: Well seems like you can retrieve the OffenderNonAssociation from the OffenderNonAssociationDetail so having one repository might better communicate which entity is the most important.

Can you have a OffenderNonAssociation without a OffenderNonAssociationDetail?

src/test/java/net/syscon/elite/repository/jpa/repository/OffenderNonAssociationRepositoryTest.java
```diff
@@ -0,0 +1,30 @@
...
+@Import({AuthenticationFacade.class, AuditorAwareImpl.class})
+@WithMockUser
+public class OffenderNonAssociationRepositoryTest {
+
+    @Autowired
+    private OffenderNonAssociationRepository nonAssociationRepository;
+
+    @Test
+    void find_non_associations_by_offender_booking() {
+        assertThat(nonAssociationRepository.findAllByOffenderBooking_BookingId(-1L)).hasSize(1);
```
JamieDraperUK, line 28: Is it work testing that the complex properties are properly hydrated? Some of the joins could be valid SQL but might be joining on the wrong properties? Be good to have a test at least from a regression standpoint.

src/main/java/net/syscon/elite/repository/jpa/repository/OffenderNonAssociationDetailRepository.java
matt-crowson-moj: I originally had the join the other way from non-associations to non-association details but it didn't really marry up with how the legacy DB was structure.  This is the one likely to stay in all honesty...

PR #632, opened 2020-07-06
Duplicate /Events Endpoint in prison API from custody api

Mjwillis:


src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderEvent.java
```diff
@@ -0,0 +1,92 @@
...
+import lombok.NoArgsConstructor;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class OffenderEvent {
```
andymarke, line 15: Eventually we should break this down into separate classes for each event type -- future work ...

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderEventsService.java
```diff
@@ -0,0 +1,94 @@
...
+    public OffenderEventsService(final OffenderEventsTransformer offenderEventsTransformer,
+                                 final OffenderEventsRepository offenderEventsRepository,
+                                 final XtagEventsService xtagEventsService) {
+        this.offenderEventsTransformer = offenderEventsTransformer;
+        this.offenderEventsRepository = offenderEventsRepository;
+        this.xtagEventsService = xtagEventsService;
+    }
+
+    @PreAuthorize("hasRole('PRISON_OFFENDER_EVENTS')")
+    public Optional<List<OffenderEvent>> getEvents(final Optional<LocalDateTime> maybeFrom,
```
andymarke, line 47: I bet you had IntelliJ telling you not to use Optionals as arguments ...

src/main/java/uk/gov/justice/hmpps/prison/api/resource/impl/BookingResourceImpl.java
```diff
@@ -204,7 +205,7 @@ public OffenderSummary recallOffenderBooking(@Valid final RecallBooking recallBo
 
     @Override
     public Movement getMovementByBookingIdAndSequence(final Long bookingId, final Integer sequenceNumber) {
-        return movementsService.getMovementByBookingIdAndSequence(bookingId, sequenceNumber);
+        return movementsService.getMovementByBookingIdAndSequence(bookingId, sequenceNumber).orElseThrow(EntityNotFoundException.withMessage(format("Movement Not found booking Id %d, seq %d", bookingId, sequenceNumber)));
```
andymarke, line 208: Why did you have to change the service to Optional?

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderEvent.java
Mjwillis: Agreed there is lots to improve!

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderEventsService.java
Mjwillis: it is

src/main/java/uk/gov/justice/hmpps/prison/api/resource/impl/BookingResourceImpl.java
Mjwillis: I did it to make it more consistent with custody api code and also realised there would be a null pointer if no movement was found.

PR #634, opened 2020-07-06
DT-989 adding new endpoint for pulling out offender non-association details

matt-crowson-moj:
This is a first cut of pulling out offender non-associations.  It may need some tweaking.


src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderNonAssociation.java
```diff
@@ -0,0 +1,29 @@
...
+import lombok.NoArgsConstructor;
+
+@ApiModel(description = "Offender non-association")
+@Builder(toBuilder = true)
+@AllArgsConstructor
+@NoArgsConstructor
+@Data
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@EqualsAndHashCode
+public class OffenderNonAssociation {
```
matt-crowson-moj, line 19: I should probably add the offender name to this also???

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderNonAssociationDetail.java
```diff
@@ -0,0 +1,52 @@
...
+import java.time.LocalDateTime;
+
+@ApiModel(description = "Offender non-association detail")
+@Builder(toBuilder = true)
+@AllArgsConstructor
+@NoArgsConstructor
+@Data
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@EqualsAndHashCode
+public class OffenderNonAssociationDetail {
```
matt-crowson-moj, line 21: I should probably add the offender name to this also???

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderNonAssociation.java
Nimphal: That would save us an extra call to get the offender details when we come to use this :)

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderNonAssociationDetails.java
```diff
@@ -0,0 +1,22 @@
...
+@ApiModel(description = "Offender non-association details")
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@EqualsAndHashCode
+public class OffenderNonAssociationDetails {
+    @ApiModelProperty(value = "Offender non-association details")
+    private Collection<OffenderNonAssociationDetail> details;
+}
```
Nimphal, line 22: More musing than anything, what's the benefit of having the list of non-associations be wrapped in a details object? Only asking because I would've totally just returned `List<OffenderNonAssociationDetail>` from the resource and not considered doing a separate model.

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderNonAssociation.java
matt-crowson-moj: I should probably also add the location of the prisoners...

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderNonAssociationDetails.java
matt-crowson-moj: It is actually GDS guidelines.  I'll see if I can dig it out....

matt-crowson-moj: https://www.gov.uk/guidance/gds-api-technical-and-data-standards#use-restful  (see section 'Use Json'

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderNonAssociation.java
matt-crowson-moj: Added first and last name.  Still need to add locations.

PR #636, opened 2020-07-07
DT-1006 Correct sentence result for offender no when multiple bookings.  Incorrectly choosing the first entry when it should be the current booking.

Mjwillis:


src/main/java/uk/gov/justice/hmpps/prison/service/BookingService.java
```diff
@@ -736,11 +740,11 @@ public void checkBookingExists(final Long bookingId) {
     }
 
     public Optional<OffenderSentenceDetail> getOffenderSentenceDetail(final String offenderNo) {
-        final var query = format("offenderNo:eq:'%s'", offenderNo);
-        return getOffenderSentenceDetails(bookingRepository.getOffenderSentenceSummary(query, getCaseLoadIdForUserIfRequired(),
-                !isViewAllOffenders(), isViewInactiveBookings()))
```
Nimphal, line 741: Are these access checks replicated in the inmateRepository?

PR #658, opened 2020-07-20
NN-2803 handle null court cases

JamieDraperUK:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderBooking.java
```diff
@@ -97,7 +97,7 @@ public boolean isActive() {
     }
 
     public List<OffenderCourtCase> getActiveCourtCases() {
-        return courtCases.stream().filter(OffenderCourtCase::isActive).collect(toUnmodifiableList());
+        return courtCases == null ? Collections.emptyList() : courtCases.stream().filter(OffenderCourtCase::isActive).collect(toUnmodifiableList());
```
matt-crowson-moj, line 100: Can we not just a default on the court cases? e.g.

```
@Builder.Default
private List<OffenderCourtCase> courtCases = new ArrayList<>();
```

JamieDraperUK: Yep, that's a lot cleaner 👍

PR #676, opened 2020-08-03
Nn 2807 visitor multiple relationships

Nimphal:
Sometimes there can be multiple `OFFENDER_CONTACT_PERSONS` for the same booking and person id. It is rare, but there are examples in live, around 25,000 records. The current VisitInformation subselect is doing a left join into that table, which means that when there is more than one record, it actually returns multiple duplicates of the same visit data, even though it always picks the most recent contact record. Similar for the visitor data.

This extracts that bit from the subselect into its own JPA repository and delegates the relationship logic to the service.

Having multiple such records also affects the last and next visit endpoints, but it doesn't matter, because we always sort and pick the top of the stack on that. The tests needed updating because we always return the latest relationship record and the test data has changed.


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/VisitorRepository.java
```diff
@@ -8,5 +8,5 @@
 
 @Repository
 public interface VisitorRepository extends PagingAndSortingRepository<VisitorInformation, String> {
-    List<VisitorInformation> findAllByVisitIdAndBookingId(Long visitId, Long bookingId);;
+    List<VisitorInformation> findAllByVisitId(Long visitId);
```
Nimphal, line 11: Visit Id is a primary key on the visits table, so there's no need to also use the booking id to identify a visitor.

PR #681, opened 2020-08-10
NN-2848: Return multiple non associations

Nimphal:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderBooking.java
```diff
@@ -72,8 +73,7 @@
     @Column(name = "ACTIVE_FLAG")
     private String activeFlag;
 
-    @OrderColumn(name = "TYPE_SEQ")
-    @ListIndexBase(1)
+    @OrderBy("effectiveDate ASC")
```
Mjwillis, line 76: Generally I don't like order bys in JPA entity. Be better to have a sortedSet and implement a comparator. (you can specific this in the JPA)  Otherwise this is issued on every request when loading a list and may not always be wanted.

PR #683, opened 2020-08-10
🐛 DT-1060 Only show balances for current prison

Mjwillis:
Where outside no balance will be shown


src/main/java/uk/gov/justice/hmpps/prison/api/model/Account.java
```diff
@@ -1,124 +1,45 @@
...
 
+    @ApiModelProperty(required = true, value = "Spends sub account balance.")
     @NotNull
     private BigDecimal spends;
 
+    @ApiModelProperty(required = true, value = "Cash sub account balance.")
     @NotNull
     private BigDecimal cash;
 
+    @ApiModelProperty(required = true, value = "Saves sub account balance.")
```
petergphillips, line 38: should be _savings_ I think

PR #690, opened 2020-08-13
Nn 2728 cell capacity location

Nimphal:
Nomis checks the capacity vs number of occupants as stored in the `AGENCY_INTERNAL_LOCATIONS` table rater than the `LIVING_UNITS` table. This updates the service accordingly so the API only returns cells with space according to those rules.


src/main/java/uk/gov/justice/hmpps/prison/service/AgencyServiceImpl.java
```diff
@@ -255,23 +257,32 @@ private boolean isBlankAddress(final PrisonContactDetail pcd) {
...
-                                .map(profile -> OffenderCellAttribute.builder()
-                                     .code(profile.getHousingAttributeReferenceCode().getCode())
-                                     .description(profile.getHousingAttributeReferenceCode().getDescription())
-                                     .build())
-                                .collect(Collectors.toList()))
-                    .build())
-                .filter(cell -> attribute == null || cell.getAttributes().stream().map(OffenderCellAttribute::getCode).collect(Collectors.toList()).contains(attribute))
-                .collect(Collectors.toList());
+                .filter(livingUnit -> {
+                    final var agencyLocation = agencyInternalLocationRepository.findOneByDescription(livingUnit.getDescription());
```
brightonsbox, line 261: Rather than calling `agencyInternalLocationRepository.findOneByDescription(livingUnit.getDescription())` twice (once to filter, then again in the map), you could do something like:
```
return livingUnits.stream()
            .map(this::transform)
            .flatMap(Optionals::stream)
            .filter(cell -> attribute == null || cell.getAttributes().stream().map(OffenderCellAttribute::getCode).collect(toList()).contains(attribute))
            .collect(toList());
```
where `transform` is a method that returns an optional like:

```
private Optional<OffenderCell> transform(final LivingUnit livingUnit) {
    return agencyInternalLocationRepository.findOneByDescription(livingUnit.getDescription())
        .filter(AgencyInternalLocation::isActiveCellWithSpace)
        .map(agencyLocation -> 
              OffenderCell.builder()
                      ...
              .build()
        );
}
```

Nimphal: Cheers! I am very sure I will back looking at this endpoint soon so will do this then :)

PR #692, opened 2020-08-17
PAT-1200 Get all assessments for latest booking

steverendell:
Correct problem with flag that was used for 2 diffenrent things.
Conveniently sorts a pathfinder issue 👍


src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderAssessmentResource.java
```diff
@@ -42,18 +42,19 @@
...
-                                                          @ApiParam(value = "Returns only active assessments if true, otherwise assessments with any status are included", defaultValue = "true") @RequestParam(value = "activeOnly", required = false, defaultValue = "true") Boolean activeOnly);
+                                                          @ApiParam(value = "Returns only assessments for the current sentence if true, otherwise assessments for all previous sentences are included", defaultValue = "true") @RequestParam(value = "latestOnly", required = false, defaultValue = "true") Boolean latestOnly,
+                                                          @ApiParam(value = "Returns only active assessments if true, otherwise inactive and pending assessments are included", defaultValue = "true") @RequestParam(value = "activeOnly", required = false, defaultValue = "true") Boolean activeOnly,
```
mikehalmamoj, line 46: Have you checked for other clients of this API to make sure they didn't rely on the old functionality?  (and for other endpoints changed)

mikehalmamoj:
That method InmateService#getInmatesAssessmentsByCode feels like it might need some surgery soon but at least it works now :smile:

PR #693, opened 2020-08-21
DT 1070 include the agency establishment types for a given agency

matt-crowson-moj:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/AgencyLocationEstablishmentRepository.java
```diff
@@ -0,0 +1,8 @@
+package uk.gov.justice.hmpps.prison.repository.jpa.repository;
+
+
+import org.springframework.data.repository.CrudRepository;
+import uk.gov.justice.hmpps.prison.repository.jpa.model.AgencyLocationEstablishment;
+
+public interface AgencyLocationEstablishmentRepository extends CrudRepository<AgencyLocationEstablishment, AgencyLocationEstablishment.Pk> {
```
andymarke, line 7: I couldn't see where this was being used (maybe I should review in Intellij)

src/main/java/uk/gov/justice/hmpps/prison/api/model/Agency.java
```diff
@@ -31,4 +34,9 @@
 
     @ApiModelProperty(value = "Agency is active", example = "true")
     private boolean active;
+
+    @ApiModelProperty(value = "Establishment types for the agency", example = "IM")
```
andymarke, line 38: Maybe add additional description of an agency type, it still means nothing to me. Does in indicate if this is a women's prison for instance?

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/api/model/Agency.java
matt-crowson-moj: Agreed and really  that was originally what I intended.  So I had some JPA mapping issues which I never found a clean solution to.  This supporting attribute for this is actually a reference code.  In this case it forms part of the composite key on the new JPA entity AgencyLocationEstablishment.  The particular entity attribute would be using a JoinColumnOrFormula annotation and JPA does not this when part of a composite key, hence I couldn't pull in the description.  So we can (at least) get the description from the reference code endpoint (for now).

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/AgencyLocationEstablishmentRepository.java
matt-crowson-moj: Yeah I probably take this it out.  It is a hangover from the first commit...

matt-crowson-moj:
Going to have another think about this one for now.

PR #694, opened 2020-08-24
DT-1070 refactoring agency establishment types into distinct (new) endpoint.

matt-crowson-moj:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/AgencyLocationEstablishment.java
```diff
@@ -0,0 +1,58 @@
...
+public class AgencyLocationEstablishment {
+
+    @Id
+    @Column(name = "AGY_LOC_ID", nullable = false, insertable = false, updatable = false)
+    private String agencyLocId;
+
+    @Id
+    @Column(name = "ESTABLISHMENT_TYPE", nullable = false, insertable = false, updatable = false)
+    private String establishmentType;
+
```
andymarke, line 49: Not sure how NOMIS ref. data works; but could you not join here to the "Reference table" to get the descriptions, rather than having to call the service individually for each type?

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/AgencyLocationEstablishment.java
matt-crowson-moj: Indeed it should have been but I had a JPA issue getting it to work.  JoinColumnOrFormula annotation issues on complex primary keys.

PR #696, opened 2020-08-25
Nn 2721 cell swap api

JamieDraperUK:
- First pass at an endpoint that moves an offender into the pre-configured cell swap location.


src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderMovementsResource.java
```diff
@@ -81,6 +81,19 @@ OffenderBooking moveToCell(
...
+
+    @PutMapping("/{bookingId}/move-to-cell-swap")
+    @ApiResponses(value = {
+            @ApiResponse(code = 200, message = "OK", response = OffenderBooking.class),
+            @ApiResponse(code = 400, message = "Invalid request.", response = ErrorResponse.class),
+            @ApiResponse(code = 404, message = "Requested resource not found.", response = ErrorResponse.class),
+            @ApiResponse(code = 500, message = "Unrecoverable error occurred whilst processing request.", response = ErrorResponse.class)})
+    OffenderBooking moveToCellSwap(
+            @ApiParam(value = "The offender booking id", example = "1200866", required = true) @PathVariable("bookingId") Long bookingId,
+            @ApiParam(value = "The reason code for the move (from reason code domain CHG_HOUS_RSN)", example = "ADM", required = true) @RequestParam("reasonCode") String reasonCode,
```
andymarke, line 93: Would a RequestBody be more appropriate for tis PUT rather than parameters?

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderMovementsResource.java
JamieDraperUK: Now you mention it, yeah it would definitely feel more natural.

JamieDraperUK: I'll address it in the next PR as I know we'll need to make some changes once we've played with it a bit.

PR #697, opened 2020-08-25
Nn 1180 offender cell history api

Nimphal:


src/main/java/uk/gov/justice/hmpps/prison/api/model/BedAssignment.java
```diff
@@ -0,0 +1,31 @@
...
+import java.time.LocalDateTime;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+@EqualsAndHashCode
+@ToString
+@Data
+public class BedAssignment {
```
matt-crowson-moj, line 21: Are you going to add some Swagger docs/annotations?

src/main/java/uk/gov/justice/hmpps/prison/api/model/BedAssignment.java
```diff
@@ -0,0 +1,31 @@
...
+    private Long bookingId;
+    private Long livingUnitId;
+    private LocalDate assignmentDate;
```
matt-crowson-moj, line 24: Do you need to separate the date and times into separate fields?

matt-crowson-moj (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/api/model/BedAssignment.java
Nimphal: They are returned like that from the mapping and we may end up using them like this. I'll probably update at some point as we've had a play.

src/main/java/uk/gov/justice/hmpps/prison/api/model/BedAssignment.java
Nimphal: That's a good idea, will do in a following PR, I know we have more things in the pipeline around this.

PR #700, opened 2020-08-26
DT-1080: upgraded spring boot plugin to 1.0.0

SimonMitchellMOJ:


src/main/resources/application.yml
```diff
@@ -43,15 +43,10 @@ management:
       show-details: always
       probes:
         enabled: true
-      group:
```
matt-crowson-moj, line 46: Has this changed again?  This was only added/updated recently.  Tests for liveness and readiness pass so must be fine, just curious.

matt-crowson-moj (approved):
LGTM

src/main/resources/application.yml
SimonMitchellMOJ: yes, there was something that spring did with 2.3.2 which meant that you need to add the group section in for liveness and readiness to continue working, in 2.3.3 they have either fixed it or reverted what they did so you no longer need it in

PR #704, opened 2020-09-03
DT-1023: ✨⬆️ Add in root offender id to offender booking

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/service/OffenderAddressService.java
```diff
@@ -27,12 +27,9 @@
     private final PhoneRepository phoneRepository;
 
     @VerifyOffenderAccess
-    public List<AddressDto> getAddressesByOffenderNo(@NotNull String offenderNo) {
-        final var offenderBookings = offenderBookingRepository.findByOffenderNomsIdAndActiveFlag(offenderNo, "Y");
-        if (offenderBookings.size() > 1)
```
petergphillips, line 32: didn't need to return a list if it enforced only one anyway

mikehalmamoj (approved):

PR #705, opened 2020-09-03
DT-1095 validating offender no format as was causing a formatting exception further down the stack.

matt-crowson-moj:


Mjwillis (approved):
LGTM

PR #706, opened 2020-09-04
NN-2904: Get attributes based on agency internal location rather than living unit

Nimphal:


src/test/resources/features/locations.feature
```diff
@@ -19,7 +19,7 @@ Feature: Locations
       | locationId | number | type | description |
       | -1         | 1      | WING | Block A     |
       | -2         | 1      | LAND | Landing A/1 |
-      | -3         | 1      | CELL | A-1-1       |
+      | -3         | 1      | CELL | LEI-A-1-1       |
```
Nimphal, line 22: Added a user description to the database as part of testing data, and the location processor does not change those, so this needed updating.

petergphillips (approved):
LGTM

PR #709, opened 2020-09-04
Revert "DT-1095 validating offender no format as was causing a formatting exception further down the stack."

matt-crowson-moj:
Reverts ministryofjustice/prison-api#705


petergphillips (approved):
LGTM

PR #711, opened 2020-09-04
AR-160: Seed data for Artemis users

timharrison-moj:
* Creates seed data for Nomis users PRISON_COLLATOR_LOCAL and PRISON_ANALYST_LOCAL
* Create seed data for Nomis role 
* Removes unused used ARTEMIS_PRISON_USER (too general)


petergphillips (approved):
LGTM

PR #724, opened 2020-09-14
Nn 2920 cell history datetime

Nimphal:
Change cell history to use the full datetime rather than just date, to ensure we get the complete picture of who has been in a cell when.


JamieDraperUK (approved):
lgtm

PR #726, opened 2020-09-14
DT-1119: 🔇 Switch to logging message instead of exception so doesn't …

petergphillips:
…appear in app insights exceptions table

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/aop/LoggingAspect.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/aop/LoggingAspect.java
@@ -8,6 +8,8 @@ import org.aspectj.lang.annotation.Around;
 import org.aspectj.lang.annotation.Aspect;
 import org.aspectj.lang.annotation.Pointcut;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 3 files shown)


mikehalmamoj (approved):

src/main/java/uk/gov/justice/hmpps/prison/aop/LoggingAspect.java
```diff
@@ -26,12 +28,12 @@ public void loggingPointcut() {
 
     @AfterThrowing(pointcut = "loggingPointcut()", throwing = "e")
     public void logAfterThrowing(final JoinPoint joinPoint, final Throwable e) {
-        log.error(
-                String.format(
-                        "Exception in pointcut %s.%s()",
-                        joinPoint.getSignature().getDeclaringTypeName(),
-                        joinPoint.getSignature().getName()),
-                e);
+        if (!(e instanceof EntityNotFoundException || e instanceof NoContentException)) {
```
petergphillips, line 31: These two account for 14,000 exceptions in the logs in the last 24 hours, so remove from here and log message instead in controller advise

JamieDraperUK (approved):
lgtm 👍

PR #731, opened 2020-09-23
Filter out inactive cases or charges for recalls

Mjwillis:


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffendersResourceTest.java
```diff
@@ -121,6 +121,40 @@ public void getFullOffenderInformation() {
...
+        final var response = testRestTemplate.exchange(
+                "/api/offenders/{nomsId}",
+                HttpMethod.GET,
+                httpEntity,
+                new ParameterizedTypeReference<String>() {
+                },
+                "A1234AC");
+
+        assertThatJsonFileAndStatus(response, 200, "offender_detail_recall.json");
+    }
```
andymarke, line 139: going forward it might be worth thinking about bring the json asserts inline rather than referring to a file.

andymarke (approved):
LGTM

PR #744, opened 2020-10-22
NN3007: first set of classes

jamesak-oncloud9:


src/main/java/uk/gov/justice/hmpps/prison/api/model/TransactionHistory.java
```diff
@@ -0,0 +1,16 @@
+package uk.gov.justice.hmpps.prison.api.model;
+
+import lombok.Builder;
+
+import java.util.Collections;
+import java.util.List;
+
+@Builder
+public class TransactionHistory {
```
JamieDraperUK, line 9: Not sure what this gives us. You could just have a TransactionHsitoryReposiutory that returns TransactionHistoryItem or  TransactionHistory.

src/main/java/uk/gov/justice/hmpps/prison/api/resource/NomisApiResource.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.web.bind.annotation.RequestParam;
+import uk.gov.justice.hmpps.prison.api.model.ErrorResponse;
+import uk.gov.justice.hmpps.prison.api.model.TransactionHistory;
+
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Pattern;
+import javax.validation.constraints.Size;
+import java.time.LocalDate;
+
+@Api(tags = {"/api"})
```
JamieDraperUK, line 15: This endpopint should live under the offender resource /offenders/

src/main/java/uk/gov/justice/hmpps/prison/api/model/TransactionHistoryDto.java
```diff
@@ -0,0 +1,12 @@
+package uk.gov.justice.hmpps.prison.api.model;
+
+import lombok.Builder;
+import uk.gov.justice.hmpps.prison.repository.jpa.model.TransactionHistory;
+
+@Builder
+public class TransactionHistoryDto {
+
+    public static TransactionHistoryDto toDto(TransactionHistory transactionHistory) {
```
JamieDraperUK, line 9: We don't normally map like this. There is usually single mapping function that lives in a transform type class, either placed in the transform namespace or a function inside the service.   

As for data types it just an entity and set of api models containing swagger docs etc. See the ones in the following namespace: **uk.gov.justice.hmpps.prison.api.model;**

src/main/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceImpl.java
```diff
@@ -231,4 +215,15 @@ public PrivilegeSummary getLatestBookingIEPSummaryForOffender(final String offen
         Optional.ofNullable(booking).orElseThrow(EntityNotFoundException.withId(offenderNo));
         return bookingService.getBookingIEPSummary(booking.getBookingId(), withDetails);
     }
+
+    @Override
+    public ResponseEntity<List<TransactionHistoryDto>> getTransactionsHistory(final String prisonId, final String nomsId,
+                                                        final String accountCode,
+                                                        final LocalDate fromDate, final LocalDate toDate) {
```
JamieDraperUK, line 222: This logic normally goes in a service.

src/main/java/uk/gov/justice/hmpps/prison/repository/OffenderTransactionHistoryRepository.java
```diff
@@ -0,0 +1,13 @@
+package uk.gov.justice.hmpps.prison.repository;
+
+import uk.gov.justice.hmpps.prison.repository.jpa.model.TransactionHistory;
+
+import java.time.LocalDate;
+import java.util.List;
+
+public interface OffenderTransactionHistoryRepository {
+
+    List<TransactionHistory> getTransactionsHistory(final String prisonId, final String nomisId,
```
JamieDraperUK, line 10: Meh, these don't need to be final.

src/main/java/uk/gov/justice/hmpps/prison/repository/impl/OffenderTransactionHistoryRepositoryImpl.java
```diff
@@ -0,0 +1,43 @@
...
+
+    private final FinanceProcs.GetAccountTransactions getAccountTransactionsProc;
+
+    public OffenderTransactionHistoryRepositoryImpl(FinanceProcs.GetAccountTransactions getAccountTransactionsProc) {
+        this.getAccountTransactionsProc = getAccountTransactionsProc;
+    }
+
+    @Override
+    public List<TransactionHistory> getTransactionsHistory(String prisonId, String nomisId, String accountCode, LocalDate fromDate, LocalDate toDate) {
+        final var params = new MapSqlParameterSource()
```
JamieDraperUK, line 29: We should be using JPA repositories for this.

src/main/java/uk/gov/justice/hmpps/prison/repository/impl/FinanceRepositoryImpl.java
```diff
@@ -5,27 +5,41 @@
...
 
-    public FinanceRepositoryImpl(@Value("${api.currency:GBP}") final String currency, final InsertIntoOffenderTrans insertIntoOffenderTrans, final ProcessGlTransNew processGlTransNew) {
+    public FinanceRepositoryImpl(@Value("${api.currency:GBP}") final String currency,
+                                 final InsertIntoOffenderTrans insertIntoOffenderTrans,
+                                 final ProcessGlTransNew processGlTransNew,
+                                 final FinanceProcs.GetAccountTransactions getAccountTransactionsProc) {
         this.currency = currency;
         this.insertIntoOffenderTrans = insertIntoOffenderTrans;
         this.processGlTransNew = processGlTransNew;
+        this.getAccountTransactionsProc = getAccountTransactionsProc;
```
JamieDraperUK, line 42: Is this meant to have been committed?

src/main/java/uk/gov/justice/hmpps/prison/api/model/TransactionHistoryDto.java
jamesak-oncloud9: it's a just place holder method, will follow existing examples

src/main/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceImpl.java
jamesak-oncloud9: sure, will make the service class expose Dto objects then

src/main/java/uk/gov/justice/hmpps/prison/repository/impl/FinanceRepositoryImpl.java
jamesak-oncloud9: no, will remove

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderTransactionHistoryService.java
```diff
@@ -0,0 +1,61 @@
...
+
+        checkNotNull(offenderId, "offender-id can't be null");
+        checkNotNull(accountCode, "accountCode optional can't be null");
+        checkNotNull(fromDate, "fromDate can't be null");
+        checkNotNull(toDate, "toDate can't be null");
+
+        var now = LocalDate.now();
+        checkState(fromDate.isBefore(toDate) || fromDate.isEqual(toDate), "toDate can't be before fromDate");
+        checkState(fromDate.isBefore(now) || fromDate.isEqual(now), "fromDate can't be in the future");
+        checkState(toDate.isBefore(now) || toDate.isEqual(now), "toDate can't be in the future");
```
jamesak-oncloud9, line 45: A DateRange, OffenderId and AccountCode value objects would remove these extra checks and reduce code noise here.

src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
```diff
@@ -201,4 +202,21 @@ CaseNote updateOffenderCaseNote(@ApiParam(value = "Noms ID or Prisoner number (a
...
+            @ApiResponse(code = 404, message = "Prison, offender or accountType not found", response = ErrorResponse.class),
+            @ApiResponse(code = 500, message = "Unrecoverable error occurred whilst processing request.", response = ErrorResponse.class)})
+    ResponseEntity<List<OffenderTransactionHistoryDto>> getTransactionsHistory(
+            @ApiParam(name = "offenderNo", value = "Offender ID", example = "123", required = true) @PathVariable(value = "offenderNo", required = true) @NotNull Long offenderId,
+            @ApiParam(name = "account_code", value = "Account code", example = "spends", required = false, allowableValues = "spends,cash,savings") @RequestParam(value = "account_code", required = false) String accountCode,
+            @ApiParam(name = "from_date", value = "Start date for transactions, format yyyy-MM-dd, defaults to today if not supplied", example = "2019-04-01") @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) @RequestParam(value = "from_date", required = false) LocalDate fromDate,
```
jamesak-oncloud9, line 220: I would use String instead of LocalDate for dates so that a more precise message is provided when a bad format is used

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderTransactionHistoryService.java
```diff
@@ -0,0 +1,61 @@
...
+@Service
+@Transactional(readOnly = true)
+@AllArgsConstructor
+@Validated
+@Slf4j
+public class OffenderTransactionHistoryService {
+
+    private OffenderTransactionHistoryRepository repository;
+
+    public List<OffenderTransactionHistoryDto> getTransactionHistory(final Long offenderId,
```
jamesak-oncloud9, line 32: in my view, a service class should not return a dto, it should return a domain class and it's up to the web controller to convert it to a dto that can be packaged in http response object.

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderTransactionHistory.java
```diff
@@ -0,0 +1,61 @@
...
+    @Column(name = "OFFENDER_ID", nullable = false)
+    private Long offenderId;
+
+    @Column(name = "TXN_ENTRY_DATE")
+    private LocalDate entryDate;
+
+    @Column(name = "TXN_TYPE", nullable = false)
+    private String transactionType;
+
+    @Column(name = "TXN_ENTRY_DESC", nullable = false)
```
JamieDraperUK, line 37: This can be null.

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderTransactionHistory.java
```diff
@@ -0,0 +1,61 @@
...
+import java.time.LocalDate;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+@Entity
+@IdClass(OffenderTransaction.Pk.class)
+@Table(name = "OFFENDER_TRANSACTIONS")
+public class OffenderTransactionHistory {
```
JamieDraperUK, line 18: The String values are missing a column length.

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderTransactionHistory.java
```diff
@@ -0,0 +1,61 @@
...
+    @Column(name = "TXN_ENTRY_DATE")
+    private LocalDate entryDate;
+
+    @Column(name = "TXN_TYPE", nullable = false)
+    private String transactionType;
+
+    @Column(name = "TXN_ENTRY_DESC", nullable = false)
+    private String entryDescription;
+
+    @Column(name = "TXN_REFERENCE_NUMBER", nullable = false)
```
JamieDraperUK, line 40: This can be null.

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderTransactionHistory.java
```diff
@@ -0,0 +1,61 @@
...
+    private String referenceNumber;
+
+    @Column(name = "TXN_ENTRY_AMOUNT", nullable = false)
+    private BigDecimal entryAmount;
+
+    @Column(name = "SUB_ACCOUNT_TYPE", nullable = false)
```
JamieDraperUK, line 46: There is an entity for OffenderSubAccount might be worth adding a reference to it here.

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderTransactionHistoryService.java
```diff
@@ -0,0 +1,61 @@
...
+                                                                     final Optional<String> accountCode,
+                                                                     final LocalDate fromDate,
+                                                                     final LocalDate toDate) {
+
+        checkNotNull(offenderId, "offender-id can't be null");
```
JamieDraperUK, line 37: Not sure this can ever be null unless used outside of Controller/Resource.

src/main/java/uk/gov/justice/hmpps/prison/service/transformers/OffenderTransactionHistoryTransformer.java
```diff
@@ -0,0 +1,21 @@
+package uk.gov.justice.hmpps.prison.service.transformers;
+
+import uk.gov.justice.hmpps.prison.api.model.OffenderTransactionHistoryDto;
+import uk.gov.justice.hmpps.prison.repository.jpa.model.OffenderTransactionHistory;
+
+public class OffenderTransactionHistoryTransformer {
+
+    public static OffenderTransactionHistoryDto transform(OffenderTransactionHistory entity) {
```
JamieDraperUK, line 8: Missing final.

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderTransactionHistoryService.java
JamieDraperUK, line 32: I forgot to mention that we need to check to see if the current user active caseload matches the offenders current residency. 

To make this check simply annotation the service method with: ` @VerifyOffenderAccess`

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderTransactionHistoryResourceTest.java
```diff
@@ -0,0 +1,338 @@
...
+                OFFENDER_NUMBER);
+
+        assertThat(response.getBody().getDeveloperMessage()).isEqualTo("toDate can't be before fromDate");
+        assertThat(response.getBody().getStatus().intValue()).isEqualTo(HTTP_BAD_REQ);
+        assertThat(response.getBody().getUserMessage()).isEqualTo("toDate can't be before fromDate");
+    }
+
+    @Test
+    public void When_GetOffenderTransactionHistory_And_FromDateIsMissing_And_ToDateIsToday_Then_NoErrorIsThrown() {
+
```
JamieDraperUK, line 195: Is it worth verifying that the repository was called with the correct default date? Might be easier to test if that default date logic lived in the service.

src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
```diff
@@ -201,4 +202,21 @@ CaseNote updateOffenderCaseNote(@ApiParam(value = "Noms ID or Prisoner number (a
...
             @ApiParam(value = "offenderNo", required = true, example = "A1234AA") @PathVariable("offenderNo") @NotNull String offenderNo,
             @ApiParam(value = "Toggle to return IEP detail entries in response (or not).", required = true) @RequestParam(value = "withDetails", required = false, defaultValue = "false") boolean withDetails);
+
+    String NOMS_ID_REGEX_PATTERN = "[a-zA-Z][0-9]{4}[a-zA-Z]{2}";
+
+    @SuppressWarnings("RestParamTypeInspection")
+    @GetMapping("/{offenderNo}/transaction-history")
+    @ApiOperation(value = "Retrieve an offender's financial transaction history for cash, spends or savings.",
+            notes = "Transactions are returned in NOMIS order (Descending date followed by id).<br/>" +
+            "All transaction amounts are represented as pence values.") //todo order?, should use int for money?
```
JamieDraperUK, line 212: Did you mean to check this //todo in?

src/main/java/uk/gov/justice/hmpps/prison/values/Money.java
```diff
@@ -0,0 +1,61 @@
+package uk.gov.justice.hmpps.prison.values;
+
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class Money {
```
JamieDraperUK, line 8: Seems strange for us to be modeling stuff outside our domain is there something in the java lib we could use for this?

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderTransactionHistoryDto.java
```diff
@@ -0,0 +1,48 @@
...
+    @ApiModelProperty(value = "Transaction Id", example = "1", position = 2)
+    private Long transactionId;
+
+    @ApiModelProperty(value = "Transaction Sequence", example = "1", position = 3)
+    private Long transactionEntrySequence;
+
+    @ApiModelProperty(value = "Transaction Date", example = "2020-12-11", position = 4)
+    private LocalDate entryDate;
+
+    @ApiModelProperty(value = "Transaction Type", example = "???", position = 5) //todo
```
JamieDraperUK, line 34: Do you mean to check this in?

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderTransactionHistoryDto.java
```diff
@@ -0,0 +1,48 @@
...
+    private String transactionType;
+
+    @ApiModelProperty(value = "Transaction Description", example = "some textual description here", position = 6)
+    private String entryDescription;
+
+    @ApiModelProperty(value = "Transaction Reference Number", example = "???", position = 7) //todo
```
JamieDraperUK, line 40: Do you mean to check this in?

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderTransactionHistoryResourceTest.java
```diff
@@ -0,0 +1,338 @@
+package uk.gov.justice.hmpps.prison.api.resource.impl;
```
JamieDraperUK, line 1: I think it's missing a test case for the Offender not found.

src/main/java/uk/gov/justice/hmpps/prison/values/Money.java
```diff
@@ -0,0 +1,61 @@
...
+        GBP("British Pound", "£");
+        public final String name;
+        public final String symbol;
+        Currency(final String name, final String symbol) {
+            this.name = name;
+            this.symbol = symbol;
+        }
+    }
+
+    private final Currency currency = Currency.GBP;
```
JamieDraperUK, line 22: I've just noticed everywhere else we are pulling the currency from config but we're are hard coding it here.

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderTransactionHistoryDto.java
jamesak-oncloud9: will try and find out what Transaction Types there are and remove

src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
jamesak-oncloud9: I'm sorting by entryDate Descending and Sequence Ascending is that correct?

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderTransactionHistoryService.java
jamesak-oncloud9: indeed this is there to protect the service code and to communicate to future developers that they can't pass null for this

src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
JamieDraperUK: Sounds right to me 👍

src/main/java/uk/gov/justice/hmpps/prison/values/Money.java
jamesak-oncloud9: as I understanding it entryAmount is money value and that needs to be represented as pence value as discussed previously and that conversion code needs to be placed somewhere, please suggest a better approach if you wish

src/main/java/uk/gov/justice/hmpps/prison/values/Money.java
jamesak-oncloud9: there are no requirements for this, it's there for completeness and hence defaulted to GBP, can be easily changed in the future when there are hard requirements for this

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderTransactionHistoryResourceTest.java
jamesak-oncloud9: it's an empty result set for which there is a test

src/main/java/uk/gov/justice/hmpps/prison/values/Money.java
JamieDraperUK: I agree there is no requirement but for consistency it would be worth doing.

src/main/java/uk/gov/justice/hmpps/prison/values/Money.java
JamieDraperUK: I've seen it done this way in most places, in pounds but seems like a simple solution -   

```
public BigDecimal getAmountInPounds() {
        return new BigDecimal(amount).setScale(2, RoundingMode.HALF_UP).divide(new BigDecimal("100"), RoundingMode.HALF_UP);
    }
```

JamieDraperUK: I'd say just return the amounts as they come (pounds). Unless there a problem we're trying to solve by returning pence.

jamesak-oncloud9: i assumed from exiting code that we represent money in pence value, if that's not the case we don't need this

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderTransactionHistoryDto.java
```diff
@@ -0,0 +1,48 @@
...
+import java.time.LocalDate;
+
+@ApiModel(description = "Offender transaction details")
+@Data
+@EqualsAndHashCode
+@ToString
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+public class OffenderTransactionHistoryDto {
```
JamieDraperUK, line 20: This model is missing a currency property.

src/main/java/uk/gov/justice/hmpps/prison/values/Currency.java
```diff
@@ -0,0 +1,19 @@
+package uk.gov.justice.hmpps.prison.values;
+
+import java.util.Arrays;
+import java.util.Optional;
+
+public enum Currency {
```
JamieDraperUK, line 6: I can't see it happening anytime soon but if we needed to make the app work using USD someone will update the properties but risk not updating this class.

jamesak-oncloud9: I used Currency to make the transformer code more expressive as I don't like using primitive types to represent domain concepts. Happy to remove it though, just let me know.

JamieDraperUK: it tightly couples it's self with the currency config. If someone changed the config to USD your code will just ignore it and default to GBP because your model doesn't support other currencies at the moment. 

Let's keep it simple for now and use primitive types then at least it's constant.

jamesak-oncloud9: if correct TDD practice is followed this is not an issue at all, will remove

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderTransactionHistoryDto.java
mikehalmamoj: A small thing but we don't generally use the Dto suffix on the API model - it's implied.

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderTransactionHistoryDto.java
```diff
@@ -0,0 +1,54 @@
...
+    @ApiModelProperty(value = "Transaction Id", example = "1", position = 2)
+    private Long transactionId;
+
+    @ApiModelProperty(value = "Transaction Sequence", example = "1", position = 3)
+    private Long transactionEntrySequence;
+
+    @ApiModelProperty(value = "Transaction Date", example = "2020-12-11", position = 4)
+    private LocalDate entryDate;
+
+    @ApiModelProperty(value = "Transaction Type", example = "???", position = 5)
```
mikehalmamoj, line 36: Just a reminder to fill in the missing examples when you can (and for referenceNumber and accountType)

src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
```diff
@@ -215,4 +216,21 @@ OffenderDamageObligationResponse getOffenderDamageObligations(
...
+
+    String NOMS_ID_REGEX_PATTERN = "[a-zA-Z][0-9]{4}[a-zA-Z]{2}";
+
+    @SuppressWarnings("RestParamTypeInspection")
+    @GetMapping("/{offenderNo}/transaction-history")
+    @ApiOperation(value = "Retrieve an offender's financial transaction history for cash, spends or savings.",
+            notes = "Transactions are returned in order of entryDate descending and sequence ascending).<br/>" +
+                    "All transaction amounts are represented as pence values.")
+    @ApiResponses(value = {
+            @ApiResponse(code = 400, message = "Not a digital offender. Offender not found. Offender has no account at this prison.", response = ErrorResponse.class),
```
mikehalmamoj, line 228: Should `Offender not found` be removed from here as it is covered by the 404?

src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
mikehalmamoj: Hmmm a noble aim but I'm not sure it's a good enough reason to lose the typing on the API.  Maybe there's another way to improve the error message?  Happy to assist.

src/main/java/uk/gov/justice/hmpps/prison/repository/OffenderTransactionHistoryRepository.java
```diff
@@ -0,0 +1,17 @@
...
+import org.springframework.data.repository.CrudRepository;
+import uk.gov.justice.hmpps.prison.repository.jpa.model.OffenderTransactionHistory;
+
+import java.time.LocalDate;
+import java.util.List;
+
+public interface OffenderTransactionHistoryRepository extends CrudRepository<OffenderTransactionHistory, String> {
+
+    @Query("select h from OffenderTransactionHistory h where h.offenderId = ?1 and h.accountType = ?2 and (h.entryDate between ?3 and ?4)")
+    List<OffenderTransactionHistory> findForGivenAccountType(Long offenderId, String accountType, LocalDate fromDat, LocalDate toDate);
```
mikehalmamoj, line 13: Typo on `fromDat` and on next method

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderTransactionHistoryService.java
mikehalmamoj: There are definitely merits to your argument but looking through prison-api it seems that a pragmatic approach has been adopted to treat the DB layer as the domain layer and we definitely don't want that leaking into the web layer... Plus it would take a lot of work to introduce a domain model between the web and DB layers now.  If we were designing from scratch - it would be an interesting conversation :)

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderTransactionHistoryService.java
mikehalmamoj: I see your point but seems overkill if you're using an Optional?

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderTransactionHistoryService.java
```diff
@@ -0,0 +1,84 @@
...
+        var fromDate = fromDateOpl.orElse(LocalDate.now());
+        var toDate = toDateOpl.orElse(LocalDate.now());
+
+        var now = LocalDate.now();
+        checkState(fromDate.isBefore(toDate) || fromDate.isEqual(toDate), "toDate can't be before fromDate");
+        checkState(fromDate.isBefore(now) || fromDate.isEqual(now), "fromDate can't be in the future");
+        checkState(toDate.isBefore(now) || toDate.isEqual(now), "toDate can't be in the future");
+
+        if(accountCodeOpl.isPresent()) {
+            boolean isAccountCodeExists = accountCodeOpl.map(AccountCode::byCodeName).filter(opl -> opl.isPresent()).isPresent();
```
mikehalmamoj, line 61: I found this hard to read - could it be pushed into the AccountCode enum instead as method `exists` or similar?

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderTransactionHistoryService.java
```diff
@@ -0,0 +1,84 @@
...
+                .reversed();
+
+        Collections.sort(histories, sortPolicy);
+
+        return histories.stream()
+                .map(h -> Pair.of(h, apiCurrency))
+                .map(OffenderTransactionHistoryTransformer::transform)
+                .collect(Collectors.toList());
+    }
+}
```
mikehalmamoj, line 84: This method is a bit long - could it be broken up?

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderTransactionHistoryService.java
jamesak-oncloud9: thanks Mike, not using optional for offenderId though

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderTransactionHistoryResourceTest.java
mikehalmamoj: Should it be Not Found instead?  I've heard arguments for and against not found vs empty list and so it's debatable, but I think in prison-api historically we've gone for a Not Found.

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderTransactionHistoryService.java
jamesak-oncloud9: I've moved out the sort policy and it reduced it a bit but will need significant rework that makes sense to reduce it further

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderTransactionHistoryDto.java
jamesak-oncloud9: removing Dto will result in a fully qualified class name in the service class uk.gov.justice.hmpps.prison.api.model.OffenderTransactionHistory as the entity has the same name uk.gov.justice.hmpps.prison.repository.jpa.model.OffenderTransactionHistory, is that okay?

mikehalmamoj: Good point.  Yes that's been done previously so it is okay but I understand if you want to avoid having to include package names in code.  If only we were using Kotlin :)

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderTransactionHistoryService.java
mikehalmamoj: It's just hard to see what is happening at a glance, so I'd be tempted to chop it up into at least 2 private methods (`validate` and `getSortedHistories`?).  But maybe that's just personal preference...

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderTransactionHistoryResourceTest.java
jamesak-oncloud9: so if no records are found it should return HTTP 404? isn't that simple wrong? because the resource does exist it just doesn't have any records associated with it? just checking before I make the change

mikehalmamoj: Sorry, should have been clearer.
Offender not found - 404
Offender found but no transaction history - 200 with empty list

jamesak-oncloud9: so no change required because as it is now if you pass an offender number that doesn't exist then it'll return an empty list and if you pass a non-number value then you get an error, agree?

mikehalmamoj: > I think it's missing a test case for the Offender not found.

> it's an empty result set for which there is a test

So do we need an a test for Offender not found bearing in mind it's not an empty result set?

jamesak-oncloud9: so how is it not found and also not empty :)

jamesak-oncloud9: that can only happen if the caller uses the wrong id number from the one they intended, which we can't detect, is my point

mikehalmamoj: There is no test for a 404 response.

jamesak-oncloud9: sure, my question is how would you detect or test for that?

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderTransactionHistoryDto.java
jamesak-oncloud9: done

src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
jamesak-oncloud9: done

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderTransactionHistoryService.java
jamesak-oncloud9: done

src/main/java/uk/gov/justice/hmpps/prison/service/transformers/OffenderTransactionHistoryTransformer.java
```diff
@@ -0,0 +1,28 @@
+package uk.gov.justice.hmpps.prison.service.transformers;
+
+import org.springframework.data.util.Pair;
+import uk.gov.justice.hmpps.prison.api.model.OffenderTransactionHistoryDto;
+
+import static uk.gov.justice.hmpps.prison.util.MoneySupport.poundsToPence;
+
+public class OffenderTransactionHistoryTransformer {
```
JamieDraperUK, line 8: One for later but could you use a static import for OffenderTransactionHistory?

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderTransactionHistoryResourceTest.java
JamieDraperUK: See offender damage obligations resource test, I recently added a test for an offender not found and an offender not in the current user's caseload, both return a 404.

jamesak-oncloud9: done

src/main/java/uk/gov/justice/hmpps/prison/repository/OffenderTransactionHistoryRepository.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.justice.hmpps.prison.repository;
+
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.CrudRepository;
+import uk.gov.justice.hmpps.prison.repository.jpa.model.OffenderTransactionHistory;
+
+import java.time.LocalDate;
+import java.util.List;
+
+public interface OffenderTransactionHistoryRepository extends CrudRepository<OffenderTransactionHistory, String> {
```
JamieDraperUK, line 10: I missed this before, but should this be:
`CrudRepository<OffenderTransactionHistory, PK> `

src/test/java/uk/gov/justice/hmpps/prison/service/OffenderTransactionHistoryServiceTest.java
```diff
@@ -0,0 +1,303 @@
...
+            final Optional<String> accountCode = Optional.of("spends");
+            final Optional<LocalDate> fromDateOpl = Optional.of(LocalDate.now().plusDays(1));
+            final Optional<LocalDate> toDateOpl = Optional.of(LocalDate.now().plusDays(2));
+            service.getTransactionHistory(offenderId, accountCode, fromDateOpl, toDateOpl);
+        });
+
+        assertEquals("fromDate can't be in the future", exception.getMessage());
+    }
+
+    @Test()
```
JamieDraperUK, line 162: Did you mean to leave the parenthesis in?

JamieDraperUK (approved):
LGTM

src/test/java/uk/gov/justice/hmpps/prison/service/OffenderTransactionHistoryServiceTest.java
```diff
@@ -0,0 +1,316 @@
...
+
+    @BeforeEach
+    public void setUp() {
+        service = new OffenderTransactionHistoryService("GBP", repository, offenderRepository);
+    }
+
+    @Test
+    public void When_getTransactionHistory_And_AccountCodeIsGiven_Then_CallRepositoryWithAccountCode() {
+
+        final Optional<String> accountCode  = Optional.of("spends");
```
JamieDraperUK, line 51: One for a later PR but the convention is to use final var unless it's not obvious what the type is.

src/test/java/uk/gov/justice/hmpps/prison/service/OffenderTransactionHistoryServiceTest.java
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    @Test
+    public void When_getTransactionHistory_And_AccountCodeIsMissing_Then_CallRepositoryWithoutAccountCode() {
+
+        final Optional<String> accountCode  = Optional.empty();
+        final Optional<LocalDate> fromDateOpl = Optional.of(LocalDate.now().minusDays(7));
+        final Optional<LocalDate> toDateOpl = Optional.of(LocalDate.now());
+
+        final List<uk.gov.justice.hmpps.prison.repository.jpa.model.OffenderTransactionHistory> txnItem = Collections.emptyList();
```
JamieDraperUK, line 73: Use static imports.

JamieDraperUK (approved):
:+1:

PR #754, opened 2020-11-02
Nn 3013 return bed assignment history sequence for cell history

JamieDraperUK:


src/main/java/uk/gov/justice/hmpps/prison/api/model/BedAssignment.java
```diff
@@ -19,13 +23,33 @@
 @ToString
 @Data
 public class BedAssignment {
+    @ApiModelProperty(value = "Unique, numeric booking id.", position = 1, example = "1234134")
```
petergphillips, line 26: Is this unique - thought it had to be combination of `bookingId` and `badAssignmentHistorySequence`?

petergphillips (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/api/model/BedAssignment.java
JamieDraperUK: Good point, I'll update the docs :+1:

petergphillips (approved):
LGTM

PR #761, opened 2020-11-11
Dt 1247 use ojdbcjar gradle dependency

dnyanesh-sonwane:
Removed jdbc jar from lib folder and added as gradle dependency


petergphillips (approved):
LGTM

PR #762, opened 2020-11-12
NN-3038: insert reference codes

jamesak-oncloud9:


src/test/java/uk/gov/justice/hmpps/prison/repository/ReferenceDataRepositoryTest.java
```diff
@@ -58,6 +60,31 @@ public void testGetReferenceCodeByDomainAndCodeWithSubCodes() {
         assertThat(refCode.get().getSubCodes()).isNotEmpty();
     }
 
+    @ParameterizedTest(name = "{index} => domain={0}, code={1}, withSubCodes={2}")
```
petergphillips, line 63: Pointless test.  You're testing that the *test* reference data has been asserted correctly

src/test/java/uk/gov/justice/hmpps/prison/repository/ReferenceDataRepositoryTest.java
```diff
@@ -58,6 +60,31 @@ public void testGetReferenceCodeByDomainAndCodeWithSubCodes() {
...
+        "TASK_SUBTYPE, CON, false",
+        "TASK_SUBTYPE, LN, false",
+        "TASK_SUBTYPE, VP, false"
+    })
+    public void testGetReferenceCodeByDomainAndCode(String domain, String code, boolean withSubCodes) {
+
+        final var refCode =
+            repository.getReferenceCodeByDomainAndCode(domain, code, withSubCodes);
+
+        assertThat(refCode.isPresent()).isTrue();
```
petergphillips, line 78: should be 
```
assertThat(refCode).isPresent();
```

src/test/java/uk/gov/justice/hmpps/prison/repository/ReferenceDataRepositoryTest.java
```diff
@@ -77,7 +104,7 @@ public void testGetReferenceCodesByDomainWithoutSubCodes1() {
         assertThat(refCodes.getItems()).isNotNull();
         assertThat(refCodes.getItems()).isNotEmpty();
         assertThat(refCodes.getItems()).hasSize(10);
-        assertThat(refCodes.getTotalRecords()).isEqualTo(58);
+        assertThat(refCodes.getTotalRecords()).isEqualTo(59);
```
petergphillips, line 107: not your fault, but this is all really naff.  We shouldn't have to fix lots of tests when we want to add in more test reference data :angry:

src/test/java/uk/gov/justice/hmpps/prison/repository/ReferenceDataRepositoryTest.java
jamesak-oncloud9: How can I make it better, please? the test verifies these codes exist and checks which ones have subcodes. I was told to look at the existing test code and follow the same style, existing tests in that file do the same type of checks, and inserting these records impacted the expectation of sorting logic results in the right way. But happy to make it better please let me know how. Thank you

src/test/java/uk/gov/justice/hmpps/prison/repository/ReferenceDataRepositoryTest.java
jamesak-oncloud9: I was thinking the same. I could assert the result with the json form saved to file, would that be better? or verify results are sorted as expected with any result set so that no code change is required when new records are inserted but would fail if the sorting is incorrect. Would that be better?

jamesak-oncloud9: this style is used in other places, for financial transactions for example, and related tables, on a previous ticket I inserted a few new rows and it impacted many tests (counts and totals, etc) so it makes inserting new custom records harder and more time-consuming.

petergphillips (approved):
LGTM

PR #764, opened 2020-11-16
NN-3041: include create-user-id

jamesak-oncloud9:


src/main/java/uk/gov/justice/hmpps/prison/api/model/BedAssignment.java
```diff
@@ -52,4 +52,7 @@
 
     @ApiModelProperty(value = "Bed assignment sequence. Used as a primary key when combined with the booking id", position = 10, example = "2")
     private Integer bedAssignmentHistorySequence;
+
+    @ApiModelProperty(value = "the staff member responsible for the movement of a prisoner", position = 11, example = "KQJ74F")
+    private String createUserId;
```
petergphillips, line 57: Is this the `staffId` or `username`?  If the latter can it be called `createUsername` instead please?  If the former then the `example` would need to be updated

jamesak-oncloud9: it is an Id value and not a user name value, I got the value 'KQJ74F' from the data table, please suggest a better example and I'll be happy to include it

jamesak-oncloud9: shall I use 'SA' as an example?

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/ReferenceDataResourceTest.java
```diff
@@ -1,38 +1,51 @@
...
 
 import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.core.ParameterizedTypeReference;
 import org.springframework.http.HttpMethod;
+import org.springframework.jdbc.core.JdbcTemplate;
 import uk.gov.justice.hmpps.prison.executablespecification.steps.AuthTokenHelper;
 
 import static net.javacrumbs.jsonunit.assertj.JsonAssertions.assertThatJson;
+import static org.assertj.core.api.Assertions.*;
```
Nimphal, line 11: I think this wildcard slipped in :)

petergphillips (approved):
LGTM

PR #766, opened 2020-11-19
DCS-685 Improve custom events for create/delete appointment.

bell-pa:
getBookingAppointment does not populate eventLocationId but this should be part of the create and delete appointment custom event data.
* Use getBookingAppointmentByEventId instead
* Remove unused getBookingAppointment.


andrewrlee (approved):
LGTM

PR #768, opened 2020-11-20
update gradle plugin and run ktlintFormat

rgforsyth:
update to version 1.1.2 of the gradle springboot plugin and run the kotlin linter.


.editorconfig
```diff
@@ -0,0 +1,24 @@
+# WARNING - THIS FILE WAS GENERATED BY THE dps-gradle-spring-boot GRADLE PLUGIN
```
petergphillips, line 1: may be better to add this to .gitignore and then will be created fresh each time?

petergphillips (dismissed):
LGTM

petergphillips (approved):
LGTM

PR #769, opened 2020-11-20
NN-2462: include damage obligations balance

jamesak-oncloud9:
Screenshot 2020-11-20 at 14 39 38


src/main/java/uk/gov/justice/hmpps/prison/api/model/Account.java
```diff
@@ -39,6 +42,11 @@
     @NotNull
     private BigDecimal savings;
 
+    @ApiModelProperty(required = true, value = "Damage obligation balance.")
+    @NotNull
+    @JsonDeserialize(using = MoneyDeserializer.class)
```
jamesak-oncloud9, line 47: should we make sure the other amounts are in the correct format?

src/main/java/uk/gov/justice/hmpps/prison/service/FinanceService.java
```diff
@@ -111,4 +128,17 @@ private void validateTransferToSavings(final String prisonId, final String offen
...
+        return pair.getFirst();
+    };
+
+    private static final Function<List<OffenderDamageObligationModel>, BigDecimal> sumAmountToPay = (obligations) -> obligations
+        .stream()
+        .map(OffenderDamageObligationModel::getAmountToPay)
+        .reduce(BigDecimal.ZERO, BigDecimal::add);
+
+    private static final Function<OffenderDamageObligationService, Function<String, BigDecimal>> createSumDamageObligationToPay =
+        (service) -> (Function<String, BigDecimal>) offenderNo -> sumAmountToPay.apply(service.getDamageObligations(offenderNo,""));
```
jamesak-oncloud9, line 143: should this include all status types?

src/main/java/uk/gov/justice/hmpps/prison/api/model/Account.java
jamesak-oncloud9: [example](blob:https://dsdmoj.atlassian.net/e872b7af-ec6d-4023-90f9-c128a85c11ee#media-blob-url=true&id=e3944a83-063c-429b-a5cb-4007fe08b737&collection=&contextId=76287&mimeType=image%2Fpng&name=Screenshot%202020-11-20%20at%2014.39.38.png&size=74001&width=1929&height=137)

JamieDraperUK, line 47: I might be missing something here but what's the reason behind introducing MoneyDeserializer here for damageObligations when the others don't require one?

jamesak-oncloud9: wasn't sure if we needed an explicit formatter to ensure the correct scale is set, is this done elsewhere in the app? tried to look for another deserializer couldn't find one so included to trigger a discussion about this, if it's not needed happy to remove it

src/main/java/uk/gov/justice/hmpps/prison/service/FinanceService.java
```diff
@@ -41,11 +48,21 @@
     private final AccountCodeRepository accountCodeRepository;
     private final OffenderSubAccountRepository offenderSubAccountRepository;
     private final OffenderTrustAccountRepository offenderTrustAccountRepository;
+    private final OffenderDamageObligationService offenderDamageObligationService;
 
     @VerifyBookingAccess(overrideRoles = {"SYSTEM_USER", "GLOBAL_SEARCH"})
     public Account getBalances(final Long bookingId) {
+
+        var damageObligationBalance = bookingRepository.getLatestBookingByBookingId(bookingId)
```
jamesak-oncloud9, line 56: all I need is the offenderNo is using this method 'getLatestBookingByBookingId' a good way to get that value?

src/main/java/uk/gov/justice/hmpps/prison/api/model/Account.java
JamieDraperUK: Not entirely sure without looking but I did wonder how the following values worked without needing a MoneyDeserializer:

`@ApiModelProperty(required = true, value = "Spends sub account balance.")
    @NotNull
    private BigDecimal spends;

    @ApiModelProperty(required = true, value = "Cash sub account balance.")
    @NotNull
    private BigDecimal cash;

    @ApiModelProperty(required = true, value = "Saves sub account balance.")
    @NotNull
    private BigDecimal savings;`

jamesak-oncloud9: prisonstaffhub/prisonerQuickLook.js uses formatCurrency, but since these are values from the domain the service should return a correct money value regardless of what happens in the view layer

JamieDraperUK: My point is it looks inconsistent now, would you agree?

jamesak-oncloud9: we can use the deserializer on the other values which is my preferred option as it's more correct in my view or remove it and rely on the view layer formatting.

jamesak-oncloud9: the other option is to remove the deserializer and make sure the service class returns a correct many value bigdecimal

src/main/java/uk/gov/justice/hmpps/prison/service/FinanceService.java
```diff
@@ -41,11 +48,21 @@
...
+            .map(OffenderSummary::getOffenderNo)
```
JamieDraperUK, line 57: There is a fair amount of indirection going on here and it makes the code hard to comprehend. Would it be worth breaking it out a bit?  It could possibly be simplified without the need for all those **private static final Function<>**

jamesak-oncloud9: happy to simplify, I think setDamageObligation shows clear intentions but I've refactored it to 
```
private static final Function<Pair<Account, BigDecimal>, Account> setDamageObligation = (pair) -> {
        val account = pair.getFirst();
        val damageObligationAmount = pair.getSecond();
        account.setDamageObligations(damageObligationAmount);
        return account;
    };
```
Does this look better?

I'm trying to avoid multi-line code blocks that are less expressive and make the reader put more effort to understand the logic.

JamieDraperUK: I think sometimes multi line code blocks can help. Especially to help keep logical steps separate. This hasn't been tested and you could probably making it even neater but, something like the following means the reader doesn't have to chase around to understand what's going on:

`final var booking = bookingRepository.getLatestBookingByBookingId(bookingId).orElseThrow(new EntityNotFoundException("Booking not found for x"));

        final var damageObligationBalance =
            offenderDamageObligationService.getDamageObligations(booking.getOffenderNo(), "")
                .stream()
                .map(OffenderDamageObligationModel::getAmountToPay)
                .reduce(BigDecimal.ZERO, BigDecimal::add);

        return financeRepository.getBalances(bookingId, booking.getAgencyLocationId())
            .toBuilder()
            .damageObligations(damageObligationBalance)
            .build();`

src/main/java/uk/gov/justice/hmpps/prison/api/model/Account.java
JamieDraperUK: @jamesak-oncloud9 Both options are good as long as it's consistent.

src/main/java/uk/gov/justice/hmpps/prison/service/FinanceService.java
JamieDraperUK: That's a good question. Should we only be counting active damage obligations?

src/main/java/uk/gov/justice/hmpps/prison/service/FinanceService.java
```diff
@@ -41,12 +47,34 @@
...
+            .orElse(zeroBalances());
+    }
+
+    private Account zeroBalances() {
+        return Account.builder()
+            .spends(toMoney("0.00"))
+            .cash(toMoney("0.00"))
+            .savings(toMoney("0.00"))
+            .damageObligations(toMoney("0.00"))
+            .currency("£").build();
```
JamieDraperUK, line 77: I think that needs to be GBP. Best to pull it from properties - @Value("${api.currency}")

src/main/java/uk/gov/justice/hmpps/prison/service/FinanceService.java
```diff
@@ -41,12 +47,34 @@
...
     public Account getBalances(final Long bookingId) {
-        return bookingRepository.getBookingAgency(bookingId)
-                .map(agency -> financeRepository.getBalances(bookingId, agency))
-                .orElse(null);
+
+        final var offenderSummary = bookingRepository.getLatestBookingByBookingId(bookingId)
+            .orElseThrow(new EntityNotFoundException("Booking not found for id: " + bookingId));
+
+        final var damageObligationBalance =
+            offenderDamageObligationService.getDamageObligations(offenderSummary.getOffenderNo(), "")
```
JamieDraperUK, line 59: Does this include only active damage obligations?

src/test/java/uk/gov/justice/hmpps/prison/service/FinanceServiceTest.java
```diff
@@ -251,4 +261,126 @@ private TransferTransaction createTransferTransaction() {
                 .clientTransactionId("transId")
                 .build();
     }
+
+    @Test
+    public void test_getBalances_HappyPath() {
+
+        var bookingId = -1L;
```
JamieDraperUK, line 268: Minor thing but these should be **final** to ensure consistency with the rest of the file.

src/main/java/uk/gov/justice/hmpps/prison/api/model/Account.java
jamesak-oncloud9: done

src/main/java/uk/gov/justice/hmpps/prison/service/FinanceService.java
jamesak-oncloud9: done

src/main/java/uk/gov/justice/hmpps/prison/service/FinanceService.java
jamesak-oncloud9: done

src/main/java/uk/gov/justice/hmpps/prison/api/model/Account.java
jamesak-oncloud9: done

src/main/java/uk/gov/justice/hmpps/prison/service/FinanceService.java
jamesak-oncloud9: done

src/main/java/uk/gov/justice/hmpps/prison/service/FinanceService.java
jamesak-oncloud9: done

src/main/java/uk/gov/justice/hmpps/prison/service/FinanceService.java
jamesak-oncloud9: done

src/test/java/uk/gov/justice/hmpps/prison/service/FinanceServiceTest.java
jamesak-oncloud9: done

src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
```diff
@@ -334,7 +335,7 @@ public PrivilegeSummary getLatestBookingIEPSummaryForOffender(@NotNull @PathVari
     @ApiOperation("Return a list of damage obligations")
     @GetMapping("/{offenderNo}/damage-obligations")
     public OffenderDamageObligationResponse getOffenderDamageObligations(@NotNull @PathVariable("offenderNo") @ApiParam(value = "offenderNo", required = true, example = "A1234AA") final String offenderNo, @RequestParam(value = "status", required = false, defaultValue = "") @ApiParam(value = "Filter by obligation status. Leave blank to return all", required = false, example = "ACTIVE", allowableValues = "INACT,PAID,ONH,ACTIVE,APPEAL") final String status) {
-        final var damageObligations = offenderDamageObligationService.getDamageObligations(offenderNo, status);
+        final var damageObligations = offenderDamageObligationService.getDamageObligations(offenderNo, Status.forCode(status));
```
JamieDraperUK, line 338: This could be replace with Status.valueOf has the Enum type has this functionality built in by default. 

Also, one for another PR but you could add a Converter so that spring takes care for deserializing for you.

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderDamageObligation.java
```diff
@@ -56,6 +57,24 @@
...
+        ACTIVE("ACTIVE"), PAID("PAID"), INACT("INACT"),
+        ONH("ONH"), APPEAL("APPEAL"), ALL("");
+        private String code;
+        Status(String code) {
+            this.code = code;
+        }
+        public String code() {
+            return code;
+        }
+        public static Status forCode(String code) {
```
JamieDraperUK, line 71: Not needed

src/main/java/uk/gov/justice/hmpps/prison/service/FinanceService.java
```diff
@@ -21,32 +25,60 @@
...
+
+        return Optional.ofNullable(financeRepository.getBalances(bookingId, offenderSummary.getAgencyLocationId()))
+            .map(Account::toBuilder)
+            .map(builder -> builder.damageObligations(toMoneyScale(damageObligationBalance)))
+            .map(AccountBuilder::build)
+            .orElse(defaultBalances());
+    }
+
+    private Account defaultBalances() {
+        val zero = toMoney("0.00");
```
JamieDraperUK, line 78: Should be final

src/main/java/uk/gov/justice/hmpps/prison/util/MoneySupport.java
```diff
@@ -8,14 +14,28 @@
...
+    }
+
+    public static BigDecimal toMoney(final String amount) {
+        return toMoneyScale(new BigDecimal(amount));
+    }
+
+    public static class MoneyDeserializer extends JsonDeserializer<BigDecimal> {
+        private NumberDeserializers.BigDecimalDeserializer delegate = NumberDeserializers.BigDecimalDeserializer.instance;
+        @Override
+        public BigDecimal deserialize(JsonParser parser, DeserializationContext context) throws IOException {
```
JamieDraperUK, line 37: Missing finals

JamieDraperUK (dismissed):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
jamesak-oncloud9: the forCode has custom logic which valueOf doesn't offer
   ```
      public static Status forCode(String code) {
            return Arrays.stream(Status.values())
                .filter(c -> c.equals(code.toUpperCase()))
                .findFirst()
                .orElse(ALL);
        }
```
how would valueOf default to ALL for an empty string?

File: src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderDamageObligation.java (hunk 9 above)
Comment (jamesak-oncloud9) [reply src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderDamageObligation.java:71]: it defaults to ALL for an empty string

File: src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java (hunk 8 above)
Comment (jamesak-oncloud9) [reply src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java:338]: done

File: src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderDamageObligation.java (hunk 9 above)
Comment (jamesak-oncloud9) [reply src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderDamageObligation.java:71]: done

File: src/main/java/uk/gov/justice/hmpps/prison/service/FinanceService.java (hunk 10 above)
Comment (jamesak-oncloud9) [reply src/main/java/uk/gov/justice/hmpps/prison/service/FinanceService.java]: done

File: src/main/java/uk/gov/justice/hmpps/prison/util/MoneySupport.java (hunk 11 above)
Comment (jamesak-oncloud9) [reply src/main/java/uk/gov/justice/hmpps/prison/util/MoneySupport.java]: done

File: src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java (hunk 8 above)
Comment (JamieDraperUK) [reply src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java:338]: One way would to set a default value for the controller status parameter. Then you would just use Status.valueOf(status).

Comment (JamieDraperUK) [reply src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java:338]: For example just set the default to 'ALL'

**Review by JamieDraperUK — DISMISSED** [review]
LGTM. One minor comment around not needing the custom String to Enum function if you defaulted the controller param value to "ALL"

File: src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java (hunk 8 above)
Comment (jamesak-oncloud9) [reply src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java]: done

**Review by JamieDraperUK — DISMISSED** [review]
LGTM

File: src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java [hunk 12]
```diff
@@ -333,12 +335,18 @@ public PrivilegeSummary getLatestBookingIEPSummaryForOffender(@NotNull @PathVari
...
-        final var damageObligations = offenderDamageObligationService.getDamageObligations(offenderNo, status);
-
+    public OffenderDamageObligationResponse getOffenderDamageObligations(@NotNull @PathVariable("offenderNo") @ApiParam(value = "offenderNo", required = true, example = "A1234AA") final String offenderNo, @RequestParam(value = "status", required = false, defaultValue = "ALL") @ApiParam(value = "Filter by obligation status. Leave blank to return all", required = false, example = "ACTIVE", allowableValues = "INACT,PAID,ONH,ACTIVE,APPEAL") final String status) {
+        final var damageObligations = offenderDamageObligationService.getDamageObligations(offenderNo, lookupStatusOrDefaultToAll(status));
         return new OffenderDamageObligationResponse(damageObligations);
     }
 
+    private Status lookupStatusOrDefaultToAll(final String status) {
+        return Arrays.stream(Status.values())
```
jamesak-oncloud9, line 344: ideally, this bit of code should go back in the enum, how it as done the first time, it's neater that way

JamieDraperUK (approved):
LGTM

PR #771, opened 2020-11-23
DT-1333 Replaced spring profile include with profile groups

dnyanesh-sonwane:
Replaces profile include with profile groups


Mjwillis (approved):
LGTM

PR #773, opened 2020-11-25
Nn 3018 prison api allow moved cell

JamieDraperUK:
A set of new/special-case case notes have been created to be used when making cell moves. These new case notes help document the reason for making a cell move.

These don't typical look like case note types which meant that the CaseNoteTypeSubTypeValidator failed. 

I've turned these case note types/cell move reasons into an Enum to help document the data a bit and help provide a way to trigger a specific validation for these types.


src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
```diff
@@ -238,6 +238,7 @@ public AdjudicationDetail getAdjudication(@PathVariable("offenderNo") @ApiParam(
                     nvl(pageLimit, 10L));
 
             return ResponseEntity.ok()
+                    .headers(pagedCaseNotes.getPaginationHeaders())
```
mikehalmamoj, line 241: Typo? (duplicated below)

JamieDraperUK: Ah, nice spot. I think it was a merge fail.

JamieDraperUK: Fixed :+1:

mikehalmamoj (approved):

src/main/java/uk/gov/justice/hmpps/prison/service/validation/MovedCell.java
```diff
@@ -0,0 +1,18 @@
...
+
+      ADM - Administrative
+      BEH - Behaviour
+      CLA - Classification or re-classification
+      CON - Conflict with other prisoners
+      LN -  Local needs
+      VP - Vulnerable prisoner
+*/
+public enum MovedCell {
+    ADM, BEH, CLA, CON, LN, VP
```
Mjwillis, line 17: Why is this hard coded?

JamieDraperUK: The idea is to document them as don't behave like the other types.

PR #776, opened 2020-11-26
DCS-731 Improve custom events for creating and deleting appointments.

bell-pa:


src/main/java/uk/gov/justice/hmpps/prison/repository/sql/BookingRepositorySql.kt
```diff
@@ -573,6 +574,8 @@ enum class BookingRepositorySql(val sql: String) {
                LEFT JOIN AGENCY_INTERNAL_LOCATIONS AIL ON OIS.TO_INTERNAL_LOCATION_ID = AIL.INTERNAL_LOCATION_ID
                LEFT JOIN AGENCY_LOCATIONS AGY ON OIS.TO_AGY_LOC_ID = AGY.AGY_LOC_ID
          WHERE OIS.EVENT_ID = :eventId
+               AND EVENT_CLASS = 'INT_MOV'
+               AND EVENT_TYPE = 'APP'
```
andrewrlee, line 578: is this definitely only used by us? It's not going to break anything else is it?

bell-pa: I added that query a few days ago.
It is only used by AppointmentsService:
* createBookingAppointment - uses the insert statement on line 582 which always sets EVENT_CLASS = 'INT_MOV' and EVENT_TYPE = 'APP'.
* getBookingAppointment added by me today. Only called from AppointmentResource for GET /appointments/{appointmentId}
* deleteBookingAppointment added by me a few days ago.  Only called from AppointmentResource DELETE /appointments/{appointmentId}
I'm being cautious.  There are other kinds of data in the OFFENDER_IND_SCHEDULES table - where EVENT_CLASS and EVENT_TYPE have other values. The extra restrictions in the query above ensures that the operations only act on 'appointments' even if the ID is valid.

petergphillips (approved):
LGTM

PR #785, opened 2020-12-08
DT-1389: ✨ Capitalise staff first and last name

petergphillips:


JamieDraperUK (approved):
:+1:

src/main/java/uk/gov/justice/hmpps/prison/api/model/UserDetail.java
```diff
@@ -73,4 +74,11 @@ public boolean isActive() {
         return "ACTIVE".equals(accountStatus);
     }
 
+    public String getFirstName() {
+        return WordUtils.capitalizeFully(firstName);
+    }
+
+    public String getLastName() {
+        return WordUtils.capitalizeFully(lastName);
```
andymarke, line 82: I wonder if we should try to be clever with double barrelled names ...

petergphillips: We weren't being clever in manage auth accounts on double barrelled names or names like O'Brian, so just added it as is.  We would never get it right for all users, but maybe at some point we should perhaps just convert the data in NOMIS instead.  Agreed though, might be good to consider a better conversion, since it is naff that we can't even get their names right.

PR #786, opened 2020-12-08
DT-1391 transfer out

Mjwillis:


mikehalmamoj (approved):

src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerReleaseAndTransferService.java
```diff
@@ -0,0 +1,238 @@
...
+        booking.setLivingUnitMv(null);
+        booking.setAssignedLivingUnit(null);
+        booking.setLocation(toLocation);
+        booking.setCreateLocation(toLocation);
+        booking.setStatusReason(TRN.getCode() + "-" + requestToTransferOut.getTransferReasonCode());
+        booking.setCommStatus(null);
+    }
+
+    private LocalDateTime getAndCheckMovementTime(final LocalDateTime movementTime, final Long bookingId) {
+        final var now = LocalDateTime.now();
```
matt-crowson-moj, line 125: So from memory there is Clock bean in the API.  So something like this `LocalDate.now(clock)` makes it testable as well.

PR #791, opened 2020-12-14
DT-1410 Transfer a prisoner IN to prison from previous transfer out.

Mjwillis:


src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerReleaseAndTransferService.java
```diff
@@ -135,6 +152,79 @@ public void transferOutPrisoner(final String prisonerIdentifier, final RequestTo
         booking.setCommStatus(null);
     }
 
+    public void transferInPrisoner(final String prisonerIdentifier, final RequestToTransferIn requestToTransferIn) {
```
mikehalmamoj, line 155: Do we need a transaction on this?

mikehalmamoj (approved):

src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerReleaseAndTransferService.java
Mjwillis: Already one at the top

PR #792, opened 2020-12-16
DT-1420 junit 4

mikehalmamoj:


build.gradle
```diff
@@ -92,7 +92,7 @@ dependencies {
 task testIntegration(type: Test) {
     include '**/*IntTest*'
     useJUnitPlatform()
-    minHeapSize = "1024m"
+    minHeapSize = "128m"
```
mikehalmamoj, line 95: Not sure this makes any difference but the tests are currently stable so :crossed_fingers:

src/test/resources/features/booking_search.feature
```diff
@@ -13,8 +13,8 @@ Feature: Booking Search
 
   Scenario: Search all offenders
     When a booking search is made without any criteria
-    Then "10" booking records are returned
-    And  "35" total booking records are available
+    Then some booking records are returned
```
mikehalmamoj, line 16: If we do care about these totals then using shared test data is not the way to do it - so I'm assuming not.

src/test/resources/features/booking_search.feature
```diff
@@ -13,8 +13,8 @@ Feature: Booking Search
 
   Scenario: Search all offenders
     When a booking search is made without any criteria
-    Then "10" booking records are returned
```
mikehalmamoj, line 16: If we do care about these numbers then using shared test data is not the way to achive this, so I'm assuming not.

src/test/resources/features/reference_domains.feature
```diff
@@ -37,14 +37,14 @@ Feature: Reference Domains
 
   Scenario: Retrieve used case note types with sub-types
     When request submitted to retrieve used case note types
-    Then "31" reference code items are returned
-    And domain for all returned items is "TASK_TYPE"
-    And description for "1st" returned item is "Accredited Programme"
```
mikehalmamoj, line 42: If we do care about the ordering of the results then using shared data is not the way to test this, so I've removed all positional indexes from this test.

andrewrlee (approved):
LGTM

PR #796, opened 2020-12-24
DCS-746 'Amend booking appointment comment' operation.

bell-pa:
`PUT /appointments/{appointmentId}/comment` with body that is a ,possibly empty, JSON string


petergphillips (approved):
LGTM

PR #799, opened 2021-01-04
DT-1452: 🔒 Suppress false positive for lang-tag

petergphillips:


petergphillips:
Closed as owasp now passing

PR #809, opened 2021-01-13
DT-1483 Add PNC check on long and short

Mjwillis:


src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerCreationService.java
```diff
@@ -124,4 +133,26 @@ public PrisonerIdentifier getNextPrisonerIdentifier() {
         }
         return PrisonerIdentifier.builder().id(currentSequence.getPrisonerIdentifier()).build();
     }
+
+    private String getShortPncNumber(String pncNumber) {
```
mikehalmamoj, line 137: Might be worth moving these into the Pnc class?

mikehalmamoj (approved):

PR #813, opened 2021-01-18
DT-1483 Corrections after testing an review.

Mjwillis:


src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerCreationService.java
```diff
@@ -64,12 +64,19 @@ public String createPrisoner(final RequestToCreate requestToCreate) {
                     throw new BadRequestException(format("Prisoner with PNC %s already exists with ID %s", longPnc, identifier.getOffender().getNomsId()));
                 });
 
+        } else if (StringUtils.isNotBlank(requestToCreate.getCroNumber())) {
```
mikehalmamoj, line 67: Should we check the CRO even if there is a PNC?

mikehalmamoj (dismissed):

src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerCreationService.java
Mjwillis: duh yes!

mikehalmamoj (approved):

PR #815, opened 2021-01-19
DT-1413 Correct issue with /movements where old bookings are not returned and movement date is not populated and null fields are returned.

Mjwillis:


andrewrlee (approved):
LGTM

PR #818, opened 2021-01-27
DT-1537 Review comments from first PR

Mjwillis:


src/main/java/uk/gov/justice/hmpps/prison/repository/sql/AdjudicationsRepositorySql.kt
```diff
@@ -50,6 +50,7 @@ enum class AdjudicationsRepositorySql(val sql: String) {
         SELECT  AL.AGY_LOC_ID AGENCY_ID
         ,       AL.AGENCY_LOCATION_TYPE AGENCY_TYPE
         ,       AL.DESCRIPTION
+        ,       CASE WHEN AL.ACTIVE_FLAG = 'Y' THEN 1 ELSE 0 END AS ACTIVE
```
Mjwillis, line 53: This mapping was missing so before they were always false

andymarke (approved):
LGTM

PR #821, opened 2021-01-28
DT-1537 - Test review fixes

Mjwillis:
- removed ability to set deactivation for now and is set by setting active to false to today’s date 
- Creating an existing agency now return 409 
- Fixed 201 response for create 
- Fixed long description to be not mandatory


src/test/resources/uk/gov/justice/hmpps/prison/api/resource/impl/single_agency_updated_inactive.json
```diff
@@ -0,0 +1,7 @@
+{
+  "agencyId": "LEI",
+  "description": "Leeds",
+  "agencyType": "INST",
+  "active": false,
+  "longDescription": "This is a prison based in Leeds"
```
mikehalmamoj, line 6: Not sure on this one, but should we return the deactivated date?

mikehalmamoj (approved):

src/test/resources/uk/gov/justice/hmpps/prison/api/resource/impl/single_agency_updated_inactive.json
Mjwillis: yes but need to fix the date first

PR #825, opened 2021-02-01
CAT-677: Use second latest assessment if latest is category U

connormaglynn:
Solution before:
- Remove earlier categorisations
- Remove all none standard categorisation (anything not B/C/D)

New solution:
- Remove category Us
- Remove earlier categorisations
- Remove all none standard categorisation (anything not B/C/D)

Category U's are used as a holding category, so the new solution ignores these in favour of an earlier assessment


steverendell (commented):
not sure about this , need to discuss perhaps

steverendell (changes requested):
blocked for now

steverendell (approved):
LGTM

PR #826, opened 2021-02-02
NN-3146 group transactions by account type

JamieDraperUK:


src/main/java/uk/gov/justice/hmpps/prison/service/OffenderTransactionHistoryService.java
```diff
@@ -81,30 +83,26 @@ private void validate(final String offenderNo, final String accountCode, final L
...
 
-        allOffenderTransactions.keySet().forEach(agencyId -> {
+        return transactions.keySet().stream().flatMap(agencyAccountType -> {
             final var runningBalance = new AtomicReference<>(BigDecimal.valueOf(0));
-            allOffenderTransactions.get(agencyId).forEach(transaction -> {
+
+            return transactions.get(agencyAccountType).stream().flatMap(transaction -> {
                 final var postingType = transaction.getPostingType();
                 if (postingType.equals("CR")) {
                     runningBalance.set(runningBalance.get().add(transaction.getEntryAmount()));
```
sp-watson, line 98: An alternative to using set/get on AtomicInt is to use updateAndGet (or getAndUpdate), e.g. runningBalance.updateAndGet((v) -> v.add(transaction.getEntryAmount()));.. but i am not sure it is worth the effort to change it as it is thread safe anyway!

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderTransactionHistoryService.java
```diff
@@ -81,30 +83,26 @@ private void validate(final String offenderNo, final String accountCode, final L
...
-        final var allOffenderTransactions = historyRepository
+        final var transactions = historyRepository
             .findByOffender_NomsId(offenderNo)
             .stream()
             .sorted(SORT_BY_OLDEST_DATE)
-            .collect(groupingBy(OffenderTransactionHistory::getAgencyId));
+            .collect(groupingBy(transaction -> Pair.of(transaction.getAgencyId(), transaction.getAccountType())));
 
-        allOffenderTransactions.keySet().forEach(agencyId -> {
+        return transactions.keySet().stream().flatMap(agencyAccountType -> {
```
sp-watson, line 92: Would it be useful to change the variable name for agencyAccountType, given that it is now a Pair?

mikehalmamoj (approved):

PR #833, opened 2021-02-04
NN-3102: cellsWithCapacity returns cells with zero operational capacity (WIP)

sp-watson:
This is the best solution I could find which was kind-of readable and didn't have an effect on the other endpoints.


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/AgencyInternalLocation.java
```diff
@@ -92,10 +92,8 @@ private boolean isActiveCell() {
         return isActive() && isCell();
     }
 
-    private boolean hasSpace() {
-        if (operationalCapacity != null && currentOccupancy != null ) {
-            return currentOccupancy < operationalCapacity;
-        }
+    private boolean hasSpace(boolean treatZeroOperationalCapacityAsNull) {
```
JamieDraperUK, line 95: Just a fyi, all parameters need to be final.

JamieDraperUK (dismissed):
LGTM!

JamieDraperUK (approved):
LGTM

PR #837, opened 2021-02-05
NN-3159 Return associated activity details in relation to transactions

JamieDraperUK:


sp-watson (commented):
All looks OK - just a general question though, have we got AC for this ticket?

sp-watson (approved):

PR #838, opened 2021-02-05
DT-1580: 🐛 Remove row mapper cache

petergphillips:


helm_deploy/values-preprod.yaml
```diff
@@ -10,6 +10,7 @@ image:
   port: 8080
 
 env:
+  JAVA_OPTS: "-Xmx1024m"
```
petergphillips, line 13: re-add memory settings

src/main/java/uk/gov/justice/hmpps/prison/repository/mapping/Row2BeanRowMapper.java
```diff
@@ -78,12 +48,8 @@ private void loadColumns(final ResultSet rs) {
 
 
     @SuppressWarnings({"unchecked", "rawtypes"})
-    public static <M> RowMapper<M> makeMapping(final String sql, final Class<M> type, final Map<String, FieldMapper> mappings) {
-        final var mappingInfo = new MappingInfo(sql, type);
```
petergphillips, line 82: the crazy thing here is that the `sql` was only ever being used as part of the key for the mappings, it isn't required to create the mappings.  Therefore we could have just left the cache and used the `type` and `mappings` fields, but since it isn't a great overhead (~1ms) to calculate each time then just removed altogether

mikehalmamoj (approved):

PR #850, opened 2021-02-12
Nn 3161 fix oracle only sql

sp-watson:
Fix the fix - the SQL assumed the initial location was a top-level location


JamieDraperUK (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/repository/sql/InmateRepositorySql.kt
```diff
@@ -258,9 +258,11 @@ enum class InmateRepositorySql(val sql: String) {
         LEFT JOIN OFFENDERS O ON B.OFFENDER_ID = O.OFFENDER_ID
                 WHERE B.ACTIVE_FLAG = 'Y' AND B.BOOKING_SEQ = 1
         AND B.LIVING_UNIT_ID IN (
-                SELECT INTERNAL_LOCATION_ID
-                        FROM AGENCY_INTERNAL_LOCATIONS START WITH INTERNAL_LOCATION_ID = :locationId
-                        CONNECT BY PRIOR INTERNAL_LOCATION_ID = PARENT_INTERNAL_LOCATION_ID
+            WITH LOCATION_SUBQUERY (INTERNAL_LOCATION_ID, PARENT_INTERNAL_LOCATION_ID) AS (
+                SELECT INTERNAL_LOCATION_ID, NULL FROM AGENCY_INTERNAL_LOCATIONS WHERE INTERNAL_LOCATION_ID = :locationId
```
Mjwillis, line 262: Best to check performance on this

sp-watson: We have checked the performance by using the app, using Swagger and executing SQL directly against the T3 DB. We have also made calls using cells and wings as input, and the performance is still fine.

Mjwillis: 👍

PR #852, opened 2021-02-17
CAT-677: Use second latest assessment for re-categorisations when lat…

connormaglynn:
…est is category U

Changes:
- Removed filtering of 'assessment status' - as we need to include inactive assessments due to the older any previous assessments usually being in this state.
- Removed Category U assessments before filtering down to the latest assessment. This is so we can return the second latest assessment for nominals who have a latest assessment of Category U. This is needed as the Category U is used as a holding category.


steverendell (approved):
lgtm

PR #858, opened 2021-02-23
DT-1628 Allow all movements to be returned even if released.

Mjwillis:


src/main/java/uk/gov/justice/hmpps/prison/service/MovementsService.java
```diff
@@ -299,8 +299,10 @@ private final String checkTransferParameters(final List<String> agencyIds, final
     }
 
     @VerifyBookingAccess
-    public Page<OffenderIn> getOffendersIn(final String agencyId, final LocalDateTime fromDate, final LocalDateTime toDate, final Pageable pageable) {
-        final var page = externalMovementRepository.findMovements(agencyId, ActiveFlag.Y, MovementDirection.IN, fromDate, toDate, pageable);
+    public Page<OffenderIn> getOffendersIn(final String agencyId, final LocalDateTime fromDate, final LocalDateTime toDate, final Pageable pageable, final boolean allMovements) {
```
JamieDraperUK, line 302: Is it worth adding a unit test for this logic? Feels like something we'd test in the MovementServiceTest

JamieDraperUK (approved):
LGTM

PR #860, opened 2021-02-24
NN-3178: Validate case note text size by number of bytes to ensure we…

sp-watson:
Ensure case notes that are over 4000 bytes (due to use of Unicode chars) are rejected, as they cause an insert error.

Also, on testing I noticed the error message shown by the UI includes the property names, so it isn't very user friendly. I fixed it, but then found lots of tests failed so reverted the commit.

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/api/model/NewCaseNote.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/api/model/NewCaseNote.java
@@ -43,6 +43,7 @@ public class NewCaseNote {
 
     private LocalDateTime occurrenceDateTime;
 
+    @Size(max = 4000) // For Swagger - custom annotations not well supported
     @MaximumTextSize
     @NotBlank
     private String text;
--- a/src/main/java/uk/gov/justice/hmpps/prison/api/model/NewCaseNote.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/api/model/NewCaseNote.java
@@ -11,6 +11,7 @@ import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
+import uk.gov.justice.hmpps.prison.service.validation.MaximumTextSize;
 
 import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.Size;
@@ -42,7 +43,7 @@ public class NewCaseNote {
 
     private LocalDateTime occurrenceDateTime;
 
-    @Size(max = 4000)
+    @MaximumTextSize
     @NotBlank
     private String text;
 
--- a/src/main/java/uk/gov/justice/hmpps/prison/service/CaseNoteService.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/service/CaseNoteService.java
@@ -68,7 +68,7 @@ public class CaseNoteService {
                            final AuthenticationFacade authenticationFacade,
                            final BookingService bookingService,
                            @Value("${batch.max.size:1000}") final int maxBatchSize,
-                           MaximumTextSizeValidator maximumTextSizeValidator) {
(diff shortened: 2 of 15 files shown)


src/main/java/uk/gov/justice/hmpps/prison/api/model/NewCaseNote.java
```diff
@@ -42,7 +43,7 @@
 
     private LocalDateTime occurrenceDateTime;
 
-    @Size(max = 4000)
+    @MaximumTextSize
```
mikehalmamoj, line 46: Seems like quite a generic name for a very specific rule - would it be better to give it a `max` parameter? It could then be dropped in as a replacement for `@Size` in other places if needs be...

src/test/java/uk/gov/justice/hmpps/prison/service/CaseNoteServiceImplIntTest.java
```diff
@@ -26,13 +28,36 @@
...
     private AuthenticationFacade authenticationFacade;
 
     @Autowired
     private CaseNoteService caseNoteService;
 
+    @Test
+    @WithMockUser(username = "ITAG_USER", roles = {"CASE_NOTE_EVENTS"})
+    public void createCaseNote_maximumTextSizeExceededDueToUtf8() {
+
+        String stringWith10Chars = "ABCDE12345";
```
JamieDraperUK, line 46: Use final var to declare types unless it's not possible or hurts readability.

src/main/java/uk/gov/justice/hmpps/prison/service/CaseNoteService.java
```diff
@@ -61,18 +61,21 @@
...
     public CaseNoteService(final CaseNoteRepository caseNoteRepository, final CaseNoteTransformer transformer,
-                               final UserService userService,
-                               final AuthenticationFacade authenticationFacade,
-                               final BookingService bookingService,
-                               @Value("${batch.max.size:1000}") final int maxBatchSize) {
+                           final UserService userService,
+                           final AuthenticationFacade authenticationFacade,
+                           final BookingService bookingService,
+                           @Value("${batch.max.size:1000}") final int maxBatchSize,
+                           MaximumTextSizeValidator maximumTextSizeValidator) {
```
JamieDraperUK, line 71: Meh, needs to be final

src/main/java/uk/gov/justice/hmpps/prison/api/model/NewCaseNote.java
JamieDraperUK: Not sure what swagger looks like for this but does it still document the maximum size of 4000?

src/test/java/uk/gov/justice/hmpps/prison/service/CaseNoteServiceImplIntTest.java
sp-watson: Ah yes

src/main/java/uk/gov/justice/hmpps/prison/service/CaseNoteService.java
sp-watson: grrr

src/main/java/uk/gov/justice/hmpps/prison/api/model/NewCaseNote.java
sp-watson: The purpose of the validation is to stop Oracle from throwing exceptions when the text size exceeds 4000 bytes, which is the max varchar size for Oracle. It is an absolute limit when saving data, hence the specific name.

mikehalmamoj: I thought 4000 chars was just the column size - didn't realise 4000 was a hard limit!  In which case it makes perfect sense to me.

JamieDraperUK (approved):

PR #874, opened 2021-03-11
Nn 3224 add csra class to offender details

sp-watson:
Add csra classification to offender detail endpoint, when requested


JamieDraperUK (approved):
LGTM

PR #877, opened 2021-03-12
GDPR-245: Introduce new provisional referral deletion

hyderkhanmoj:


src/main/java/uk/gov/justice/hmpps/nomis/datacompliance/service/DataComplianceReferralService.java
```diff
@@ -91,6 +107,29 @@ private OffenderPendingDeletion generateOffenderPendingDeletionEvent(final Offen
...
+
+        return transform(offenderNo, offenderAliases, referralId);
+    }
+
+    private ProvisionalDeletionReferralResult transform(final String offenderNumber, final List<OffenderAliasPendingDeletion> offenderAliases, Long referralId) {
+
+        return ProvisionalDeletionReferralResult.builder()
+            .referralId(referralId)
+            .offenderIdDisplay(offenderNumber)
+            .subsequentChangesIdentified(false)
```
brightonsbox, line 125: I guess this is a future piece of work to identify this?

brightonsbox (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/nomis/datacompliance/service/DataComplianceReferralService.java
hyderkhanmoj: if we find the offender still only has a single booking then we mark it as false. In the future when we are dealing with more bookings that will need to change

hyderkhanmoj: (Building the object with `subsequentChangesIdentified` is subject to finding a single booking for the same offender)

If we do not find a single booking we set that as true. 

In the dps service, we also check that the last known locations match

PR #879, opened 2021-03-15
NN-3233: Add date of applicable CSRA to offender details with csraSum…

sp-watson:
When getting offender details, and the csraSummary is requested, we need the assessment date as well as the overall classification


JamieDraperUK (approved):
LGTM

PR #880, opened 2021-03-19
NN-3254: support multiple answers to csra question

sp-watson:
In a CSRA assessment there are some questions where more than 1 answer can be selected - at the moment that causes a server error.

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/api/model/AssessmentQuestion.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/api/model/AssessmentQuestion.java
@@ -9,6 +9,7 @@ import lombok.Data;
 import lombok.NoArgsConstructor;
 
 import javax.validation.constraints.NotNull;
+import java.util.List;
 
 @SuppressWarnings("unused")
 @ApiModel(description = "AssessmentQuestion")
@@ -23,6 +24,9 @@ public class AssessmentQuestion {
     @ApiModelProperty(value = "Question", position = 1)
     private String question;
 
-    @ApiModelProperty(value = "Answer", position = 2)
+    @ApiModelProperty(value = "The answer given. More than one answer might have been given, in which case the other answers will be in the additionalAnswers property", position = 2)
     private String answer;
+
+    @ApiModelProperty(value = "If a question has more than one answer, all but the first answer will be in this property", position = 3)
+    private List<String> additionalAnswers;
 }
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/justice/hmpps/prison/service/OffenderAssessmentService.java
```diff
@@ -100,9 +102,20 @@ private AssessmentSummary getAssessmentSummary(final OffenderAssessment assessme
...
+            .map(OffenderAssessmentItem::getAssessmentAnswer)
+            .collect(groupingBy((aa) -> aa.getParentAssessment().getAssessmentId(), mapping(AssessmentEntry::getDescription, toList())));
+
+        return assessmentQuestions.stream().map(aq -> getAssessmentQuestionAndAnswers(aq, assessmentAnswersByQuestionId.get(aq.getAssessmentId()))).collect(toList());
+    }
 
-        return assessmentQuestions.stream().map(aq -> new AssessmentQuestion(aq.getDescription(), assessmentAnswersByQuestionId.get(aq.getAssessmentId()))).collect(Collectors.toList());
+    private AssessmentQuestion getAssessmentQuestionAndAnswers(final AssessmentEntry assessment, final List<String> answers) {
+        if (answers == null) {
+            return new AssessmentQuestion(assessment.getDescription(), null, null);
```
andrewrlee, line 114: is it worth having a test covering when there aren't any answers?

src/main/java/uk/gov/justice/hmpps/prison/api/model/AssessmentQuestion.java
```diff
@@ -23,6 +24,9 @@
     @ApiModelProperty(value = "Question", position = 1)
     private String question;
 
-    @ApiModelProperty(value = "Answer", position = 2)
+    @ApiModelProperty(value = "The answer given. More than one answer might have been given, in which case the other answers will be in the additionalAnswers property", position = 2)
     private String answer;
+
+    @ApiModelProperty(value = "If a question has more than one answer, all but the first answer will be in this property", position = 3)
+    private List<String> additionalAnswers;
```
andrewrlee, line 31: are there any integration test changes that should be made for this?

andrewrlee (approved):
Couple of Qs around testing, but otherwise LGTM

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderAssessmentService.java
sp-watson: There already is a test for that, as part of the main test. See https://github.com/ministryofjustice/prison-api/blob/main/src/test/java/uk/gov/justice/hmpps/prison/service/OffenderAssessmentServiceTest.java#L132

src/main/java/uk/gov/justice/hmpps/prison/api/model/AssessmentQuestion.java
sp-watson: We tend to use integration tests to cover the happy path(s) and deal with details (like this) in the service layer unit tests

PR #887, opened 2021-04-06
DT-1829: ⬆️ Use latest versions

petergphillips:


andymarke (dismissed):
LGTM

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderTransactionHistoryResourceTest.java
```diff
@@ -232,8 +232,8 @@ public void When_GetOffenderTransactionHistory_And_BadFromDateFormat_Then_ErrorR
             OFFENDER_NO);
 
         assertThat(response.getBody().getStatus().intValue()).isEqualTo(HTTP_BAD_REQ);
-        assertThat(response.getBody().getDeveloperMessage()).isEqualTo("Invalid value for MonthOfYear (valid values 1 - 12): 30");
```
petergphillips, line 235: had to change test here due to spring boot upgrade

build.gradle
```diff
@@ -45,7 +45,6 @@ dependencies {
     implementation 'javax.activation:activation:1.1.1'
 
     implementation 'commons-codec:commons-codec:1.15'
-    implementation 'com.fasterxml.jackson.module:jackson-modules-java8:2.12.1'
```
petergphillips, line 48: not needed since jackson uses databind anyway

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderTransactionHistoryResourceTest.java
petergphillips: note that the response body will change for clients, but very dubious that any will notice since still returns the bad request response as expected

andymarke: @petergphillips do we not get a body anymore or is there some other problem?

andymarke (approved):
LGTM

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderTransactionHistoryResourceTest.java
petergphillips: we do, it is just that the validation message is now different.

PR #891, opened 2021-04-14
NN-3308:

JamieDraperUK:
- Add new endpoint to return bed assignments for all living units


sp-watson (dismissed):
Could do with a test method in BedAssignmentHistoriesRepositoryTest to be consistent with the other methods in that repo class

sp-watson (approved):
LGTM

PR #900, opened 2021-04-23
DT-1897: 🐛 Fix info endpoint not working after java upgrade (second t…

petergphillips:
…ime lucky)


src/test/java/uk/gov/justice/hmpps/prison/api/resource/InfoIntTest.java
```diff
@@ -52,7 +52,7 @@ public void testReadInfoWithCachePopulated() {
             createHttpEntity(token, null),
             new ParameterizedTypeReference<String>() {
             },
-            "ADDRESS_TYPE");
+            "TASK_TYPE");
```
petergphillips, line 55: this task has subtypes, so therefore exposes the `util` bug

mikehalmamoj (approved):

PR #901, opened 2021-04-26
CAT-677: Add Cat U assessments for re-categorisations if offender has…

connormaglynn:
… had a previous standard assessment

The goal is ton include category U assessments into the re-categorisation list since category U assessments are often used as a holding category for an offender.

The logic:
- Filter out any offenders that don't have at least one standard categorisation (B,C,D)
- Filter out all but the latest assessments for each offender
- Filter out any assessments that aren't B,C,D or U
- Return remaining assessments


src/main/java/uk/gov/justice/hmpps/prison/repository/InmateRepository.java
```diff
@@ -571,13 +572,11 @@ private void calcAdditionalInformation(final OffenderBooking booking) {
         final var rawData = jdbcTemplate.query(
                 InmateRepositorySql.GET_RECATEGORISE.getSql(),
                 createParams("agencyId", agencyId,
-                        "assessStatus", Set.of("A", "P"),
```
connormaglynn, line 574: If I'm taking out this filter in the SQL, do we think we should add in the filter on the final step removeNonStandardCategoryRecords() to only return assessments with a status of 'A' or 'P' 🤔

src/main/java/uk/gov/justice/hmpps/prison/repository/InmateRepository.java
```diff
@@ -610,6 +609,14 @@ private Long getCategoryAssessmentTypeId() {
                 .collect(Collectors.toList());
     }
 
+    private List<OffenderCategorise> removeOffendersWithNoPreviousStandardCategorisations(final List<OffenderCategorise> catList) {
+        final var standardCategoryCodes = Set.of("B", "C", "D");
+        final var offenderNoMap = catList.stream().collect(Collectors.groupingBy(OffenderCategorise::getOffenderNo));
+        return offenderNoMap.values().stream().filter(oc ->
```
connormaglynn, line 615: thinking about it again this morning, should this be offenderNo or bookingId?

src/main/java/uk/gov/justice/hmpps/prison/repository/InmateRepository.java
steverendell: yes, we should return the 'current' cat, not an inactive one

steverendell: Also does assessStatus need removing from the SQL

src/main/java/uk/gov/justice/hmpps/prison/repository/InmateRepository.java
steverendell: We should remove the offender if no std cat in the current bookingId. But as it happens getOffenderNo is synonymous here as we should only get the current booking from the db for each offender (as they can only be 'in prison' for one booking)

steverendell: Another thought is that we should probably do all this in one strean pipeline !

src/main/java/uk/gov/justice/hmpps/prison/repository/InmateRepository.java
connormaglynn: It's what I did on the last PR, if memory serves right I think it was because anything but the recent one was marked as 'I' (Inactive), so wouldn't be able to look at the previous assessments for an offender if we left it in (might be wrong though!)

Suppose the question is do we want to return the latest assessment if it's Inactive?

src/main/java/uk/gov/justice/hmpps/prison/repository/InmateRepository.java
connormaglynn: Brill cheers thought I'd double check, yeah thought a bit a refactoring might be needed! 😁

steverendell: 👍

src/main/java/uk/gov/justice/hmpps/prison/repository/InmateRepository.java
steverendell: Yes what i meant was that its fine to remove assessStatus from the SQL, and we should return the 'current' cat (i.e. the A or P), not an inactive one, _from the api_.

steverendell (approved):

PR #904, opened 2021-04-29
P4-2813 adding (optional) non-breaking movement type filter param to movements out for given date endpoint so can filter by type where needed.

matt-crowson-moj:
Changes:

Updating existing endpoint (with non-braking change) to add an optional request query parameter 'movementType' to filter the movements 'out' for a given date.

This enables PECS (Book a Secure Move) to work out the number of releases for a given agency on a specific date.  Other existing endpoints were considered but did not really make sense in terms of easily changing.  Adding a new endpoint seemed like overkill.

### DIFF
--- a/src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/MovementResourceTest.java
+++ b/src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/MovementResourceTest.java
@@ -223,4 +223,47 @@ public class MovementResourceTest extends ResourceTest {
         assertThatStatus(response, HttpStatus.OK.value());
         assertThat(getBodyAsJsonContent(response)).isStrictlyEqualToJson("movements_upcoming_court.json");
     }
+
+    @Test
+    public void testGetAllMovementsOutForAGivenDate() {
+        final var token = authTokenHelper.getToken(AuthTokenHelper.AuthToken.NORMAL_USER);
+
+        final var response = testRestTemplate.exchange(
+            "/api/movements/{agencyId}/out/{isoDate}",
+            HttpMethod.GET,
+            createHttpEntity(token, null),
+            new ParameterizedTypeReference<String>() {
+            }, "LEI", LocalDate.of(2017, 7, 16)
+        );
+
+        assertThatStatus(response, HttpStatus.OK.value());
+        assertThat(getBodyAsJsonContent(response)).isStrictlyEqualToJson("movements_out_on_given_day.json");
+    }
+
+    @Test
+    public void testGetAllMovementsOutForAGivenDateAndMovementType() {
+        final var token = authTokenHelper.getToken(AuthTokenHelper.AuthToken.NORMAL_USER);
+
+        final var movementsFoundResponse = testRestTemplate.exchange(
+            "/api/movements/{agencyId}/out/{isoDate}?movementType={movementType}",
+            HttpMethod.GET,
+            createHttpEntity(token, null),
+            new ParameterizedTypeReference<String>() {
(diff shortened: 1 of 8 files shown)


src/test/java/uk/gov/justice/hmpps/prison/service/MovementsServiceImplTest.java
```diff
@@ -329,9 +358,9 @@ public void testMovementsForAgenciesNoAgencyCodes() {
         final var transferEvents = true;
         final var movements = true;
 
-        assertThatThrownBy(() -> {
-            final var transferSummary = movementsService.getTransferMovementsForAgencies(agencyList, from, to, courtEvents, releaseEvents, transferEvents, movements);
```
matt-crowson-moj, line 333: Non-functional change.  Just removing unused var and warning.

src/test/resources/uk/gov/justice/hmpps/prison/api/resource/impl/movements_out_on_given_day.json
```diff
@@ -0,0 +1,10 @@
+[
+  {
```
JamieDraperUK, line 2: Feels like we should be returning movementType for when the filter is left blank.

src/test/resources/uk/gov/justice/hmpps/prison/api/resource/impl/movements_out_on_given_day.json
```diff
@@ -0,0 +1,10 @@
...
+    "offenderNo": "Z0020ZZ",
+    "dateOfBirth": "1966-01-01",
+    "reasonDescription": "Funerals And Deaths",
+    "timeOut": "00:00:00",
+    "firstName": "Burt",
+    "lastName": "Reynolds"
+  }
+]
```
Nimphal, line 10: Would it be worth adding some data for the filtering? I saw that this file was used for both resource tests, which I assume is because the test database only has one move for the date. I know it can be a pain to add more data, so definitely not blocking.

Nimphal (approved):
Nice work! I really like the additional tests you've added around the repository, resource and service.

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/MovementResourceTest.java
```diff
@@ -223,4 +223,47 @@ public void testGetUpcomingCourtAppearances() {
...
+        final var response = testRestTemplate.exchange(
+            "/api/movements/{agencyId}/out/{isoDate}",
+            HttpMethod.GET,
+            createHttpEntity(token, null),
+            new ParameterizedTypeReference<String>() {
+            }, "LEI", LocalDate.of(2017, 7, 16)
+        );
+
+        assertThatStatus(response, HttpStatus.OK.value());
+        assertThat(getBodyAsJsonContent(response)).isStrictlyEqualToJson("movements_out_on_given_day.json");
```
JamieDraperUK, line 240: Is it worth picking a date with more data?

src/test/resources/uk/gov/justice/hmpps/prison/api/resource/impl/movements_out_on_given_day.json
matt-crowson-moj: I nearly got away with it too!

src/test/resources/uk/gov/justice/hmpps/prison/api/resource/impl/movements_out_on_given_day.json
matt-crowson-moj: Yeah I did already think about this..

PR #908, opened 2021-05-05
NN-3363: New CSRA ratings endpoint for multiple offenders

sp-watson:


src/main/java/uk/gov/justice/hmpps/prison/service/OffenderAssessmentService.java
```diff
@@ -118,6 +120,22 @@ private AssessmentQuestion getAssessmentQuestionAndAnswers(final AssessmentEntry
             answers.stream().skip(1).collect(toList()));
     }
 
+    public List<AssessmentClassification> getOffendersAssessmentRatings(List<String> offenderList) {
```
JamieDraperUK, line 123: You mentioned above about dealing with large offender numbers, should we add an additional finder method that will return them all in one go? We'd have to batch them of course so comes with some complexity.

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderAssessmentService.java
```diff
@@ -118,6 +120,22 @@ private AssessmentQuestion getAssessmentQuestionAndAnswers(final AssessmentEntry
...
+        return offenderList.stream().map(this::getOffenderCurrentAssessmentRating).collect(Collectors.toList());
+    }
+
+    private AssessmentClassification getOffenderCurrentAssessmentRating(String offenderNo) {
```
JamieDraperUK, line 127: Sorry, missing final :cry:

src/test/resources/uk/gov/justice/hmpps/prison/api/resource/impl/csra_ratings.json
```diff
@@ -0,0 +1,10 @@
+[
+  {
+    "offenderNo": "A1183JE",
+    "classificationCode": "HI",
+    "classificationDate": "2019-01-05"
+  },
+  {
+    "offenderNo": "A1234BB"
```
JamieDraperUK, line 8: Is this right? Feels strange to return data without any csra rating data.

src/test/java/uk/gov/justice/hmpps/prison/service/OffenderAssessmentServiceTest.java
```diff
@@ -402,6 +402,50 @@ public void getCurrentCsraClassification_returnsNullIfNoAssessments() {
...
+        when(repository.findByCsraAssessmentAndByOffenderNo_OrderByLatestFirst("N1234AA")).thenReturn(List.of(
+            getOffenderAssessment_CsraClassificationBuilder(new AssessmentClassification("HI", "High"), LocalDate.parse("2019-01-02"))
+            .build()
+        ));
+        when(repository.findByCsraAssessmentAndByOffenderNo_OrderByLatestFirst("N2345BB")).thenReturn(List.of());
+
+        final var csraRatings = service.getOffendersAssessmentRatings(List.of("N1234AA", "N2345BB"));
+
+        assertThat(csraRatings).isEqualTo(List.of(
+            uk.gov.justice.hmpps.prison.api.model.AssessmentClassification.builder()
```
JamieDraperUK, line 437: Adding a comment here in case you missed the absolute type reference.

JamieDraperUK (approved):
LGTM

src/test/java/uk/gov/justice/hmpps/prison/service/OffenderAssessmentServiceTest.java
sp-watson: I didn't - there is a name clash, but I couldn't think of a better name so i just used the full type ref

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderAssessmentService.java
sp-watson: arrggh!

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderAssessmentService.java
sp-watson: Yes, I did consider it but I thought it wasn't a big enough gain for the work required. We are close to the DB here so making many calls to the DB doesn't have a large overhead compared to 1 call. And the memory consumption will be the same (if not better).

It's a good point though - now I am wondering if there might be a performance problem for lot of them...  I need to test it

src/test/resources/uk/gov/justice/hmpps/prison/api/resource/impl/csra_ratings.json
sp-watson: i just couldn't think of a good reason to add extra code to remove them. If you think it is more consistent then I will change it.

JamieDraperUK: It probably makes sense to filter them out, If only to reduce the amount of noise that a request for a large data set could cause.

PR #911, opened 2021-05-06
NN-3363: Improvements to new CSRA ratings endpoint

sp-watson:
Batch DB calls for CSRA ratings for multiple offenders. Also only return data for offenders that have assessments

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/service/OffenderAssessmentService.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/service/OffenderAssessmentService.java
@@ -1,8 +1,10 @@
 package uk.gov.justice.hmpps.prison.service;
 
+import com.google.common.collect.Lists;
 import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.validation.annotation.Validated;
@@ -19,9 +21,9 @@ import uk.gov.justice.hmpps.prison.security.VerifyBookingAccess;
 import uk.gov.justice.hmpps.prison.security.VerifyOffenderAccess;
 
 import java.time.LocalDate;
+import java.util.ArrayList;
 import java.util.List;
-import java.util.Map;
-import java.util.stream.Collectors;
+import java.util.Optional;
 
 import static java.util.stream.Collectors.*;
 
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/justice/hmpps/prison/service/OffenderAssessmentService.java
```diff
@@ -120,20 +124,39 @@ private AssessmentQuestion getAssessmentQuestionAndAnswers(final AssessmentEntry
             answers.stream().skip(1).collect(toList()));
     }
 
-    public List<AssessmentClassification> getOffendersAssessmentRatings(List<String> offenderList) {
-        return offenderList.stream().map(this::getOffenderCurrentAssessmentRating).collect(Collectors.toList());
+    @Transactional(readOnly = true)
+    public List<AssessmentClassification> getOffendersAssessmentRatings(final List<String> offenderNos) {
+        final var classifications = new ArrayList<AssessmentClassification>();
+        final var batch = Lists.partition(new ArrayList<>(offenderNos), maxBatchSize);
+        batch.forEach(offenderNosBatch -> classifications.addAll(getOffendersCurrentAssessmentRating(offenderNosBatch)));
```
JamieDraperUK, line 131: Meh, could be a flatMap to keep up with consistency

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderAssessmentService.java
```diff
@@ -120,20 +124,39 @@ private AssessmentQuestion getAssessmentQuestionAndAnswers(final AssessmentEntry
...
+            .map(e -> getOffenderCurrentAssessmentRating(e.getKey(), e.getValue()))
+            .filter(Optional::isPresent)
+            .map(Optional::get)
+            .collect(toList());
     }
 
-    private AssessmentClassification getOffenderCurrentAssessmentRating(String offenderNo) {
-        var currentClassification = getCurrentCsraClassification(offenderNo);
+    private Optional<AssessmentClassification> getOffenderCurrentAssessmentRating(final String offenderNo, final List<OffenderAssessment> assessmentsLatestFirst) {
+        var currentClassification = calculateCurrentCsraClassification(assessmentsLatestFirst);
```
JamieDraperUK, line 146: Prob needs to be final (sorry :runner: )

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderAssessmentService.java
```diff
@@ -120,20 +124,39 @@ private AssessmentQuestion getAssessmentQuestionAndAnswers(final AssessmentEntry
...
-        return AssessmentClassification.builder()
+        return Optional.of(AssessmentClassification.builder()
             .offenderNo(offenderNo)
             .classificationCode(currentClassification.classificationCode)
             .classificationDate(currentClassification.classificationDate)
-            .build();
+            .build());
+    }
+
+    private CurrentCsraAssessment calculateCurrentCsraClassification(final List<OffenderAssessment> offenderAssessmentsLatestFirst) {
```
JamieDraperUK, line 157: Would be be better to return a Optional here?

JamieDraperUK (dismissed):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderAssessmentService.java
sp-watson: Yes, that was the comment you had before and i fixed it... I'll do it again

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderAssessmentService.java
sp-watson: I spent about 5 minutes thinking about this and decided to leave as is. The other method that does use Optional was done so we could better chain calls in a stream.

JamieDraperUK (approved):
LGTM

PR #929, opened 2021-05-18
NN-3395: Return information about created appointments

sp-watson:


src/main/java/uk/gov/justice/hmpps/prison/service/AppointmentsService.java
```diff
@@ -31,19 +33,20 @@
...
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
+import static java.util.stream.Collectors.toList;
 import static uk.gov.justice.hmpps.prison.security.AuthenticationFacade.hasRoles;
 
 @Service
 @Validated
 @Transactional(readOnly = true)
+@Slf4j
```
sp-watson, line 49: This is temporary - it is here so we can see how long it takes

src/main/java/uk/gov/justice/hmpps/prison/service/AppointmentsService.java
```diff
@@ -100,8 +97,17 @@ public void createAppointments(@NotNull @Valid final AppointmentsToCreate appoin
 
         final var withRepeats = withRepeats(appointments.getRepeat(), flattenedDetails);
```
sp-watson, line 98: I will refactor this if we go with this solution.. but right now we just need to test it.

JamieDraperUK (approved):
Approving so we can test the performance in t3 and preprod

PR #934, opened 2021-05-20
DT-2061: 🐛 Another attempt at setting the sampling correctly

petergphillips:


applicationinsights.json
```diff
@@ -32,11 +32,6 @@
               "key": "http.url",
               "value": "https?://[^/]+/health.*",
               "matchType": "regexp"
-            },
-            {
```
petergphillips, line 36: not sure why this isn't working, so removing for now

applicationinsights.json
```diff
@@ -32,11 +32,6 @@
...
-              "key": "http.status_code",
-              "value": "200",
```
petergphillips, line 38: tried both a string here and a number, neither seem to work.
According to https://docs.microsoft.com/en-us/azure/azure-monitor/app/java-standalone-sampling-overrides it should be able to be set like this, but just causes it to be ignored

SimonMitchellMOJ (approved):
LGTM

PR #937, opened 2021-05-25
NN-3407:

JamieDraperUK:
- Remove the concept of a main appointment


sp-watson (approved):
My only concern with this is performance - returning 5 fields for each appointment will end up been quite a chunk of return data, when we only need the id's. But it should be OK with the 300 app'tments we saw in App Insights

PR #940, opened 2021-05-28
Nn 3422

JamieDraperUK:
- Create a new endpoint to delete multiple appointments in one go.
- Tidy up appointment service test

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/service/AppointmentsService.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/service/AppointmentsService.java
@@ -98,16 +98,16 @@ public class AppointmentsService {
         assertThatAppointmentsFallWithin(appointmentsWithRepeats, appointmentTimeLimit());
 
         final var createdAppointments = appointmentsWithRepeats.stream().map(a -> {
-            final var appointmentId = bookingRepository.createAppointment(a, defaults, agencyId);
+                final var appointmentId = bookingRepository.createAppointment(a, defaults, agencyId);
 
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/justice/hmpps/prison/service/AppointmentsService.java
```diff
@@ -137,6 +137,18 @@ public ScheduledEvent getBookingAppointment(Long appointmentId) {
         return getScheduledEventOrThrowEntityNotFound(appointmentId);
     }
 
+    @Transactional
+    @PreAuthorize("hasAnyRole('GLOBAL_APPOINTMENT')")
+    public void deleteBookingAppointments(List<Long> appointmentIds) {
+        appointmentIds.forEach(appointmentId -> bookingRepository
```
sp-watson, line 143: This is quite inefficient - I think it can all be done in bulk. But you would need to detect which id's are present first so the the telemetry calls are accurate, which could be done in a single query rather than getting the whole object for each appt.

I don't think it makes a massive difference though - it's up to you if you want to change it.

sp-watson (approved):

src/main/java/uk/gov/justice/hmpps/prison/service/AppointmentsService.java
JamieDraperUK: I did consider just deleting them in a batch but realised we still need to fetch each one to gather info for the tracking event.

sp-watson (approved):

PR #947, opened 2021-06-11
NN-3438:

JamieDraperUK:
- Add unit tests


sp-watson (approved):

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderBooking.java
```diff
@@ -304,22 +304,28 @@ public RestrictivePatient getRestrictivePatientDetails() {
         return getLastMovement().map(lastMovement -> mapRestrictivePatient(lastMovement, getLegalStatus(), null)).orElse(null);
     }
 
-    public static RestrictivePatient mapRestrictivePatient(final ExternalMovement lastMovement, LegalStatus legalStatus, LocalDate releaseDate) {
-        if (REL.getCode().equals(lastMovement.getMovementType().getCode()) &&
-            DISCHARGE_TO_PSY_HOSPITAL.getCode().equals(lastMovement.getMovementReason().getCode())) {
+    public static RestrictivePatient mapRestrictivePatient(final ExternalMovement lastMovement, final LegalStatus legalStatus, final LocalDate releaseDate) {
+        if (!isReleasedAndDischargedToHospital(lastMovement)) return null;
```
JamieDraperUK, line 308: I think this logic should be applied when someone tries to discharge an offender. Possibly coming in future PRs. Or better yet this information lives in another service with it's own data store

sp-watson: Yes re. a different data store, but the problem is making sure that new service is always used (by client services) instead of calling prison api directly.

sp-watson (approved):

PR #948, opened 2021-06-14
NN-3453: Prevent SQL error when getting residential locations and no top-level locations found

sp-watson:
…etting residential locations


JamieDraperUK (approved):
LGTM

PR #950, opened 2021-06-21
DaCS-1049 Add status filter to visits with visitors endpoint

andrewrlee:
Associated hub ticket:  https://trello.com/c/YHRAOUQB

Adding the ability to filter by status of the visit so we can exclude cancelled visits.


src/main/java/uk/gov/justice/hmpps/prison/api/model/VisitWithVisitors.java
```diff
@@ -19,7 +19,7 @@
 @Builder
 @AllArgsConstructor
 @NoArgsConstructor
-public class VisitWithVisitors<V extends VisitDetails> {
+public class VisitWithVisitors {
```
andrewrlee, line 22: This didn't need to be generic so removed

mikehalmamoj (approved):

PR #953, opened 2021-06-23
DT-2244 Add establishment name to the /offenders/{nomsId}/full-status endpoint

Mjwillis:


src/test/resources/uk/gov/justice/hmpps/prison/api/resource/impl/prisoners_information_Z0023ZZ.json
```diff
@@ -1,5 +1,6 @@
 {
   "establishmentCode": "OUT",
+  "establishmentName": "Outside",
```
petergphillips, line 3: don't we normally try to add in the ` - release from xxx` too here?

petergphillips (approved):
LGTM

PR #962, opened 2021-06-30
NN-3162 Change /{agencyId}/eventLocationsBooked endpoint to return locations regardless of whether all the offenders associated have been suspended

yasinmustafa:


yasinmustafa:
more work needed

PR #963, opened 2021-06-30
NN-3162 Change /{agencyId}/eventLocationsBooked endpoint to return locations regardless of whether all the offenders associated have been suspended

yasinmustafa:
Change for prison-staff-hub, checked via app insights that it was safe to change it directly as only PSH uses this endpoint
https://mojdt.slack.com/archives/CK0QLCQ1G/p1624970146367300


src/test/java/uk/gov/justice/hmpps/prison/repository/AgencyRepositoryTest.java
```diff
@@ -201,6 +201,12 @@ public void testFilterPrisonIepReviewByIepLevel() {
         assertThat(results.getItems()).containsExactly(OFFENDER_6_IEP_REVIEW);
     }
 
+    @Test
+    public void testGetAgencyLocationsWhenAllOffendersSuspended() {
+        final var locations = repository.getAgencyLocationsBooked("BXI", LocalDate.of(2021, Month.MAY, 1), null);
+        assertThat(locations).hasSize(2);
```
mikehalmamoj, line 207: Might be worth asserting that you are getting the records back that you think you are?  Probably ok now but if more test data goes in it might start passing by accident?

mikehalmamoj (approved):

src/test/java/uk/gov/justice/hmpps/prison/repository/AgencyRepositoryTest.java
yasinmustafa: Thanks @mikehalmamoj - good point,  will make the change

mikehalmamoj (approved):

PR #967, opened 2021-07-07
P4-2976 trialing a change to see if it resolves an issue in production

matt-crowson-moj:
**Changes:**

BaSM is seeing a `NullPointer` exception when trying to schedule court hearings for bookings with a court case id.

`"java.lang.NullPointerException: Cannot invoke \"uk.gov.justice.hmpps.prison.repository.jpa.model.OffenderCourtCase.getId()\" because \"cc\" is null\n\tat uk.gov.justice.hmpps.prison.repository.jpa.model.OffenderBooking.lambda..."`

I have not actually been able to recreate this locally but would like to try this to see if it resolves it.

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderBooking.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderBooking.java
@@ -35,6 +35,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


Nimphal (approved):

JamieDraperUK (approved):

JamieDraperUK:
Might be a good idea to add an integration test as the issue is most likely due to the way the object is materialised by JPA

matt-crowson-moj:
> Might be a good idea to add an integration test as the issue is most likely due to the way the object is materialised by JPA

@JamieDraperUK there is already integration test coverage.  Agree though it is probably a JPA issue....

PR #971, opened 2021-07-15
NN-3450: Get discharge-to-hospital endpoint to work on the first call

sp-watson:
This was caused by a DB trigger on the EXTERNAL_MOVEMENTS table (trigger id = `OFFENDER_EXTERNAL_MOVEMENTS_T1 `), which updates the OFFENDER_BOOKINGS STATUS_REASON column.

When using JPA, the execution order of the SQL statements is independent of when the changes are made. In our instance, the SQL to deactivate the EXTERNAL_MOVEMENTS was being run after the SQL to update the OFFENDER_BOOKINGS table - this caused the STATUS_REASON to be overwritten by the afore-mentioned trigger.

The real bug IMO is that this trigger runs even when an EXTERNAL_MOVEMENTS entry has ACTIVE = "N", but we cannot wait for that to be changed (if indeed Syscon are willing to update it at all).


Mjwillis (approved):
LGTM

PR #973, opened 2021-07-15
Revert "Revert "DT-2321 Remove deprecated Case Note finder, handle access to data consistently""

Mjwillis:
Reverts ministryofjustice/prison-api#972


petergphillips (approved):
LGTM

PR #979, opened 2021-07-21
Nn 3554

JamieDraperUK:
- new endpoint to create an external movement


src/main/java/uk/gov/justice/hmpps/prison/PrisonApiServer.java
```diff
@@ -3,11 +3,10 @@
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.boot.autoconfigure.domain.EntityScan;
-import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
 import org.springframework.scheduling.annotation.EnableScheduling;
 
+
 @SpringBootApplication(scanBasePackages = { "uk.gov.justice.hmpps" })
-@EnableJpaRepositories(basePackages = { "uk.gov.justice.hmpps" })
```
JamieDraperUK, line 10: This has been pulled out into a separate @Configuration class to avoid JPA being loaded with the WebMvcTest tests

src/main/java/uk/gov/justice/hmpps/prison/service/MovementsService.java
```diff
@@ -314,6 +331,75 @@ private final String checkTransferParameters(final List<String> agencyIds, final
...
+            .orElseThrow(EntityNotFoundException.withMessage("movementReason not found using: %s", createExternalMovement.getMovementReason()));
+
+        final var movementReasons =
+            movementTypeAndReasonRespository.findMovementTypeAndReasonByTypeIs(createExternalMovement.getMovementType());
+
+        if (movementReasons.stream().noneMatch(r -> r.getReasonCode().equals(createExternalMovement.getMovementReason())))
+            throw new EntityNotFoundException("Invalid movement reason for supplied movement type");
+
+        if (createExternalMovement.getMovementType().equals(MovementType.REL.getCode()) && offenderBooking.isActive())
+            throw new IllegalStateException("Can not create an external movement of type REL if the offender is active");
```
JamieDraperUK, line 353: I suspect we'd want to add more constrains here in the future.

mikehalmamoj (dismissed):
LGTM but I think it would be good for @Mjwillis to also have a look when he is back in the office as he's been thinking a lot about this kind of stuff recently.

mikehalmamoj (approved):

src/main/java/uk/gov/justice/hmpps/prison/service/MovementsService.java
```diff
@@ -314,6 +331,74 @@ private final String checkTransferParameters(final List<String> agencyIds, final
...
+            .movementTime(createExternalMovement.getMovementTime())
+            .fromAgency(fromAgency)
+            .toAgency(toAgency)
+            .movementDirection(createExternalMovement.getDirectionCode())
+            .movementType(movementType)
+            .movementReason(movementReason)
+            .activeFlag(ActiveFlag.Y)
+            .build();
+
+        offenderBooking.addExternalMovement(externalMovement);
```
sp-watson, line 374: You need to check whether all previous movements need to be deactivated as well. If so, you also need to consider that the order of execution matters - see PrisonerReleaseAndTransferService.deactivatePreviousMovements

JamieDraperUK: I don't fully understand why we need to but I'm going to have a catch up with @Mjwillis once he's back to make sure he's happy with the approach overall.

PR #980, opened 2021-07-26
DT-2365 Validate new booking's initial location capacity

mikehalmamoj:
Currently if the location is full then an SQL exception is returned to the user - made the error a bit more friendly.


petergphillips (approved):
LGTM

PR #981, opened 2021-07-26
DT-2365 Improve error for recall/transfer-in where cell is full

mikehalmamoj:


JamieDraperUK (approved):
LGTM, prob could of got away with just a unit test in the service for this.

mikehalmamoj:
> LGTM, prob could of got away with just a unit test in the service for this.

I would have liked to but the service has 32 dependencies injected into it and so the unit test would have been more complicated than the integration test!  (I'm hoping that service this will be broken up when it is used in anger on Receive a Prisoner).

PR #991, opened 2021-08-04
DCS-1134 Respecting the location parameter during prioritised search

andrewrlee:
Currently if a prioritised search falls through to personalised details if no results found for offender number or PNC.
In this scenario, the location parameter is ignored - this adds it back in


andymarke (approved):
LGTM

PR #994, opened 2021-08-09
NN-3589 fix connection proxy issues

JamieDraperUK:
There seems to be a problem attaining an Oracle proxy connection. The username doesn't seem to exists and we think this is because the verify access checks are being triggered on the read only replication database where this user doesn't exists. Moving these checks to the service should ensure that the correct db is checked as the whole service is within a write transaction.

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
@@ -112,7 +112,6 @@ public class OffenderResource {
         @ApiResponse(code = 500, message = "Unrecoverable error occurred whilst processing request.", response = ErrorResponse.class)})
     @ApiOperation("Full details about the current state of an offender")
     @GetMapping("/{offenderNo}")
-    @VerifyOffenderAccess(overrideRoles = {"SYSTEM_USER", "GLOBAL_SEARCH", "VIEW_PRISONER_DATA"})
     public InmateDetail getOffender(@Pattern(regexp = "^[A-Z]\\d{4}[A-Z]{2}$", message = "Offender Number format incorrect") @PathVariable("offenderNo") @ApiParam(value = "The offenderNo of offender", example = "A1234AA", required = true) final String offenderNo) {
         return inmateService.findOffender(offenderNo, true);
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
```diff
@@ -123,7 +122,7 @@ public InmateDetail getOffender(@Pattern(regexp = "^[A-Z]\\d{4}[A-Z]{2}$", messa
         @ApiResponse(code = 500, message = "Unrecoverable error occurred whilst processing request.", response = ErrorResponse.class)})
     @ApiOperation("Full details about the current state of an offender")
     @GetMapping("/{offenderNo}/prison-timeline")
-    @VerifyOffenderAccess(overrideRoles = {"SYSTEM_USER", "VIEW_PRISONER_DATA"})
+    @PreAuthorize("hasAnyRole('SYSTEM_USER', 'VIEW_PRISONER_DATA')")
```
petergphillips, line 125: will need to ponder this - these annotations are doing completely different things:
`VerifyOffenderAccess` allows all users to call the endpoint with overrides for those two roles if the user doesn't have privileges
`PreAuthorize` requires that the users must have those two roles in order to access the endpoint.

Need to decide on an endpoint by endpoint basis which is correct I fear

JamieDraperUK: I've just moved to @VerifyOffenderAccess to the service layer and pulled @PreAuthorize up to the controller level for consistency

src/main/java/uk/gov/justice/hmpps/prison/service/MovementsService.java
```diff
@@ -115,7 +116,7 @@ public MovementsService(final MovementsRepository movementsRepository,
                 .build());
     }
 
-    @PreAuthorize("hasAnyRole('SYSTEM_USER', 'VIEW_PRISONER_DATA')")
```
JamieDraperUK, line 118: Moved to the controller

petergphillips (approved):
LGTM

PR #1000, opened 2021-08-10
DT-2371: 🐛 Fix permissions for offender sentences

petergphillips:


Mjwillis (approved):
Thanks for all the tests!

PR #1001, opened 2021-08-11
DT-2440 Allow multiple roles to be specified when filtering

Mjwillis:


src/test/java/uk/gov/justice/hmpps/prison/repository/UserRepositoryTest.java
```diff
@@ -118,6 +118,15 @@ public void testFindUsersWithCaseloadSearch() {
         assertThat(items).extracting(UserDetail::getUsername).contains("JBRIEN");
     }
 
+    @Test
+    public void testFindUsersMulitpleRoles() {
```
petergphillips, line 122: would be good to test the multiple roles code with a user that has multiple roles, difficult to tell if it works in this test since no data is returned

Mjwillis: I know! I was being lazy with setting up more data!  I did test on T3 locally at it works but will look at that now

petergphillips (approved):
LGTM

PR #1004, opened 2021-08-12
Add Metadata config for automated service catalogue

swilson-mojhub:
**Context** 

We are building an **automated single source of truth service catalogue** proof of concept using [backsatge.io](https://hmpps-backstage-dev.hmpps.service.justice.gov.uk). 

**Intent**

This is a small PR to add the metadata for the prison API so that backstage.io can pick up the data and add the Prisons API into the service catalogue


petergphillips (approved):
LGTM

PR #1007, opened 2021-08-13
DT-2329 Additional alert filter that are used by booking version of API

andymarke:


src/test/java/uk/gov/justice/hmpps/prison/repository/OffenderAlertRepositoryTest.java
```diff
@@ -185,8 +214,178 @@ void canFilterByCodeEvenWhenReferenceLinkIsMissing() {
...
+    void canFilterByAlertTypes() {
+        final var filter = OffenderAlertFilter
+            .builder()
+            .offenderNo("A1179MT")
+            .alertTypes("V")
+            .build();
+        final var alerts = repository.findAll(filter);
+
+        assertThat(alerts)
+            .hasSize(2)
```
mikehalmamoj, line 227: I'm not sure that you should worry about the size of the results for these - makes the tests a bit flaky.  The `containsOnly` should cover the test?

mikehalmamoj (approved):

src/test/java/uk/gov/justice/hmpps/prison/repository/OffenderAlertRepositoryTest.java
andymarke: I did think that, but what if it only returned 1 result because of a faulty filter when it should have returned 8?

I think a better solution would be a dynamic data loaded that clears up after the test finishes so we are not reliant on seed that. What do you think @mikehalmamoj 

I will not do that in this epic but I think t needs to be done as we done more of this kind of stuff

PR #1013, opened 2021-08-18
DT-2458 Log when alert service is processing legacy "query", 

andymarke:
so we know when it can be removed

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/repository/InmateAlertRepository.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/repository/InmateAlertRepository.java
@@ -1,6 +1,7 @@
 package uk.gov.justice.hmpps.prison.repository;
 
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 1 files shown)


lbennett-moj (approved):

src/main/java/uk/gov/justice/hmpps/prison/repository/InmateAlertRepository.java
```diff
@@ -234,4 +241,12 @@ public long createNewAlert(final long bookingId, final CreateAlert alert) {
 
         return alertSeq;
     }
+
+
+    private static void checkQueryIsNotUsed(String query) {
+        Optional.ofNullable(query)
+            .filter(not(String::isBlank))
+            .ifPresent(q -> log.warn("Alert repository has received a query parameter that might be vulnerable"));
```
Mjwillis, line 249: Wonder is its worth thinking of putting an aspect around all endpoints with `String query` @andymarke ?

PR #1015, opened 2021-08-19
ESWE-67: New endpoint (and resource) to expose work data

sp-watson:
This is the first endpoint to provide data for the Education screens in DPS.

There will be at least 1 more endpoint coming soon - to get work history, and the screens might change - hence we have hidden the API docs so we don't have to worry about breaking changes.


src/main/java/uk/gov/justice/hmpps/prison/service/OffenderActivitiesService.java
```diff
@@ -0,0 +1,53 @@
...
+        final var allAllocatedWorkActivities = offenderProgramProfileRepository.findByOffenderBooking_BookingIdAndProgramStatus(currentBookingId, "ALLOC");
+        final var workActivities =  allAllocatedWorkActivities.stream().filter(this::isValidCurrentActivity).map(this::transform).collect(Collectors.toList());
+        return OffenderActivities.builder()
+            .offenderNo(offenderNo)
+            .bookingId(currentBookingId)
+            .workActivities(workActivities)
+            .build();
+    }
+
+    private boolean isValidCurrentActivity(OffenderProgramProfile offenderProgramProfile) {
```
sp-watson, line 35: I have put these checks here rather than in the JPA query for several reasons:

 - To prevent the repo from having lots of JPQL
 - I think we will need to reuse this logic for other queries
 - Easier to test (no SQL to set up the tables)

It might be that we have to move these checks to the SQL for performance, but for the moment they are better here.

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderProgramProfile.java
```diff
@@ -0,0 +1,40 @@
...
+    @Column(name = "OFF_PRGREF_ID")
+    private Long offenderProgramReferenceId;
+
+    @ManyToOne(optional = false, fetch = FetchType.LAZY)
+    @JoinColumn(name = "OFFENDER_BOOK_ID", nullable = false)
+    private OffenderBooking offenderBooking;
+
+    @ManyToOne(fetch = FetchType.LAZY)
+    // TODO - @NotFound(action = IGNORE)
+    @JoinColumn(name = "CRS_ACTY_ID")
```
andymarke, line 28: maybe check if there is a foreign key in NOMIS and assuming not is there any bad data that would cause this to fail?

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderActivitiesService.java
andymarke: Maybe consider moving the method on to OffenderProgramProfile, it *might* be easier to reuse. Seems to be a pattern we are using more, get the JPA model to understand itself

andymarke (dismissed):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderActivitySummary.java
```diff
@@ -0,0 +1,29 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.NotNull;
+import java.time.LocalDate;
+
+@ApiModel(description = "Prisoner Movement")
```
Mjwillis, line 15: Naming wrong?

sp-watson: Yes thanks. I'll fix it

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderProgramProfile.java
sp-watson: Ah yes. I am fixing a naming issue, so i'll sort this out

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderActivitiesService.java
```diff
@@ -0,0 +1,40 @@
...
+    private final OffenderProgramProfileRepository offenderProgramProfileRepository;
+    private final BookingService bookingService;
+
+    public OffenderActivitiesService(final OffenderProgramProfileRepository offenderProgramProfileRepository, final BookingService bookingService) {
+        this.offenderProgramProfileRepository = offenderProgramProfileRepository;
+        this.bookingService = bookingService;
+    }
+
+    public OffenderActivities getCurrentWorkActivities(String offenderNo) {
+        final var currentBookingId = bookingService.getLatestBookingByOffenderNo(offenderNo).getBookingId();
```
Mjwillis, line 24: Should use `OffenderRespository` not the old repo, then you have the JPA object available. @sp-watson

PR #1019, opened 2021-08-23
DT-2332 Log query parameter prior to removal

andymarke:
Pretty certain this is always blank, so logging to double check prior to its removal


petergphillips (approved):
LGTM

PR #1028, opened 2021-09-01
DT-2335 Map IEP into JPA

Mjwillis:


src/main/java/uk/gov/justice/hmpps/prison/api/resource/BookingResource.java
```diff
@@ -200,10 +200,13 @@ private static void checkQueryIsNotUsed(String query) {
         @RequestParam(value = "prisonId", required = false) @ApiParam(value = "Filter by prison Id", example = "MDI") final String prisonId,
         @RequestParam(value = "bookingId", required = false) @ApiParam("Filter by a list of booking ids") final List<Long> bookingIds,
         @RequestParam(value = "offenderNo", required = false) @ApiParam("Filter by a list of offender numbers") final List<String> offenderNos,
+        @RequestParam(value = "iepLevel", defaultValue = "false", required = false) @ApiParam(value = "Return IEP level data", defaultValue = "false") final boolean iepLevel,
```
andymarke, line 203: @Mjwillis Really interested to know what other people feel about these extra flags, I can see why we do it, but feels a bit messy, but maybe it is better than having 4 versions of the same API.

I guess we do it for performance reasons, but still feels a bit messy.

What I don't like is the filter parameters are mixed in with the data flags.

@mikehalmamoj  Do you have any ideas on a better way to do this? Almost like a GraphQL type API where you can control what you want returning

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/CourtEvent.java
```diff
@@ -77,15 +77,15 @@
     @Column(name = "START_TIME", nullable = false)
     private LocalDateTime startTime;
 
-    @ManyToOne
+    @ManyToOne(fetch = FetchType.EAGER)
     @NotFound(action = IGNORE)
```
andymarke, line 81: I thought having the NotFound meant it would be EAGER loaded anyway?

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/CourtEvent.java
Mjwillis: > Yes so being explicit as it generates a warning

src/main/java/uk/gov/justice/hmpps/prison/api/resource/BookingResource.java
Mjwillis: Agree be good to look at a better approach

PR #1045, opened 2021-09-17
NN-3603 remove case note creation as one is already created by a trigger

JamieDraperUK:


petergphillips (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerReleaseAndTransferService.java
```diff
@@ -137,9 +137,6 @@ public InmateDetail releasePrisoner(final String prisonerIdentifier, final Reque
 
         createOutMovement(booking, REL, movementReason, supportingPrison, toLocation, releaseDateTime, requestToReleasePrisoner.getCommentText(), null );
 
-        // generate the release case note
```
Mjwillis, line 140: Shame about this - because we now rely on the trigger to create a case note so its a hidden effect

PR #1046, opened 2021-09-17
Revert "NN-3603 remove case note creation as one is already created by a trigger"

JamieDraperUK:
Reverts ministryofjustice/prison-api#1045


Mjwillis (approved):
LGTM

PR #1047, opened 2021-09-20
DT-2586 Current image of prisoner not displayed if not most recent

Mjwillis:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderImage.java
```diff
@@ -62,8 +60,7 @@
     private Long sequence;
 
     @Column(name = "ACTIVE_FLAG")
-    @Enumerated(EnumType.STRING)
-    private ActiveFlag activeFlag;
+    private String activeFlag;
```
andymarke, line 63: could this just be mapped as a boolean so you don't have to do the `= 'Y'` below?

andymarke (approved):
LGTM

PR #1056, opened 2021-10-01
NN-3435: release date does not always match

richardpopple:


src/main/java/uk/gov/justice/hmpps/prison/repository/sql/BookingRepositorySql.kt
```diff
@@ -89,6 +89,8 @@ enum class BookingRepositorySql(val sql: String) {
         O.LAST_NAME,
         O.BIRTH_DATE                                   DATE_OF_BIRTH,
         OB.agy_loc_id                                  agency_location_id,
+        CASE WHEN(OB.ACTIVE_FLAG = 'Y' AND OB.BOOKING_SEQ = 1)
```
sp-watson, line 92: Are both these checks required (`ACTIVE_FLAG` and `BOOKING_SEQ`)?

The variable name suggests that the last booking seq might not be active?!

sp-watson (approved):

PR #1058, opened 2021-10-05
ESWE-164 endpoint to get unacab count data for a given offender over a given amount of time

steverendell:


src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderAttendance.java
```diff
@@ -0,0 +1,42 @@
...
+public class OffenderAttendance {
+    @NotNull
+    private Long bookingId;
+
+    @NotNull
+    @ApiModelProperty(value = "The date of this activity")
+    private LocalDate eventDate;
+
+    @ApiModelProperty(value = "Whether the offender attended")
+    private String outcome;
```
andymarke, line 32: maybe give examples ? like "YES", "ATTENDED" or whatever outcome means, in fact can it be a boolean?

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderAttendance.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @ApiModelProperty(value = "The course code")
+    private String code;
```
andymarke, line 35: Feels like this should be inside a Course object, else it read as OffenderAttendance.code whereas OffenderAttendance.course.code is better

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderActivitiesService.java
```diff
@@ -1,38 +1,44 @@
...
         this.offenderProgramProfileRepository = offenderProgramProfileRepository;
-        this.bookingService = bookingService;
+        this.attendanceRepository = attendanceRepository;
     }
 
-    public Page<OffenderActivitySummary> getStartedActivities(final String offenderNo, final LocalDate earliestEndDate, final PageRequest pageRequest) {
-        final var startedActivities = offenderProgramProfileRepository.findByNomisIdAndProgramStatusAndEndDateAfter(offenderNo, List.of("ALLOC", "END"), earliestEndDate,
-            pageRequest);
-        return startedActivities.map(this::transform);
+    @PreAuthorize("hasRole('SYSTEM_USER')")
```
andymarke, line 32: should this be on the resource to better document the role required at resource level

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderActivitiesResourceTest.java
```diff
@@ -68,5 +69,76 @@ public void successfulRequest_page() {
...
+            final var entity = createHttpEntity(validToken(List.of("ROLE_SYSTEM_USER")), null);
+            final var today = LocalDate.now().format(DateTimeFormatter.ISO_LOCAL_DATE);
+
+            final var response = testRestTemplate.exchange(
+                "/api/offender-activities/A1234AB/attendance-history?fromDate=2017-01-01&toDate=" + today + "&page=1&size=2&sort=eventId,desc",
+                HttpMethod.GET,
+                entity,
+                String.class);
+
+            assertThatJsonFileAndStatus(response, HttpStatus.OK.value(), "offender-attendence-history.json");
```
andymarke, line 100: I think these tests would read better if you insert the data in the test setup then use JSON asserts inline to assert on the data returned, else you have to scan the seed that and that file to understand what the API is doing

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderAttendance.java
steverendell: Will need to revisit anyway and though youre probably right, it is not yet fixed which fields will actually be needed

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderActivitiesService.java
steverendell: There seems to be 2 conflicting conventions - annotation the services or the resources 🤷 . Ive added the role to the swagger description

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderActivitiesResourceTest.java
steverendell: Agree - the seed data has become very 'entangled' with multiple tests. Will address next time.

PR #1078, opened 2021-11-03
NN-3742: Create adjudications using offenderNo not booking id

sp-watson:
Also return offenderNo when retrieving


JamieDraperUK (approved):
LGTM

PR #1081, opened 2021-11-08
NN-3745:

JamieDraperUK:
- get multiple adjudications

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/service/AdjudicationsService.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/service/AdjudicationsService.java
@@ -2,7 +2,6 @@ package uk.gov.justice.hmpps.prison.service;
 
 import com.google.common.collect.Lists;
 import com.microsoft.applicationinsights.TelemetryClient;
-import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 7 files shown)


sp-watson (approved):

src/main/java/uk/gov/justice/hmpps/prison/service/AdjudicationsService.java
```diff
@@ -53,6 +51,31 @@
     @Value("${batch.max.size:1000}")
```
sp-watson, line 51: Guess we don't need this now - we/i can remove later

src/test/java/uk/gov/justice/hmpps/prison/service/AdjudicationsServiceTest.java
```diff
@@ -57,7 +57,7 @@
     private static final String EXAMPLE_AGENCY_ID = "LEI";
     private static final String EXAMPLE_STATEMENT = "Example statement";
     private static final LocalDateTime EXAMPLE_INCIDENT_TIME = LocalDateTime.of(2020, 1, 1, 2, 3, 4);
-    private final int BATCH_SIZE = 1;
+    private Integer BATCH_SIZE = 1;
```
sp-watson, line 60: Think this is a left-over - we can sort later

sp-watson (approved):

PR #1084, opened 2021-11-15
NN-3723 your completed reports

richardpopple:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/AdjudicationRepository.java
```diff
@@ -15,4 +17,7 @@
 
     Optional<Adjudication> findByParties_AdjudicationNumber(final Long adjudicationNumber);
     List<Adjudication> findByParties_AdjudicationNumberIn(final List<Long> adjudicationNumbers);
+
+    @Query("SELECT DISTINCT a FROM Adjudication a JOIN a.parties p WHERE (:agencyLocationId IS null Or a.agencyLocation.id = :agencyLocationId) AND (p.adjudicationNumber IN :adjudicationNumbers)")
```
richardpopple, line 21: This is done so we can allow passing in of a null agencyId and get results back for all agencies - which doesn't work with using standard spring data naming conventions.

src/main/java/uk/gov/justice/hmpps/prison/api/resource/AdjudicationsResource.java
```diff
@@ -72,4 +80,20 @@ public AdjudicationDetail getAdjudication(
...
+    @ApiOperation(value = "Search of adjudications", notes = "Requires MAINTAIN_ADJUDICATIONS access")
+    @ApiImplicitParams({
+        @ApiImplicitParam(name = "page", dataType = "java.lang.Integer", paramType = "query",
+            value = "Results page you want to retrieve (0..N). Default 0, e.g. the first page", example = "0"),
+        @ApiImplicitParam(name = "size", dataType = "java.lang.Integer", paramType = "query",
+            value = "Number of records per page. Default 10"),
+        @ApiImplicitParam(name = "sort", dataType = "java.lang.String", paramType = "query",
+            value = "Sort as combined comma separated property and uppercase direction. Multiple sort params allowed to sort by multiple properties. Default to incidentDate,incidentTime ASC")})
+    @PostMapping("/search")
+    @PreAuthorize("hasRole('MAINTAIN_ADJUDICATIONS')")
```
richardpopple, line 93: Although we are using this to get a users reported adjudications of a particular agency, I suspect we will need to do something like this for other calls, and so I thought a generic name like search was best here

sp-watson (approved):

PR #1086, opened 2021-11-17
DT-2556: include additional fields on the interface

neilmendum:
https://dsdmoj.atlassian.net/browse/DT-2556

See [Slack message](https://mojdt.slack.com/archives/C02MV6MP933/p1637146006008500)

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderKeyDates.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderKeyDates.java
@@ -26,4 +26,10 @@ public class OffenderKeyDates {
 
     @ApiModelProperty(value = "SED - date on which sentence expires.", example = "2020-02-03")
     private LocalDate sentenceExpiryDate;
+
+    @ApiModelProperty(required = true, value = "Effective sentence end date.", example = "2020-02-03")
+    private LocalDate effectiveSentenceEndDate;
+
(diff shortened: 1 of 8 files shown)


Mjwillis (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderKeyDates.java
```diff
@@ -26,4 +26,10 @@
 
     @ApiModelProperty(value = "SED - date on which sentence expires.", example = "2020-02-03")
     private LocalDate sentenceExpiryDate;
+
+    @ApiModelProperty(required = true, value = "Effective sentence end date.", example = "2020-02-03")
+    private LocalDate effectiveSentenceEndDate;
+
+    @ApiModelProperty(required = true, value = "Sentence length.", example = "11/00/00")
```
yasinmustafa, line 33: is this example of sentenceLength right? e.g. "11/00/00".. what does that mean?

yasinmustafa: Note to self - we need to update CRD to include these required fields otherwise the POST will fail once this is merged. Our service is in development yet so not a big deal

neilmendum: Updated the SwaggerDoc so hopefully the format is clearer

Mjwillis (approved):
LGTM

PR #1088, opened 2021-11-18
DT-2556: log event to AppInsights; remove Hidden Swagger annotation

neilmendum:
https://dsdmoj.atlassian.net/browse/DT-2556

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/service/OffenderDatesService.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/service/OffenderDatesService.java
@@ -75,7 +75,12 @@ public class OffenderDatesService {
                 .build();
         offenderBooking.addSentenceCalculation(sentenceCalculation);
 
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/justice/hmpps/prison/service/OffenderDatesService.java
```diff
@@ -71,6 +74,9 @@ public SentenceCalcDates updateOffenderKeyDates(Long bookingId, RequestToUpdateO
                 .judiciallyImposedSentenceLength(keyDatesFromPayload.getSentenceLength())
                 .build();
         offenderBooking.addSentenceCalculation(sentenceCalculation);
+
+        telemetryClient.trackEvent("OffenderKeyDatesUpdated", ImmutableMap.of("bookingId", bookingId.toString(), "calculationUuid", requestToUpdateOffenderDates.getCalculationUuid().toString()), null);
```
Mjwillis, line 78: Log the user

Mjwillis (changes requested):
I would log the user too

Mjwillis (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderDatesService.java
neilmendum: Good point, updated

PR #1093, opened 2021-11-19
CRS-572 adding line sequence and case seqeuence to sentence api.

ldlharper:
Adding case sequence and line sequence to prison api so we can match the sentence display in NOMIS.


joelstobart-moj (approved):
LGTM

PR #1094, opened 2021-11-21
CVSL-241: Add a POST endpoint to get HDC status for a list of booking IDs

timharrison-moj:
The HDC status is only currently available for a single booking Id in a GET endpoint.
For Create and Vary a Licence, it is useful to retrieve this information for several booking Ids (the cases which are allocated to one probation officer, generally 10-25).  This PR add a POST endpoint, service method and repository + SQL and tests for this purpose.

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderSentenceResource.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderSentenceResource.java
@@ -90,6 +90,17 @@ public class OffenderSentenceResource {
     }
 
     @ApiResponses({
+        @ApiResponse(code = 200, message = "List of HDC status information", response = HomeDetentionCurfew.class, responseContainer = "List"),
+        @ApiResponse(code = 400, message = "Invalid request.", response = ErrorResponse.class),
+    })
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderSentenceResource.java
```diff
@@ -89,6 +89,17 @@ public HomeDetentionCurfew getLatestHomeDetentionCurfew(@PathVariable("bookingId
         return offenderCurfewService.getLatestHomeDetentionCurfew(bookingId);
     }
 
+    @ApiResponses({
+        @ApiResponse(code = 200, message = "List of HDC status information", response = HomeDetentionCurfew.class, responseContainer = "List"),
```
Mjwillis, line 93: Don't think you need the 200

Mjwillis (approved):
Small
Comment

src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderSentenceResource.java
timharrison-moj: I think you're right, but the other endpoints in this resource all define their 200 responses. Leave in or take out? I'm easy.

PR #1098, opened 2021-11-25
NN-3789: PUT endpoint for adjudications

sp-watson:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/Adjudication.java
```diff
@@ -34,7 +34,7 @@
 
 @Data
 @EqualsAndHashCode(callSuper=false)
-@Builder
+@Builder(toBuilder=true)
```
sp-watson, line 37: Needed to make update tests easier

natclamp-moj (approved):

PR #1104, opened 2021-12-03
Pat 2109 add more iep attributes to agency iep review

garethmdavies:
As a offender manager I want to be able to see the following values when assessing an individual for an IEP review:

- the number of days that the individual has been on their current level
- the total number of positive case notes
- the total number of negative case notes

Currently to find the above values, multiple requests need to be made to the prison api case note endpoints for each sub type of positive and negative case note (8 api calls) as well as a separate call to the prisoner iep summary record to calculate the days on level.

When displaying the IEP review by agency location data additional calls per prisoner at the location to retrieve these values above, can result in a large number of requests being fired at the API.

This PR modifies the Agency IEP review SQL and API endpoint response to pull the required data back in a single api call. For one use case in the manage my incentives service this would reduce the number of api calls required to build the page from 500+ to 1 for a wing of one prison.


connormaglynn (approved):
lgtm

PR #1105, opened 2021-12-03
DCS-1344 :bug: Moving an offender out of a cell should decrement, not increment, occupancy.

bell-pa:


src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerReleaseAndTransferService.java
```diff
@@ -674,18 +674,18 @@ private void generateAdmissionNote(final OffenderBooking booking, final AgencyLo
...
-            assignedLivingUnit.incrementCurrentOccupancy();
+            assignedLivingUnit.decrementCurrentOccupancy();
             agencyInternalLocationRepository.save(assignedLivingUnit);
-            incrementCurrentOccupancy(assignedLivingUnit.getParentLocation());
+            decrementCurrentOccupancy(assignedLivingUnit.getParentLocation());
         }
     }
 
 
     private void updateBeds(final OffenderBooking booking, final LocalDateTime releaseDateTime) {
```
sp-watson, line 686: Not something you changed, so you don't have to change.... but the function name is too ambiguous - `deallocateBed` might be a better name

sp-watson (approved):

PR #1107, opened 2021-12-06
DCS-1330: Create endpoint to process court returns in prison API

bell-pa:
Handles the case where an offender returns to the prison they left.  
Arriving at a different prison is currently an error. 
That behaviour will be implemented for another story.


src/test/resources/sql/scheduledPrisonerReturnFromCourt_clean.sql
```diff
@@ -0,0 +1,7 @@
+DELETE from COURT_EVENTS where EVENT_ID in (455654698,455654697);
```
bell-pa, line 1: Use the offender_book_id of the offender_bookings row you created for these deletes.

src/test/resources/sql/scheduledPrisonerReturnFromCourt_init.sql
```diff
@@ -0,0 +1,139 @@
+INSERT INTO OFFENDERS (OFFENDER_ID, OFFENDER_NAME_SEQ, ID_SOURCE_CODE, LAST_NAME, NAME_TYPE, FIRST_NAME,
```
bell-pa, line 1: You don't need to populate most of the columns in these tables.  Certainly not the audit columns and very few of the other columns.  Adding these values makes it harder to see what's important for the test(s).

src/test/resources/sql/scheduledPrisonerReturnFromCourt_init.sql
```diff
@@ -0,0 +1,139 @@
...
+VALUES (2554472, null, 'SEQ', 'AIMONIS', null, 'EF''LIAICO', 'URUA', TO_DATE('1987-08-07', 'YYYY-MM-DD'), 'M',
+        null, 'A552', 'SHEFFIELD', 'ENG', TO_DATE('2017-02-17', 'YYYY-MM-DD'), 'AIMONIS', null, 'EFLIAICO',
+        'URUA', 'G6942UN', 2554472, 'INST', 'OMS_OWNER',
+        TO_DATE('2017-06-26', 'YYYY-MM-DD'), 'CN', null, null, null, null, 'W1',
+        null, null, null, null, null, 'MR', null, 'QQM25S', 'A',
+        TO_DATE('2017-02-17', 'YYYY-MM-DD'), '1234',
+        TO_DATE('2017-06-26', 'YYYY-MM-DD'), 'OMS_OWNER', 'MERGE', 'pgrant',
+        '10.102.1.69', 'unknown', null);
+
+INSERT INTO AGENCY_LOCATIONS (AGY_LOC_ID, DESCRIPTION, AGENCY_LOCATION_TYPE, ACTIVE_FLAG)
```
bell-pa, line 20: I think you should use one of the AGENCY_LOCATIONS that have been created by the flyway scripts.

src/test/resources/sql/scheduledPrisonerReturnFromCourt_init.sql
```diff
@@ -0,0 +1,139 @@
...
+                                                 INTERNAL_LOCATION_TYPE, DESCRIPTION, SECURITY_LEVEL_CODE, CAPACITY,
+                                                 CREATE_USER_ID, PARENT_INTERNAL_LOCATION_ID, ACTIVE_FLAG, LIST_SEQ,
+                                                 CREATE_DATETIME, MODIFY_DATETIME, MODIFY_USER_ID, CNA_NO,
+                                                 CERTIFIED_FLAG, DEACTIVATE_DATE, REACTIVATE_DATE,
+                                                 DEACTIVATE_REASON_CODE, COMMENT_TEXT, USER_DESC, ACA_CAP_RATING,
+                                                 UNIT_TYPE, OPERATION_CAPACITY, NO_OF_OCCUPANT, TRACKING_FLAG,
+                                                 AUDIT_TIMESTAMP, AUDIT_USER_ID, AUDIT_MODULE_NAME,
+                                                 AUDIT_CLIENT_USER_ID, AUDIT_CLIENT_IP_ADDRESS,
+                                                 AUDIT_CLIENT_WORKSTATION_NAME, AUDIT_ADDITIONAL_INFO)
+VALUES (106221, 'F', 'NMI', 'WING', 'NMI-F', null, 116, 'OMS_OWNER', null, 'Y', 10,
```
bell-pa, line 35: Likewise - try to use one of the pre-existing agency_internal_locations that are added by the flyway scripts.

src/test/resources/sql/scheduledPrisonerReturnFromCourt_init.sql
```diff
@@ -0,0 +1,139 @@
...
+        TO_DATE('2017-05-09', 'YYYY-MM-DD'), 'OMS_OWNER',
+        'sqlplus@DAPDL0001 (TNS V1-V3)', 'oracle', null, 'pts/6', null),
+       (106258, '009', 'NMI', 'CELL', 'NMI-F-2-009', null, 2, 'OMS_OWNER', 106249, 'Y', 9,
+        TO_DATE('2009-12-10', 'YYYY-MM-DD'),
+        TO_DATE('2017-05-09', 'YYYY-MM-DD'), 'OMS_OWNER', 1, 'Y', null, null,
+        null, null, null, null, 'NA', null, 2, 'N',
+        TO_DATE('2017-05-09', 'YYYY-MM-DD'), 'OMS_OWNER',
+        'sqlplus@DAPDL0001 (TNS V1-V3)', 'oracle', null, 'pts/6', null);
+
+INSERT INTO MOVEMENT_REASONS (MOVEMENT_TYPE, MOVEMENT_REASON_CODE, DESCRIPTION, OPEN_CONTACT_FLAG,
```
bell-pa, line 54: If this is a valid MOVEMENT_REASON then I think it should be put in a new flyway script and inserted once as static data.

src/test/resources/sql/scheduledPrisonerReturnFromCourt_clean.sql
michalnawrockidj: done

src/test/resources/sql/scheduledPrisonerReturnFromCourt_init.sql
michalnawrockidj: done

src/test/resources/sql/scheduledPrisonerReturnFromCourt_init.sql
michalnawrockidj: done

src/test/resources/sql/scheduledPrisonerReturnFromCourt_init.sql
michalnawrockidj: done

src/test/resources/sql/scheduledPrisonerReturnFromCourt_init.sql
michalnawrockidj: done

src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
```diff
@@ -242,6 +243,21 @@ public InmateDetail transferInPrisoner(
         return prisonerReleaseAndTransferService.transferInPrisoner(offenderNo, requestToTransferIn);
     }
 
+    @ApiResponses({
+        @ApiResponse(code = 400, message = "Invalid request.", response = ErrorResponse.class),
+        @ApiResponse(code = 403, message = "Forbidden - user not authorised to transfer a prisoner", response = ErrorResponse.class),
+        @ApiResponse(code = 404, message = "Requested resource not found.", response = ErrorResponse.class),
+        @ApiResponse(code = 500, message = "Unrecoverable error occurred whilst processing request.", response = ErrorResponse.class)})
+    @ApiOperation("Transfer a prisoner into a prison from court. Must be an out prisoner in currently in transfer status, requires the TRANSFER_PRISONER role")
```
andrewrlee, line 251: does this also require a token with a username? if so we could add this to the doc. 

The existing calls don't mention that this is required but should, they also fail in strange ways when not present. 
Will this fail in strange ways if not provided? - do we have a test for that case?

michalnawrockidj: @andrewrlee this endpoint is based on "/{offenderNo}/transfer-in".The test is here OffendersResourceCourtTransfersTest but nothing added for 403 scenario

andrewrlee (approved):
LGTM

PR #1122, opened 2021-12-15
GDPR-334: Introduce offender with images endpoint for compliance aws rekognition collection

hyderkhanmoj:


Mjwillis (dismissed):
I think this need to be moved into main package now if its permanent - there is already an Image JPA object and Repo - suggest you use that.

hyderkhanmoj:
In an ideal scenario the endpoint name would be something like:
`/offenders?hasImages={true}&fromDateTime={someDate}&page={true}`
However, as that is not in keeping with what is already there and there already endpoints that follow the pattern:
`images/offenders` I have followed the convention that exists.

Mjwillis (approved):
LGTM

PR #1125, opened 2021-12-17
NN-3868 code to find the users with roles for a caseload - reinstated

richardpopple:


Mjwillis:
Should use the endpoints in Nomis user roles api - there is one to find users with roles

PR #1136, opened 2022-01-10
CRS-665 fixing bug with indicator query.

ldlharper:
The way I'd added offence indicators to the CRD query meant that a new row was being returned for every offence indicator. This results in duplicate offences coming out the the API. Its been removed from the query graph now and is fetched lazily in the mapping.


src/main/resources/db/migration/data/R__9_12__OFFENCE_INDICATORS.sql
```diff
@@ -1 +1,2 @@
 INSERT INTO OFFENCE_INDICATORS (OFFENCE_INDICATOR_ID, OFFENCE_CODE, STATUTE_CODE, INDICATOR_CODE) VALUES (-1, 'RC86355', 'RC86', 'S15/CJIB');
+INSERT INTO OFFENCE_INDICATORS (OFFENCE_INDICATOR_ID, OFFENCE_CODE, STATUTE_CODE, INDICATOR_CODE) VALUES (-2, 'RC86355', 'RC86', 'SOR');
```
ldlharper, line 2: Adding another indicator to the test data here made the test fail with the old query approach.

joelstobart-moj (approved):
LGTM

yasinmustafa (approved):
LGTM

PR #1144, opened 2022-01-19
DCS-1413 Handling when no next visit is present

andrewrlee:


petergphillips (dismissed):
Approved, but think that DPS will need changes to cope with this change otherwise the 404 will cause quick look page not to appear

petergphillips (approved):
LGTM

PR #1146, opened 2022-01-20
NN-3890 add offence code to adjudications v2

richardpopple:


richardpopple:
On going - so it can be previewed by dps core team

PR #1148, opened 2022-01-20
INC-35 Multiple Proven Adjudication Endpoint

Mjwillis:


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/AdjudicationsResourceTest.java
```diff
@@ -329,4 +327,53 @@ public void returns403IfInvalidRole() {
             assertThatStatus(response, 403);
         }
     }
+
+    @Nested
+    public class GetProvenAdjudications {
```
sp-watson, line 332: A slight issue here: This file covers int tests for the AdjudicationsResource class. This test ought to be in whatever file covers `BookingResource` integration tests.

Mjwillis: I'll move it

PR #1149, opened 2022-01-20
CVSL-339: Wider roles allowed for HDC get status endpoints

timharrison-moj:
Was formerly just allowing ROLE_SYSTEM_USER.
Now additionally allows ROLE_VIEW_PRISONER_DATA.
This allows the CVL system client to avoid needing ROLE_SYSTEM_USER
HDC still OK as it currently uses ROLE_SYSTEM_USER.


Mjwillis (approved):
LGTM

PR #1151, opened 2022-01-20
SDI-4: 🐛 Fix contact missing exception

petergphillips:


andymarke (approved):
LGTM

src/test/resources/uk/gov/justice/hmpps/prison/api/resource/impl/visits_with_visitors_paged.json
```diff
@@ -3,76 +3,136 @@
...
           "personId": -3,
           "lastName": "JOHNSON",
           "firstName": "JOHN",
-          "leadVisitor": true,
-          "relationship": "Brother"
+          "dateOfBirth": "1956-02-28",
+          "leadVisitor": true
         }
       ],
       "visitDetails": {
```
andymarke, line 22: This is why I hate the prison-api usage of assertJSONFile.  Maybe you have broken the system since it is now a different order. Who knows??  The test tells me nothing about why really is expected just what we got

petergphillips: agreed.  I started writing some of the new tests just asserting on json path but that is a bit naff to since a non matching path just fails without a proper error message telling me what is wrong

PR #1153, opened 2022-01-21
SDI-4: 🎨 Use Stream.toList() instead

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/service/AgencyService.java
```diff
@@ -266,8 +266,7 @@ public void verifyAgencyAccess(final String agencyId) {
 
         final var locations = agencyRepository.getAgencyLocationsBooked(agencyId, bookedOnDay, bookedOnPeriod);
         final var processedLocations = LocationProcessor.processLocations(locations, true);
-        processedLocations.sort(LOCATION_DESCRIPTION_COMPARATOR);
-        return processedLocations;
+        return processedLocations.stream().sorted(LOCATION_DESCRIPTION_COMPARATOR).toList();
```
petergphillips, line 269: sorting on an immutable list now

andrewrlee (approved):
LGTM

PR #1157, opened 2022-01-25
INC-120 Provide single location lookup by description (unique)

Mjwillis:


src/main/java/uk/gov/justice/hmpps/prison/service/LocationService.java
```diff
@@ -100,6 +100,24 @@ public Location getLocation(final long locationId, boolean includeInactive) {
             .orElseThrow(EntityNotFoundException.withId(locationId));
     }
 
+    public Optional<Location> getLocationByDescription(final String description) {
+        return agencyInternalLocationRepository.findOneByDescription(description)
+            .map(
+                l -> LocationProcessor.processLocation(Location.builder()
```
steverendell, line 106: Not too sure about the parameter name !

Mjwillis: I know, but that is what it is called in the DB and the repo

Mjwillis: @steverendell what do you think ?  code, id, locationCode, Basically it is the description field that happens to be unique

steverendell: Anything like code or location is fine, its just that an 'l' looks like a '1' .. and is hard to spot in the block too 🙂

steverendell (approved):
lgtm

PR #1166, opened 2022-02-15
SDI-60: 🐛 Fix swagger issues with deployment

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/web/config/SwaggerConfig.kt
```diff
@@ -28,8 +25,6 @@ import java.time.ZonedDateTime
 import java.util.Optional
 
 @Configuration
-@EnableOpenApi
-@Import(BeanValidatorPluginsConfiguration::class)
```
petergphillips, line 32: not needed as in `SpringFoxConfiguration`

andymarke (approved):
LGTM

PR #1174, opened 2022-02-23
SDI-76 New endpoint to retrieve list of domains

andymarke:
Endpoint is likely to only be used by developers who want to guess at the domain they need for other APIs

- Use JPA Repository for domain rather then SQL version since this is standard going forward
- Use JsonPath for testing rather than json file assert that is confusing

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/ReferenceDomain.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/ReferenceDomain.java
@@ -1,0 +1,31 @@
+package uk.gov.justice.hmpps.prison.repository.jpa.model;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@AllArgsConstructor
+@NoArgsConstructor
+@Getter
+@Entity
+@Table(name = "reference_domains")
+public class ReferenceDomain {
+    @Id
+    @Column
+    private String domain;
+    @Column
+    private String description;
+    @Column
+    private String domainStatus;
+    @Column
+    private String ownerCode;
+    @Column
+    private String applnCode;
+    @Column
+    private String parentDomain;
+}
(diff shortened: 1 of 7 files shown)


lbennett-moj (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/ReferenceDomain.java
```diff
@@ -0,0 +1,31 @@
...
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@AllArgsConstructor
+@NoArgsConstructor
+@Getter
+@Entity
+@Table(name = "reference_domains")
+public class ReferenceDomain {
```
petergphillips, line 17: new classes should be kotlin :grin:

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/ReferenceDataResourceTest.java
```diff
@@ -181,4 +185,74 @@ public void testReadDomainReverseLookupWithWildcard() {
...
+
+        @Test
+        @DisplayName("will return a list of domains")
+        void willReturnAListOfDomains() {
+            final var token = authTokenHelper.getToken(AuthTokenHelper.AuthToken.NORMAL_USER);
+
+            final var response = getBodyAsJsonContent(getDomains(token));
+
+            assertThat(response).hasJsonPathArrayValue("$");
+            assertThat(response).extractingJsonPathNumberValue("$.length()").isEqualTo(430);
```
petergphillips, line 213: dubious about this one - looks like the test will be fragile as any new domain will cause it to fail.  Could you do a `> 300` and `< 600` instead?

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/ReferenceDataResourceTest.java
```diff
@@ -181,4 +185,74 @@ public void testReadDomainReverseLookupWithWildcard() {
...
+
+            assertThat(response)
+                .hasJsonPathValue(domainAt, "ETHNICITY");
+            assertThat(response)
+                .hasJsonPathValue(domainAt, "SKL_SUB_TYPE");
+
+            assertThat(response)
+                .extractingJsonPathValue(domainPropertyAt, "ETHNICITY", "description").asList().element(0)
+                .isEqualTo("Ethnicity");
+            assertThat(response)
```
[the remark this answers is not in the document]
petergphillips: you can use `assertThatJson` instead in this project.  But I agree, we should just add webflux and be done with it

PR #1181, opened 2022-02-25
NN-3961: Set Adjudication staff victim by username not id

sp-watson:


src/main/java/uk/gov/justice/hmpps/prison/service/AdjudicationsPartyService.java
```diff
@@ -98,7 +98,7 @@ public AdjudicationDetail updateAdjudicationParties(
             entityManager.flush();
         });
         adjudicationRepository.save(adjudication);
-        return AdjudicationsTransformer.transformToDto(adjudication);
+        return;
     }
```
richardpopple, line 102: No longer return?

richardpopple (approved):

PR #1184, opened 2022-03-02
SDI-95 add ordering to domain endpoint

lbennett-moj:


src/main/java/uk/gov/justice/hmpps/prison/service/ReferenceDomainService.java
```diff
@@ -186,6 +186,7 @@ public boolean isReferenceCodeActive(final String domain, final String code) {
                 .parentDomain(it.getParentDomain())
                 .build()
             )
+            .sorted(Comparator.comparing(uk.gov.justice.hmpps.prison.api.model.ReferenceDomain::getDomain))
```
petergphillips, line 189: would be better to get the database to sort instead - there is a version of `findAll` that takes a `Sort`

petergphillips (approved):
LGTM

PR #1187, opened 2022-03-03
SDI-71: 📝 Add in spring docs annotations

petergphillips:


.circleci/config.yml
```diff
@@ -126,23 +126,10 @@ workflows:
             - check
             - test-integration
             - test-schema-nomis
-      - request-stage-approval:
```
petergphillips, line 129: removed the stage deployment so that we have one environment with correct swagger docs

build.gradle.kts
```diff
@@ -45,6 +45,9 @@ dependencies {
   implementation("com.zaxxer:HikariCP:5.0.1")
 
   implementation("io.springfox:springfox-boot-starter:3.0.0")
+  implementation("org.springdoc:springdoc-openapi-ui:1.6.1")
```
petergphillips, line 48: leave spring fox there and just add in springdocs for now

src/main/java/uk/gov/justice/hmpps/prison/web/config/SpringFoxConfiguration.kt
```diff
@@ -1,88 +0,0 @@
-package uk.gov.justice.hmpps.prison.web.config
```
petergphillips, line 1: remove spring fox configuration for now

PR #1190, opened 2022-03-04
CRS-784 fixed term recalls return to custody date.

ldlharper:
CRD are moving on to calculate recall sentences. In order to calculate fixed term recall sentences we need to know the return to custody date. This isn't available in any other endpoints so we've added a new one here.


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderFixedTermRecall.java
```diff
@@ -0,0 +1,41 @@
...
+import java.io.Serializable;
+import java.time.LocalDate;
+import java.util.List;
+
+@Getter
+@Entity
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+@EqualsAndHashCode(of = {"offenderBooking"}, callSuper = false)
```
yasinmustafa, line 19: Is this right for the equals and hashcode?

ldlharper: Yea good spot. It should be bookingId.

neilmendum (approved):
LGTM

PR #1192, opened 2022-03-07
INC-257 Add additional Information about IEP Level for prisoner

Mjwillis:


src/main/java/uk/gov/justice/hmpps/prison/api/model/PrivilegeDetail.java
```diff
@@ -51,4 +53,7 @@
 
     @Schema(description = "Identifier of user related to this privilege entry.")
     private String userId;
+
+    @Schema(description = "The Screen (e.g. NOMIS screen OIDOIEPS) or system (PRISON_API) that made the change", example = "PRISON_API")
+    private String auditModuleName;
 }
```
andymarke, line 59: curious; why do you needs this?

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/api/model/PrivilegeDetail.java
Mjwillis: Its a bit complex, but basically more IEP level entries at NOT actually reviewed, but generated by transferring them from an another prison or correcting a level that is wrong as a result of a transfer. The module name allows us to know it least a person went to the IEP screen in NOMIS or used DPS.

PR #1193, opened 2022-03-07
INC-257 Fix ambiguously defined column

Mjwillis:
Note there are no tests for this!  Will patch in next commit


andymarke (approved):
lgtm

PR #1195, opened 2022-03-09
SDI-71: 🔧 Configure resource limits

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/api/model/AccessRole.java
```diff
@@ -1,41 +0,0 @@
-package uk.gov.justice.hmpps.prison.api.model;
```
petergphillips, line 1: unused files

helm_deploy/prison-api/values.yaml
```diff
@@ -24,6 +24,14 @@ generic-service:
     timeoutSeconds: 30
     failureThreshold: 15
 
+  resources:
+    requests:
+      cpu: 100m
+      memory: 640Mi
+    limits:
+      cpu: 2000m
+      memory: 2048Mi
```
petergphillips, line 33: Screenshot from 2022-03-09 10-20-25
Max recent usage is 1.26Gb, however it hasn't been using all the available heap allocated.  Since our `mx` setting is `1024` I've set this high enough not to hit the limit even if all available heap is allocated.
Nodes are `Standard DS3 v2 (4 vcpus, 14 GiB memory)` so should easily be able to cope as normally only 2 or 3 pods are installed to each node.

lbennett-moj (approved):

PR #1191, opened 2022-03-07
Revert "SDI-71: 📝 Add in spring docs annotations"

petergphillips:
Reverts ministryofjustice/prison-api#1187

PR #1197, opened 2022-03-09
pgp SDI 94 reapply spring docs resources

petergphillips:
- Revert "Revert "SDI-71: 📝 Add in spring docs annotations (#1187)" (#1191)"
- SDI-94: 👷 Exclude model classes from spring doc upgrade


.circleci/config.yml
```diff
@@ -127,23 +127,10 @@ workflows:
             - check
             - test-integration
             - test-schema-nomis
-      - request-stage-approval:
```
petergphillips, line 130: leave out deployment to stage so that one environment has the correct configuration

src/main/java/uk/gov/justice/hmpps/prison/api/model/v1/AccountBalance.java
```diff
@@ -1,15 +1,16 @@
...
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.Hidden;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.ToString;
 
-@ApiModel(description = "Account Balance")
+@Schema(description = "Account Balance")
```
petergphillips, line 13: the `v1` model files use stored procedures and map through intermediate objects so performance is unaffected

src/main/java/uk/gov/justice/hmpps/prison/api/resource/AdjudicationsResource.java
```diff
@@ -1,12 +1,14 @@
 package uk.gov.justice.hmpps.prison.api.resource;
 
-import io.swagger.annotations.Api;
```
petergphillips, line 3: resources can be mapped over too

PR #1201, opened 2022-03-10
SDI-71: ⬆️  Latest versions

petergphillips:


.circleci/config.yml
```diff
@@ -170,6 +170,7 @@ workflows:
       - hmpps/gradle_owasp_dependency_check:
           context:
             - hmpps-common-vars
+          cache_key: "v2_0"
```
petergphillips, line 173: needed as orb was updated before the gradle plugin

Mjwillis (approved):
LGTM

PR #1202, opened 2022-03-11
SDI-71: 📝 Add dtos and convert model classes to latest swagger

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/api/model/CourtEventDto.kt
```diff
@@ -0,0 +1,48 @@
+package uk.gov.justice.hmpps.prison.api.model
+
+import java.time.LocalDate
+import java.time.LocalDateTime
+
+data class CourtEventDto(
+  var offenderNo: String? = null,
```
petergphillips, line 7: would have liked to use the DataClassRowMapper to call the constructor here, but found that calling `getColumn(name)` in Oracle was ridiculously slow so binned the idea (would have required to rewrite the mapper).  Therefore have to make all fields nullable and also `var` so that each property can then be set.

src/main/java/uk/gov/justice/hmpps/prison/repository/MovementsRepository.java
```diff
@@ -49,12 +61,12 @@
         final var sql = MovementsRepositorySql.GET_RECENT_MOVEMENTS_BY_DATE_FOR_BATCH.getSql();
         final var types = (movementTypes == null || movementTypes.isEmpty()) ? Set.of("TRN", "REL", "ADM") : movementTypes;
 
-
-        return jdbcTemplate.query(sql,
+        final var movements = jdbcTemplate.query(sql,
```
petergphillips, line 64: have to get the values and then stream them into the swagger class to avoid the performance penalty of having annotations on a row mapper class

src/main/java/uk/gov/justice/hmpps/prison/repository/MovementsRepository.java
```diff
@@ -191,61 +214,53 @@ public int getEnrouteMovementsOffenderCount(final String agencyId, final LocalDa
...
             MovementsRepositorySql.GET_COURT_EVENTS_BY_AGENCY_AND_TIME_PERIOD.getSql(),
             createParams("agencyListFrom", agencies,
                 "agencyListTo", agencies,
                 "fromDateTime", DateTimeConverter.fromLocalDateTime(from),
                 "toDateTime", DateTimeConverter.fromLocalDateTime(to)),
             COURT_EVENT_MAPPER);
-    }
-
-
-    public List<TransferEvent> getOffenderTransfers(final List<String> agencies, final LocalDateTime from, final LocalDateTime to) {
```
petergphillips, line 221: unused

andymarke (approved):
LGTM

Now if you were really clever you would introduced a new annotation that would have created a DTO at compile time from the model class and added a `toModel` method automatically, like a lombok annotation on steroids :-)

PR #1204, opened 2022-03-15
SDI-71: 🎨 Switch to immutable data classes

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/api/model/MovementSummaryDto.kt
```diff
@@ -1,27 +1,27 @@
...
-  var toProvStatCode: String? = null,
-  var escortCode: String? = null,
-  var escortText: String? = null,
-  var movementType: String? = null,
-  var movementTypeDescription: String? = null,
-  var directionCode: String? = null,
-  var movementTime: LocalDateTime? = null,
-  var movementReason: String? = null,
-  var commentText: String? = null,
+  val offenderNo: String?,
```
andymarke, line 4: So can't these go back to be non-nullable where appropriate ?

PR #1207, opened 2022-03-15
SDI-71: 📝 Add more dtos and convert model classes to latest swagger

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/api/model/StaffLocationRoleDto.kt
```diff
@@ -0,0 +1,45 @@
+package uk.gov.justice.hmpps.prison.api.model
+import java.math.BigDecimal
+import java.time.LocalDate
+
+data class StaffLocationRoleDto(
+  var staffId: Long? = null,
```
petergphillips, line 6: will switch to `val` in next PR

Mjwillis (approved):
LGTM

PR #1211, opened 2022-03-17
SDI-71: 📝 Add more dtos and convert model classes to latest swagger

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/repository/sql/PersonRepositorySql.kt
```diff
@@ -3,8 +3,8 @@ package uk.gov.justice.hmpps.prison.repository.sql
 enum class PersonRepositorySql(val sql: String) {
   GET_PERSON_IDENTIFIERS(
     """
-        select pi.IDENTIFIER_TYPE identifierType,
-        pi.IDENTIFIER identifierValue
+        select pi.IDENTIFIER_TYPE identifier_Type,
```
petergphillips, line 6: the previous mapper used to check both the java and oracle column names, might need to get the new mapper to do the same, but this is the first occurrence I've seen

src/main/java/uk/gov/justice/hmpps/prison/util/QueryBuilderFactory.java
```diff
@@ -27,6 +28,10 @@ public QueryBuilderFactory(@Value("${schema.database.dialect}") final DatabaseDi
         return getQueryBuilder(initialSql, rowMapper.getFieldMap());
     }
 
+    public <T> IQueryBuilder getQueryBuilder(final String initialSql, final DataClassByColumnRowMapper<T> rowMapper) {
```
petergphillips, line 31: new helper for the new mapper

andymarke (approved):
LGTM

PR #1212, opened 2022-03-17
CRS-813 adding court case reference to the CRD sentence endpoint.

ldlharper:
We are required to display the court case reference for CRD service. This is just an extra column from a table we're already joining to for our sentences and offences endpoint


petergphillips (approved):
LGTM

PR #1218, opened 2022-03-24
CRS-785 removing unused fields.

ldlharper:
This endpoint is used by CRD, which has migrated to use the "terms" field and no longer references days,weeks,months,years so they can be removed.


yasinmustafa (approved):
LGTM

PR #1239, opened 2022-04-13
SDI-152 Tiny refactor for webTestClient

andymarke:
Passing the 2 required parameters in during save rather than insisting on a instance variable that looked confusing


petergphillips (approved):
LGTM

PR #1241, opened 2022-04-14
INC-485: return default IEP level for a prison if there is no current IEP for a list of booking IDs

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-485

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/service/BookingService.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/service/BookingService.java
@@ -273,23 +273,7 @@ public class BookingService {
         final var offenderBooking = offenderBookingRepository.findById(bookingId).orElseThrow(EntityNotFoundException.withId(bookingId));
         final var iepSummary = offenderBooking.getIepSummary(withDetails);
 
-        if (iepSummary.isPresent()) {
-            return iepSummary.get();
-        } else {
-            final var iepDefault = availablePrisonIepLevelRepository.findByAgencyLocation_IdAndDefaultIep(offenderBooking.getLocation().getId(), true)
-                .stream().findFirst()
-                .map(iep -> iep.getIepLevel().getDescription())
-                .orElse(defaultIepLevel);
-            final var now = LocalDateTime.now();
-
-            return PrivilegeSummary.builder()
-                .bookingId(bookingId)
-                .iepLevel(iepDefault)
-                .iepTime(now)
-                .iepDate(now.toLocalDate())
-                .daysSinceReview(0L)
-                .build();
-        }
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/justice/hmpps/prison/service/BookingServiceTest.java
```diff
@@ -367,16 +367,37 @@ public void getBookingIEPSummary_offenderHasNoIepSummary() {
     public void getBookingIEPSummary_multipleBooking_globalSearchUser() {
         when(authenticationFacade.isOverrideRole(anyString(), anyString(), anyString(), anyString(), anyString())).thenReturn(true);
         when(bookingRepository.getBookingIEPDetailsByBookingIds(anyList())).thenReturn(Map.of(-5L, List.of(PrivilegeDetail.builder().iepDate(LocalDate.now()).build())));
+        when(offenderBookingRepository.findById(anyLong())).thenReturn(Optional.of(OffenderBooking.builder().location(AgencyLocation.builder().id("MDI").build()).build()));
```
neilmendum, line 370: Had to update existing tests because I introduced `offenderBookingRepository.findById` and without this mock they were throwing `EntityNotFoundException`

src/main/java/uk/gov/justice/hmpps/prison/service/BookingService.java
```diff
@@ -342,16 +326,32 @@ public Map<Long, PrivilegeSummary> getBookingIEPSummary(final List<Long> booking
...
                 .filter(bookingId -> !mapOfEip.containsKey(bookingId))
-                .collect(toList())
-                .forEach(bookingId -> mapOfEip.put(bookingId, PrivilegeSummary.builder()
-                        .bookingId(bookingId)
-                        .iepLevel(defaultIepLevel)
-                        .iepDetails(Collections.emptyList())
-                        .build()));
+                .toList()
+                .forEach(bookingId -> {
+                    final var offenderBooking = offenderBookingRepository.findById(bookingId).orElseThrow(EntityNotFoundException.withId(bookingId));
```
neilmendum, line 331: Not ideal doing another lookup but it's needed so we can get a prison's default IEP. This is an edge case so not expecting it to called frequently

src/test/java/uk/gov/justice/hmpps/prison/service/BookingServiceTest.java
Mjwillis: Does this definitely only get called if no IEP found?

Mjwillis: Ignore I can see that it does

Mjwillis (approved):
LGTM

PR #1253, opened 2022-04-25
INC-481: mark OffenderResource IEP endpoint as deprecated

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-481


Mjwillis (approved):
Lgtm

PR #1258, opened 2022-04-25
SDI-152 Update test for VO balances after transfer

andymarke:
NOMIS was suppose to update visit order balances after a transfer that reset the IEP level.

Issues with that NOMIS release meant it was pulled, so effectively VO balances are not updated after a transfer.

This change documents that behaviour with pointers to what will need to change in the future


petergphillips (approved):
LGTM

PR #1260, opened 2022-04-26
SDI-172 Add basic tests for returning from court

andymarke:
Requires basic endpoint to transfer out to court, this is an amended version of the transfer-out (also not used in production)

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
@@ -55,6 +55,7 @@ import uk.gov.justice.hmpps.prison.api.model.RequestToRecall;
 import uk.gov.justice.hmpps.prison.api.model.RequestToReleasePrisoner;
 import uk.gov.justice.hmpps.prison.api.model.RequestToTransferIn;
 import uk.gov.justice.hmpps.prison.api.model.RequestToTransferOut;
+import uk.gov.justice.hmpps.prison.api.model.RequestToTransferOutToCourt;
 import uk.gov.justice.hmpps.prison.api.model.ScheduledEvent;
 import uk.gov.justice.hmpps.prison.api.model.SentenceSummary;
 import uk.gov.justice.hmpps.prison.api.model.UpdateCaseNote;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
```diff
@@ -241,6 +242,20 @@ public InmateDetail transferOutPrisoner(
         @RequestBody @NotNull @Valid final RequestToTransferOut requestToTransferOut) {
         return prisonerReleaseAndTransferService.transferOutPrisoner(offenderNo, requestToTransferOut);
     }
+    @ApiResponses({
```
andymarke, line 245: API required just for testing, but can the basis of a future API to transfer to court but will required more analysis and testing

src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerReleaseAndTransferService.java
```diff
@@ -251,6 +252,41 @@ public InmateDetail transferOutPrisoner(final String prisonerIdentifier, final R
 
         return offenderTransformer.transform(booking);
     }
+    @VerifyOffenderAccess(overrideRoles = {"TRANSFER_PRISONER_ALPHA"})
+    public InmateDetail transferOutPrisonerToCourt(final String prisonerIdentifier, final RequestToTransferOutToCourt requestToTransferOutToCourt) {
+        // NB This API requires further validation before it can be used for production - it is currently used just to generate test data
```
andymarke, line 257: clone of above API but for courts and checked against current NOMIS behaviour

src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerReleaseAndTransferService.java
```diff
@@ -752,6 +788,7 @@ public InmateDetail courtTransferIn(final String offenderNo, final RequestForCou
 
         final MovementReason movementReason = getMovementReason(requestForCourtTransferIn.getMovementReasonCode(), latestExternalMovement.getMovementReason());
 
+        offenderBooking.setStatusReason(MovementType.CRT.getCode() + "-" + movementReason.getCode());
```
andymarke, line 791: minor bug in existing transfer from court, status reason was never being set so remained at the reason set during the transfer out to court. 
It would only be noticeable if they change the court reason when returning back from court (which I guess WPIP never does)

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffendersResourceTransferImpTest.kt
```diff
@@ -472,7 +478,221 @@ class OffendersResourceTransferImpTest : ResourceTest() {
...
+          }
+                  """.trimIndent()
+                )
+              )
+              .exchange()
+              .expectStatus().isOk
+              .expectBody()
+              .jsonPath("assignedLivingUnit.agencyId").isEqualTo("LEI")
+              .jsonPath("assignedLivingUnit.description").isEqualTo("COURT") // as set when bed was released
+          }
```
andymarke, line 603: when bed is released the cell is set to "COURT"

lbennett-moj (approved):

PR #1261, opened 2022-04-26
SDI-172 Failing (disabled) tests describing basic transfer via court behaviour

andymarke:
Still need to write tests describing what happens at original prison; activities, bed capacity, visits etc


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffendersResourceTransferImpTest.kt
```diff
@@ -526,15 +529,12 @@ class OffendersResourceTransferImpTest : ResourceTest() {
...
-          {
-            "agencyId":"LEI",
-            "commentText":"admitted",
-            "movementReasonCode":"CRT",
-            "dateTime": "${
-                  LocalDateTime.now().minusMinutes(2).format(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
-                  }"
-            
-          }
+                  {
```
andymarke, line 532: little bit of formatting ...

PR #1263, opened 2022-04-26
SDI-172 test builders and framework to support ending activities and …

lbennett-moj:
…wait list allocations, tests in progress


src/test/java/uk/gov/justice/hmpps/prison/util/OffenderProgramProfileBuilder.kt
```diff
@@ -0,0 +1,47 @@
...
+  var prisonId: String = "MDI",
+  var startDate: LocalDate = LocalDate.of(2016, 11, 9),
+  var programStatus: String = "ALLOC",
+  var courseActivityId: Long = -1,
+  var offenderBookingId: Long = -1,
+  var programId: Long = -1
+) : WebClientEntityBuilder() {
+
+  fun save(
+    courseActivityRepository: CourseActivityRepository? = null,
```
andymarke, line 22: Maybe a future change will be to add a service method to add a program that is only used by tests; that means we only need the service and not all the repositories at the test level.

Something to think about

andymarke (approved):
LGTM

PR #1264, opened 2022-04-27
SDI-180 Add temporary endpoint while teh transfer via court is built

andymarke:


src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
```diff
@@ -289,6 +289,22 @@ public InmateDetail courtTransferIn(
         return prisonerReleaseAndTransferService.courtTransferIn(offenderNo, requestForCourtTransferIn);
     }
 
+    @ApiResponses({
```
andymarke, line 292: once this is built and ready to test we will remove the old version and this will be renamed back

Mjwillis (approved):
Lgtm

PR #1271, opened 2022-04-29
SDI-180 Test for court hearing events

andymarke:
This requires the ability for the ALPHA version of court transfer out to associate with court events
Also the ability to create court cases and court hearings


Mjwillis (commented):
LGTM

PR #1273, opened 2022-05-03
CRS-855 adding endpoint to get calculated dates in the same format th…

ldlharper:
…at CRD sends.

The ESL and JSL are internal dates that NOMIS calculates. When CRD service calculates a release date we send over a sentence length that we've calculated. This sentence length is persisted to the ESL and JSL columns. This data is used in reporting and we need to support it.

This PR exposes the ESL and JSL so that we can compare the dates NOMIS calculates vs the dates CRD service calculates as part of our regression testing.

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderKeyDates.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderKeyDates.java
@@ -61,4 +61,7 @@ public class OffenderKeyDates {
 
     @Schema(required = true, description = "Sentence length in the format 00 years/00 months/00 days.", example = "11/00/00")
     private String sentenceLength;
+
+    @Schema(description = "Judicially imposed length in the format 00 years/00 months/00 days. Will default to 'sentenceLength'", example = "11/00/00")
+    private String judiciallyImposedSentenceLength;
 }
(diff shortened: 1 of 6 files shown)


Mjwillis (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderKeyDates.java
```diff
@@ -61,4 +61,7 @@ public class OffenderKeyDates {
 
     @Schema(required = true, description = "Sentence length in the format 00 years/00 months/00 days.", example = "11/00/00")
     private String sentenceLength;
+
+    @Schema(description = "Judicially imposed length in the format 00 years/00 months/00 days. Will default to 'sentenceLength'", example = "11/00/00")
+    private String judiciallyImposedSentenceLength;
```
andymarke, line 66: Is there any reason we couldn't make this a small object even if we think people won't use it ?
{
 years: Int
months: Int
days: Int
}

Or whatever?
Maybe not worth it?

andymarke (approved):
lgtm

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderKeyDates.java
ldlharper: Yea I think that would be better. We're just matching what nomis is doing here. In CRD its modelled as a java Period, which works really well. Hopefully these fields will be going away soon, if not I'll look at changing it.

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/api/model/OffenderKeyDates.java
Mjwillis: Yes!

PR #1283, opened 2022-05-06
SDI-187 Test for team assignment side effects 

andymarke:
H2 schema was missing TEAMS and associated sequence so added that as well


src/main/resources/db/migration/nomis/ddl/V1_413__TEAMS.sql
```diff
@@ -0,0 +1,41 @@
+create table TEAMS
```
andymarke, line 1: generated from T3 oracle schema

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffendersResourceTransferImpTest.kt
```diff
@@ -54,6 +62,13 @@ class OffendersResourceTransferImpTest : ResourceTest() {
   @Autowired
   private lateinit var dataLoader: DataLoaderRepository
 
+  @MockBean
+  private lateinit var workflowTaskService: WorkflowTaskService
+
+  private val team: Team by lazy {
+    dataLoader.load(TeamBuilder())
+  }
+
```
andymarke, line 71: we only need one team that is static created rather then one per test since we never mutate the team

lbennett-moj (approved):

PR #1287, opened 2022-05-11
SDI-189: ✨ Try out buildx for multi arch builds

petergphillips:


Dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM openjdk:17-slim AS builder
+FROM --platform=$BUILDPLATFORM openjdk:17-slim AS builder
```
petergphillips, line 1: the `builder` step produces class files which are platform agnostic.  Therefore only need to do this step on the build platform architecture

Dockerfile
```diff
@@ -1,11 +1,11 @@
...
 
 ARG BUILD_NUMBER
 ENV BUILD_NUMBER ${BUILD_NUMBER:-1_0_0}
 
 WORKDIR /app
 ADD . .
-RUN ./gradlew assemble -Dorg.gradle.daemon=false
+RUN ./gradlew --no-daemon assemble
```
petergphillips, line 8: slightly nicer syntax

Mjwillis (approved):
LGTM

PR #1290, opened 2022-05-11
NN-4065 changes to check for restricted patient if a POM user

timoj58:
1: sql added to check for restricted patient if pom role set, and the first call has failed.  This would only fail as either the POM has decided to change URLs and guess at ids, or they have been directed their by an RP.  restricted patient api always updates nomis with the exterrnal movement record. Note  SQL does not use dense_rank as the in memory testing did not like it (postgres / oracle are fine with it).
2: had to add the Authentication object for testing due to use of static methods, added a test in for INTERNAL_BOOKINGS now we can.

3: i would like to test this in development to fully understand the impact of this, so it may well get rolled back but there is limited things i can do locally with it.


timoj58:
directed to make changes in DPS

PR #1295, opened 2022-05-13
SDI-195: 🐛 Return activities which end halfway through the date range

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/repository/sql/ScheduleRepositorySql.kt
```diff
@@ -37,7 +37,7 @@ enum class ScheduleRepositorySql(val sql: String) {
         AND OCA.CRS_SCH_ID = CS.CRS_SCH_ID
         WHERE CA.INTERNAL_LOCATION_ID = :locationId
         AND (OPP.OFFENDER_PROGRAM_STATUS = 'ALLOC' OR
-        (OPP.OFFENDER_PROGRAM_STATUS = 'END' AND OPP.OFFENDER_END_DATE >= :toDate))
+        (OPP.OFFENDER_PROGRAM_STATUS = 'END' AND OPP.OFFENDER_END_DATE >= CS.SCHEDULE_DATE))
```
petergphillips, line 40: the issue here was that the program ended halfway through a week and before the `toDate` so therefore wasn't returned.  We now restrict to ensure that the program ends on or after the scheduled event date.

yasinmustafa (approved):
LGTM

PR #1297, opened 2022-05-16
SDI-147: 🐛 Fix running tests in intellij

petergphillips:


build.gradle.kts
```diff
@@ -150,6 +150,4 @@ tasks {
       jvmTarget = "17"
     }
   }
-
-  named("test") { dependsOn("testIntegration", "testWithSchemaNomis") }
```
petergphillips, line 154: this meant that when running the `test` profile in intellij integration and test with schema nomis were always run as well

andrewrlee (approved):
LGTM

This was bugging me!

PR #1303, opened 2022-05-20
NN-4156 history sort uses hearing time when present

timoj58:
previously sorted on report_time which can be the same for multiple records, using hearing time if present will ensure the headline item used in DPS is always first.


src/test/java/uk/gov/justice/hmpps/prison/repository/AdjudicationsRepositoryTest.java
```diff
@@ -134,6 +137,22 @@ public void retrieveAdjudicationsForOffender() {
         assertThat(results.getItems()).containsExactly(LATEST_ADJUDICATION, MIDDLE_ADJUDICATION, EARLIEST_ADJUDICATION);
     }
 
+    @Sql(scripts = {"/sql/adjudicationHistorySort_init.sql"},
```
andymarke, line 140: Good.
Going forward it might be worth looking at the pattern  of builders in src/test/java/uk/gov/justice/hmpps/prison/util/builders where we attempting to creating a set of builders that read in a fluid way to improve integration testing.

No changes required now but the concept of creating data on fly like you have done is so much better than relying on seed data.

andymarke (approved):
LGTM

PR #1304, opened 2022-05-20
NN-4156

timoj58:
report time should really be combination of hearing time / report time


timoj58:
leaving this here.  I am not sure yet its required, need to fix test, but the reportTime of the whole adjudication needs to be discovered.

PR #1306, opened 2022-05-23
SDI-212: ✨ New endpoint to count activities

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/service/SchedulesService.java
```diff
@@ -175,6 +180,13 @@ public List<PrisonerSchedule> getActivitiesAtAllLocations(final String agencyId,
         return filterByTimeSlot(timeSlot, activities);
     }
 
+    @VerifyAgencyAccess(overrideRoles = {"SYSTEM_USER", "GLOBAL_SEARCH"})
+    public PrisonerActivitiesCount getCountActivities(String agencyId, LocalDate fromDate, LocalDate toDate, List<TimeSlot> timeSlots) {
+        final var startDate = fromDate == null ? LocalDate.now() : fromDate;
```
mikehalmamoj, line 185: Can maybe do `startDate = fromDate ?: LocalDate.now()`? (and for line below)

mikehalmamoj: Oh sorry! It's Java 🤣

mikehalmamoj: Ah so am I right in thinking that if you don't pas in start and end dates then it just retrieves the stats for today? Just wondering if it's worth writing a test for that to demonstrate?

mikehalmamoj (approved):

src/main/java/uk/gov/justice/hmpps/prison/service/SchedulesService.java
petergphillips: yes, that's correct.  I'll deal with it in subsequent PR since I need to check what whereabouts will be passing through. If the date parameters are required there then no point in setting defaults in Prison API

PR #1312, opened 2022-05-25
SDI-212: 🐛 Switch to counting in java / kotlin and filter not recorded

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/PrisonerActivitiesCountRepository.kt
```diff
@@ -16,8 +17,9 @@ interface PrisonerActivitiesCountRepository : CrudRepository<PrisonerActivitiesC
   @Query(
     """
         SELECT 
-          COUNT(*) as total,
```
petergphillips, line 19: When this was run in production it was exceedingly slow, even though in pre-prod is was quite fast.
I've therefore switched to the existing method of retrieving the bookings and then filtering in the java / kotlin instead in the hope that it helps

src/main/java/uk/gov/justice/hmpps/prison/api/resource/ScheduleResource.java
```diff
@@ -117,12 +119,13 @@ public List<PrisonerSchedule> getActivitiesAtAllLocationsByDateRange(@PathVariab
...
                                                                  @RequestParam(value = "fromDate") @DateTimeFormat(iso = ISO.DATE) @Parameter(description = "From date of whereabouts list") @NotNull final LocalDate fromDate,
                                                                  @RequestParam(value = "toDate") @DateTimeFormat(iso = ISO.DATE) @Parameter(description = "To Date of whereabouts list") @NotNull final LocalDate toDate,
-                                                                 @RequestParam(value = "timeSlots") @Parameter(description = "List of one or more of AM, PM or ED") @NotEmpty final List<TimeSlot> timeSlots) {
-        return schedulesService.getCountActivities(agencyId, fromDate, toDate, timeSlots);
+                                                                 @RequestParam(value = "timeSlots") @Parameter(description = "Set of one or more of AM, PM or ED") @NotEmpty final Set<TimeSlot> timeSlots,
+                                                                 @RequestBody @Parameter(description = "Map of booking IDs to their occurrence counts") final Map<Long, Long> attendanceCounts) {
```
petergphillips, line 127: Changed to a `POST` so can send in a map of the booking ids (and their counts) that have been recorded in whereabouts.

src/main/java/uk/gov/justice/hmpps/prison/service/SchedulesService.java
```diff
@@ -181,8 +185,16 @@ public List<PrisonerSchedule> getActivitiesAtAllLocations(final String agencyId,
...
-        return prisonerActivitiesCountRepository.getCountActivities(agencyId, fromDate, toDate, timeSlots.stream().map(Enum::name).toList());
+    public PrisonerActivitiesCount getCountActivities(String agencyId, LocalDate fromDate, LocalDate toDate, Set<TimeSlot> timeSlots, Map<Long, Long> attendanceCounts) {
+        final var activities = prisonerActivitiesCountRepository.getActivities(agencyId, fromDate, toDate);
+        final var activitiesInTimeSlots = activities.stream()
+            .filter(p -> CalcDateRanges.eventStartsInTimeslots(p.getStartTime(), timeSlots)).toList();
+
+        final var total = activitiesInTimeSlots.size();
+        final var suspended = activitiesInTimeSlots.stream().filter(p -> "Y".equals(p.getSuspended())).count();
+        final var bookingIds = activitiesInTimeSlots.stream().map(PrisonerActivity::getBookingId).collect(Collectors.toSet());
+        final var relevantAttendanceCounts = attendanceCounts.entrySet().stream().filter(a -> bookingIds.contains(a.getKey())).map(Entry::getValue).reduce(0L, Long::sum);
```
petergphillips, line 196: This replicates the existing functionality that exists in whereabouts.
We only want to count the attendances if the booking is in the scheduled list.  This means that prisoners who attended at the time but don't have an active booking any more are excluded from the not recorded stats.

PR #1314, opened 2022-05-26
NN-4171 ensure both roles methods strip ROLE_

timoj58:
issue found with verifyOffenderAccess.  It uses the isOverrideRole and the default annotation is defined as ROLE_


Mjwillis (approved):
LGTM

PR #1315, opened 2022-05-26
SDI-212: 🐛 Remove entity as was not hydrating properly

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/PrisonerActivity.kt
```diff
@@ -0,0 +1,11 @@
+package uk.gov.justice.hmpps.prison.repository.jpa.repository
+
+import java.time.LocalDateTime
+
+interface PrisonerActivity {
```
petergphillips, line 5: moved `PrisonerActivity` to an interface and not a managed type as spring data was using the `@Id` of the booking for uniqueness so not hydrating all rows properly

mikehalmamoj (approved):

PR #1316, opened 2022-05-31
OFFENCES-81 Change offence related POST  and PUT endpoints to accept Lists rather than single objects

yasinmustafa:
Also, in the creation of statutes or ho_codes, do not throw a 409 exception if a record already exists with the same PK, instead just skip that record.


joelstobart-moj (approved):
LGTM

PR #1321, opened 2022-06-09
SDI-226: 🐛 Fix suspended counts to include programs that end halfway through

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/api/model/PrisonerScheduleDto.kt
```diff
@@ -57,4 +62,9 @@ data class PrisonerScheduleDto(
     this.locationCode,
     this.suspended,
   )
+
+  fun programHasntEnded(): Boolean =
+    // SQL ensures that the activity happens after the start date parameter, now need to check that the offender
+    // program hasn't ended. END indicates that the program has ended, in which case the end date will also be populated
+    programStatus != "END" || programEndDate == null || scheduleDate == null || programEndDate >= scheduleDate
```
petergphillips, line 69: kotlin can compare dates using `>=` 😁

src/main/java/uk/gov/justice/hmpps/prison/repository/sql/ScheduleRepositorySql.kt
```diff
@@ -84,7 +87,7 @@ enum class ScheduleRepositorySql(val sql: String) {
         AND OCA.CRS_SCH_ID = CS.CRS_SCH_ID
                 WHERE CA.AGY_LOC_ID = :agencyId
         AND (OPP.OFFENDER_PROGRAM_STATUS = 'ALLOC'
-                OR (OPP.OFFENDER_PROGRAM_STATUS = 'END' AND OPP.OFFENDER_END_DATE >= :toDate))
+                OR (OPP.OFFENDER_PROGRAM_STATUS = 'END' AND OPP.OFFENDER_END_DATE >= :fromDate))
```
petergphillips, line 90: have to restrict by the `fromDate` here and then `filter` in the kotlin.  Tried to add in the `SCHEDULE_DATE` into the criteria, but then the explain plan looked really nasty.

mikehalmamoj (approved):

PR #1329, opened 2022-06-15
DCS-1674 Call correct stored procedure

andrewrlee:
This will copy across the workflow data associated with alerts


GurnankCheema (approved):
lgtm

PR #1342, opened 2022-06-28
SDI-255: ✨ New endpoint to return database restore information

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/web/config/ResourceServerConfiguration.java
```diff
@@ -32,7 +32,7 @@ public void configure(final HttpSecurity http) throws Exception {
                         "/health/**", "/info", "/ping", "/h2-console/**",
                         "/v3/api-docs/**", "/swagger-ui.html", "/swagger-ui/**",
                         "/swagger-resources", "/swagger-resources/configuration/ui",
-                        "/swagger-resources/configuration/security").permitAll()
+                        "/swagger-resources/configuration/security", "/api/restore-info").permitAll()
```
petergphillips, line 35: don't need the new endpoint to be protected.  Was thinking about adding to the `/info` endpoint but then thought it wasn't needed most of the time so discounted that idea

mikehalmamoj (approved):

PR #1344, opened 2022-06-29
SDI-256 and DCS-1638 Fix youth flag being set

andymarke:
Attempts to set youth flag failed since the copy booking stored procedure updated booking record at same time as API causing primary key violation.

Mapping for profile also meant updates never worked anyway

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderProfileDetail.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderProfileDetail.java
@@ -61,6 +61,13 @@ public class OffenderProfileDetail extends AuditableEntity {
     @NotFound(action = IGNORE)
     private ProfileCode code;
 
+    public void setProfileCode(ProfileCode code) {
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderProfileDetail.java
```diff
@@ -61,6 +61,13 @@ public static class PK implements Serializable {
     @NotFound(action = IGNORE)
     private ProfileCode code;
 
+    public void setProfileCode(ProfileCode code) {
+        this.profileCode = code.getId().getCode();
+    }
+
+    @Column(name = "PROFILE_CODE")
+    private String profileCode;
+
```
andymarke, line 70: ProfileCode is immutable due to mapping (update bale false etc), this is due to ProfileType mapped as primary key above.  So double map the actual code so it is updatable - hence need for strange `setProfileCode` helper

petergphillips (approved):
LGTM

PR #1345, opened 2022-06-29
SDI-256 Ensure profiles with no code can be updated to prevent 2 rows being created.

andymarke:
Any row in OFFENDER_PROFILE_DETAILS with  a null code was accidentally not being read via JPA due to the mapping.
This change allows them to be read so that a row for YOUTH can be updated with the new code; without this both JPA and the NOMIS tore procedure will try to create the a row with the same primary key

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderBooking.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderBooking.java
@@ -483,6 +483,7 @@ public class OffenderBooking extends AuditableEntity {
 
     public List<OffenderProfileDetail> getActiveProfileDetails() {
         return profileDetails.stream()
+            .filter(pd -> Objects.nonNull(pd.getCode()))
             .filter(pd -> {
                 final var profileType = pd.getId().getType();
                 return profileType.getCategory().equals("PI") && (profileType.isActive() || profileType.getType().equals("RELF"));
--- a/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderProfileDetail.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderProfileDetail.java
@@ -53,10 +53,10 @@ public class OffenderProfileDetail extends AuditableEntity {
         private Integer sequence;
     }
 
-    @ManyToOne(optional = false)
+    @ManyToOne
     @JoinColumns(value = {
         @JoinColumn(name = "PROFILE_CODE", referencedColumnName = "PROFILE_CODE", insertable = false, updatable = false),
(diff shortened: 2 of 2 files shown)


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderBooking.java
```diff
@@ -483,6 +483,7 @@ public List<OffenderPropertyContainer> getActivePropertyContainers() {
 
     public List<OffenderProfileDetail> getActiveProfileDetails() {
         return profileDetails.stream()
+            .filter(pd -> Objects.nonNull(pd.getProfileCode()))
```
andymarke, line 486: not really needed since when this is used the nulls are filtered out anyway by the caller, but given the method name  this at least means this method behaves the same after the mapping change below

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderProfileDetail.java
```diff
@@ -53,10 +53,10 @@ public static class PK implements Serializable {
         private Integer sequence;
     }
 
-    @ManyToOne(optional = false)
```
andymarke, line 56: the code is indeed optional, and typically not present for rows created by NOMIS

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderProfileDetail.java
```diff
@@ -53,10 +53,10 @@ public static class PK implements Serializable {
...
+    @ManyToOne
     @JoinColumns(value = {
         @JoinColumn(name = "PROFILE_CODE", referencedColumnName = "PROFILE_CODE", insertable = false, updatable = false),
-        @JoinColumn(name = "PROFILE_TYPE", referencedColumnName = "PROFILE_TYPE", nullable = false, insertable = false, updatable = false)
```
andymarke, line 59: though type isn't nullable (and is mapped correctly in the primary key above), in this context it makes no sense to describe it as not  nullable when the entire link is nullable - and JPA won't allow it anyway

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderBooking.java
```diff
@@ -483,6 +483,7 @@ public List<OffenderPropertyContainer> getActivePropertyContainers() {
 
     public List<OffenderProfileDetail> getActiveProfileDetails() {
         return profileDetails.stream()
+            .filter(pd -> Objects.nonNull(pd.getCode()))
```
andymarke, line 486: so that this method returns same values, filter out the extra records.
(Since the only client of this method filtered nulls anyway there is no test changes - but feels right not to change this behaviour)

lbennett-moj (approved):
LGTM

PR #1347, opened 2022-06-30
SDI-256 Ensure profile code is actually saved

andymarke:
Integration test wasn't actually ensure the data was written correctly so now do a GET.
(also actually fixed bug)


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffendersResourceTest.java
```diff
@@ -1020,6 +1021,21 @@ public void testReceiveAPrisonerAsNewBookingForFirstTime() {
         );
         assertThatJsonFileAndStatus(response, 200, "first_new_booking_prisoner.json");
```
andymarke, line 1022: this test past even though data was never actually being written correctly

lbennett-moj (approved):

PR #1348, opened 2022-07-01
SDI-259 Moved test to new format so it can easily be extended

andymarke:
Preparation for a load more integration tests for new booking endpoint by moving to own file and using more flexible testing framework (webTestClient, Kotlin etc)


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceNewBookingTest.kt
```diff
@@ -0,0 +1,118 @@
...
+          .jsonPath("lastMovementTypeCode").isEqualTo("ADM")
+          .jsonPath("lastMovementReasonCode").isEqualTo("24")
+          .jsonPath("agencyId").isEqualTo("SYI")
+          .jsonPath("assignedLivingUnit.agencyId").isEqualTo("SYI")
+          .jsonPath("assignedLivingUnit.description").isEqualTo("A-1-1")
+          .jsonPath("imprisonmentStatus").isEqualTo("CUR_ORA")
+          .jsonPath("bookingId").exists()
+          .jsonPath("bookingNo").exists()
+      }
+    }
```
andymarke, line 116: old test did a release which I assume to reset status, we don't need to since the prisoner is created on the fly and we throw it away

petergphillips (approved):
LGTM

PR #1350, opened 2022-07-01
SDI-257: 🔒️ Suppress false positive for log4j

petergphillips:
- SDI-261: 💩 Force pinning of versions so that upgrade versions ignores it each time
- SDI-257: 🔒️ Suppress false positive for log4j


SimonMitchellMOJ (approved):
LGTM

PR #1351, opened 2022-07-01
SDI-259 Added test with disabled failing test for youth profile 

andymarke:
To be able to write test to show issue this require an emulation of a NOMIS stored procedure that copies booking data from one booking to another.
This has a cut down version of that functionality that will copy OFFENDER_PROFILE_DETAILS

Previous store procedure moved to Spring bean so we can have a Oracle version and a H2 version.

We have two failing tests that are disabled


src/main/java/uk/gov/justice/hmpps/prison/service/createbooking/CopyPreviousBookingService.kt
```diff
@@ -0,0 +1,64 @@
...
+    jdbcTemplate.update(
+      """INSERT INTO offender_profile_details (offender_book_id, profile_seq, profile_type, profile_code, list_seq, comment_text, caseload_type) 
+        |SELECT ?, profile_seq, profile_type, profile_code, list_seq, comment_text, caseload_type 
+        |FROM offender_profile_details 
+        |WHERE offender_book_id = ?
+        |""".trimMargin()
+    ) { ps ->
+      ps.setLong(1, booking.bookingId)
+      ps.setLong(2, previousBooking.bookingId)
+    }
```
andymarke, line 62: just enough copying to prove  the create booking service works when this table is amended outside of the JPA world

src/test/java/uk/gov/justice/hmpps/prison/util/builders/OffenderBookingBuilder.kt
```diff
@@ -10,13 +10,13 @@ import uk.gov.justice.hmpps.prison.api.model.RequestForNewBooking
 import uk.gov.justice.hmpps.prison.service.DataLoaderRepository
 import uk.gov.justice.hmpps.prison.util.JwtAuthenticationHelper
 import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
 
 class OffenderBookingBuilder(
   var prisonId: String = "MDI",
-  var bookingInTime: LocalDateTime = LocalDateTime.now(),
+  var bookingInTime: LocalDateTime = LocalDateTime.now().minusDays(1),
   var fromLocationId: String = "OUT",
```
andymarke, line 18: since we will now releases some of the prisoners make sure their booking time is way in the past

src/test/java/uk/gov/justice/hmpps/prison/util/builders/OffenderBookingBuilder.kt
```diff
@@ -64,7 +71,7 @@ class OffenderBookingBuilder(
     val request =
       RequestForNewBooking.builder().bookingInTime(bookingInTime).cellLocation(cellLocation)
         .fromLocationId(fromLocationId).imprisonmentStatus(imprisonmentStatus).movementReasonCode(movementReasonCode)
-        .prisonId(prisonId).youthOffender(youthOffender).build()
+        .prisonId(prisonId).build()
```
andymarke, line 74: move the setting of youth offender to the OffenderProfileDetailsBuilder precisely because it doesn't work correctly via the service

src/test/java/uk/gov/justice/hmpps/prison/util/builders/OffenderBookingBuilder.kt
```diff
@@ -82,6 +89,33 @@ class OffenderBookingBuilder(
       .exchange()
       .expectStatus().isOk
       .returnResult<InmateDetail>().responseBody.blockFirst()!!.also {
+      if (released) {
```
andymarke, line 92: convent method of creating an inactive booking; book them in and then release them

src/test/java/uk/gov/justice/hmpps/prison/util/builders/OffenderProfileDetailsBuilder.kt
```diff
@@ -0,0 +1,32 @@
...
+  YOUTH
+}
+
+class OffenderProfileDetailsBuilder(
+  private val code: String,
+  private val type: ProfileType,
+) {
+
+  // use SQL to save because the JPA mapping may not be correct
+  fun save(
```
andymarke, line 15: we could use the JPA repo to do this, but SQL is more reliable exactly because there is issue to be fixed in the JPA, eventually we will relay on either JPA or the service for creating the test data, but for now use good old fashioned SQL

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceNewBookingTest.kt
```diff
@@ -113,6 +117,391 @@ class OffenderResourceNewBookingTest : ResourceTest() {
           .jsonPath("bookingId").exists()
           .jsonPath("bookingNo").exists()
       }
+
+      @Test
+      @Disabled("TODO: fix this test")
```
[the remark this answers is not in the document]
andymarke: If I can be bothered :-) 

Yes, indeed - that is what SDI-259 is all about - but a hell of lot of work to show it doesn't work because of the interaction between our service and the NOMIS store procedure

Mjwillis (approved):
LGTM

PR #1352, opened 2022-07-04
SDI-259 Added create booking integration tests describing most of the behaviour

andymarke:
Still need tests that assert store procedures are called, that requires a small refactor

### DIFF
--- a/src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceNewBookingTest.kt
+++ b/src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceNewBookingTest.kt
@@ -1,5 +1,7 @@
 package uk.gov.justice.hmpps.prison.api.resource.impl
 
+import org.assertj.core.api.Assertions.assertThat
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceNewBookingTest.kt
```diff
@@ -503,5 +863,290 @@ class OffenderResourceNewBookingTest : ResourceTest() {
...
+      prisonId = "MDI",
+      released = true
+    ).withIEPLevel(iepLevel)
+  ).save(
+    webTestClient = webTestClient,
+    jwtAuthenticationHelper = jwtAuthenticationHelper,
+    dataLoader = dataLoader
+  ).offenderNo
+
+  private fun getMovements(bookingId: Long) = externalMovementRepository.findAllByOffenderBooking_BookingId(bookingId)
```
andymarke, line 1093: A load of these methods are duplicated in the transfer test; might see how they can be shared

PR #1353, opened 2022-07-05
SDI-259 Added a notion of BuilderContext

andymarke:
So that we can add helpers using extensions rather then polluting the base ResourceTest class added builder context that has everything you would ever need to add or retrieve data.

Next step is to add some extensions e.g `releasePrisoner` so that it can be shared between tests

### DIFF
--- a/src/test/java/uk/gov/justice/hmpps/prison/service/DataLoaderRepository.kt
+++ b/src/test/java/uk/gov/justice/hmpps/prison/service/DataLoaderRepository.kt
@@ -2,7 +2,6 @@ package uk.gov.justice.hmpps.prison.service
 
 import org.springframework.jdbc.core.JdbcTemplate
 import org.springframework.stereotype.Service
-import org.springframework.test.web.reactive.server.WebTestClient
 import uk.gov.justice.hmpps.prison.repository.BookingRepository
 import uk.gov.justice.hmpps.prison.repository.jpa.model.CaseStatus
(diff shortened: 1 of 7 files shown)


src/test/java/uk/gov/justice/hmpps/prison/service/DataLoaderRepository.kt
```diff
@@ -36,24 +35,23 @@ class DataLoaderRepository(
   val institutionAreaRepository: ReferenceCodeRepository<InstitutionArea>,
   val teamCategoryRepository: ReferenceCodeRepository<TeamCategory>,
   val jdbcTemplate: JdbcTemplate,
-) {
+)
+
+@Service
+class DataLoaderTransaction {
```
andymarke, line 41: It was always confusing have this transactional methods on the loader itself, so create a special class that used purely to create a transaction when loading data

src/test/java/uk/gov/justice/hmpps/prison/util/builders/BuilderHelpers.kt
```diff
@@ -1,8 +1,18 @@
-package uk.gov.justice.hmpps.prison.util
+package uk.gov.justice.hmpps.prison.util.builders
```
andymarke, line 1: package was always wrong - never noticed before

petergphillips (approved):
LGTM

PR #1354, opened 2022-07-05
SDI-259 Moved duplicate helpers to single file

andymarke:
It uses the BuilderContext to group all the data helper functions, I think this is a better approach then adding to the base class that is too big as it stands


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceNewBookingTest.kt
```diff
@@ -119,7 +114,7 @@ class OffenderResourceNewBookingTest : ResourceTest() {
 
       @Test
       internal fun `400 when offender is inactive but not OUT, for instance currently being transferred`() {
-        val offenderNo = createActiveBooking(prisonId = "MDI").also { transferOut(it) }
+        val offenderNo = createActiveBooking(prisonId = "MDI").also { builderContext.transferOut(it) }
```
andymarke, line 118: minor issue; I called the object `builderContext`   - I think I will rename to `testDataContext`  or something since we are not building anything in this scenario just reading data

petergphillips (approved):
LGTM

PR #1355, opened 2022-07-05
SDI-259 Rename BuilderContext to TestDataContext given it used for building and querying data

andymarke:


petergphillips (approved):
LGTM

PR #1356, opened 2022-07-06
SDI-259 Always set the youth profile marker - N for adult - consistent with NOMIS

andymarke:
With this fixed there was no need for the data loader since the main "createBooking" will set this data correctly


lbennett-moj (approved):
LGTM

PR #1357, opened 2022-07-06
NN-3240 apply rules to csra latest used in profile endpoint

timoj58:
csra end point needs to behave in the same way prison profile does (link to code on jira ticket).  Therefore, its not simply a case of selecting the first item returned by the sql order, but based on last reviewed / calculated assuming no override present.


src/main/java/uk/gov/justice/hmpps/prison/service/InmateService.java
```diff
@@ -500,6 +508,16 @@ private boolean validCsra(final Assessment firstAssessment) {
...
+    private boolean isReviewedCsra(final AssessmentDto assessmentDto) {
+        return (assessmentDto.getReviewSupLevelType() != null && !"PEND".equals(assessmentDto.getReviewSupLevelTypeDesc()));
+    }
+
+    private boolean isCalculatedCsra(final AssessmentDto assessmentDto) {
+        return (assessmentDto.getCalcSupLevelType() != null && !"PEND".equals(assessmentDto.getCalcSupLevelTypeDesc()))
+            && assessmentDto.getOverridedSupLevelType() == null;
+    }
+
+
```
andymarke, line 520: I suppose if this Kotlin we would have added some extension functions to the AssessmentsDTO to improve readability with the not/and/negation above 

`fun AssessementDto.hasNonPendingReview() ` or whatever

andymarke (approved):
LGTM

PR #1358, opened 2022-07-07
SDI-259 Reuse TrustAccountService for new booking and recall

andymarke:


src/main/java/uk/gov/justice/hmpps/prison/service/transfer/TrustAccountService.kt
```diff
@@ -4,26 +4,31 @@ import org.slf4j.LoggerFactory
...
 import uk.gov.justice.hmpps.prison.repository.FinanceRepository
+import uk.gov.justice.hmpps.prison.repository.jpa.model.AgencyLocation
 import uk.gov.justice.hmpps.prison.repository.jpa.model.ExternalMovement
 import uk.gov.justice.hmpps.prison.repository.jpa.model.OffenderBooking
 
 interface TrustAccountService {
-  fun createTrustAccount(booking: OffenderBooking, movementOut: ExternalMovement, movementIn: ExternalMovement)
+  fun createTrustAccount(
+    booking: OffenderBooking,
+    fromAgency: AgencyLocation,
```
andymarke, line 14: pretty certain we don't need to pass in fromAgency since it is in the movement, but that will be a later PR

src/main/java/uk/gov/justice/hmpps/prison/service/transfer/PrisonTransferService.kt
```diff
@@ -117,7 +117,7 @@ class PrisonTransferService(
             OffenderProgramEndReason.TRF.code
           )
           trustAccountService.createTrustAccount(
-            booking = this, movementOut = toCourtMovement, movementIn = createdMovement
+            booking = this, fromAgency = toCourtMovement.fromAgency, movementIn = createdMovement
```
andymarke, line 120: switched to just passing in the fromAgency rather than whole movement since that is all service needs - see not below that that might not be necessary either ....

PR #1362, opened 2022-07-13
SDI-268: ⬆ Kotlin upgrade

petergphillips:


build.gradle.kts
```diff
@@ -14,28 +14,13 @@ configurations {
...
 
 // Temporarily keep at 2.5.1 until can switch to h2 instead (tests break anyway with 2.6.1)
 val hsqldbVersion by extra("2.5.1")
 
-// groovy errors with latest version
-val restAssuredVersion by extra("4.5.1")
-
-// Temporarily keep at 3.2.2 as seems to bring in groovy incompatibilites by upgrading to 3.2.4
-val serenityVersion by extra("3.2.2")
+ext["rest-assured.version"] = "5.1.1"
```
petergphillips, line 23: turns out this is pinned by spring boot to 4.5.1, so need to override in order to avoid dependency conflict

build.gradle.kts
```diff
@@ -52,8 +37,6 @@ dependencies {
 
   implementation("org.springframework.boot:spring-boot-starter-oauth2-resource-server")
 
-  implementation("org.hibernate:hibernate-core:$hibernateCoreVersion")
```
petergphillips, line 55: this is pulled in anyway so no need to specify

PR #1383, opened 2022-08-31
SDI-352 Example of how to set the Nomis context to ignore Xtag events

mikehalmamoj:
The idea is that when calling an endpoint that updates Nomis you can send in a header on the request (in this example "no-xtag-events" -> "true") to indicate we want to use the Nomis MERGE context instead of the PRISON_API context. This then triggers a bunch of defensive clauses in various XTAG generating stored procedures to prevent the XTAG event from being generated.

This has been tested with the endpoint `/offenders/{offenderNo}/discharge-to-hospital` and it did indeed prevent any XTAG events from being raised.

Note that to test this running locally you'll need to configure the replica database otherwise the OracleConnectionAspect is never fired.

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/web/filter/NomisContextFilter.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/web/filter/NomisContextFilter.java
@@ -1,0 +1,61 @@
+package uk.gov.justice.hmpps.prison.web.filter;
+
+import lombok.extern.slf4j.Slf4j;
+import org.slf4j.MDC;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+import javax.servlet.Filter;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/justice/hmpps/prison/web/filter/NomisContextFilter.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public void doFilter(
+            final ServletRequest request,
+            final ServletResponse response,
+            final FilterChain chain) throws IOException, ServletException {
+
+        MDC.put(NOMIS_CONTEXT, PRISON_API.name());
+        final var req = (HttpServletRequest) request;
+        if ("true".equals(req.getHeader("no-xtag-events"))) {
```
mikehalmamoj, line 51: I'm less than convinced by the header`no-xtag-events`. A good suggestion from @andrewrlee was `prevent-event-propagation` which doesn't leak implementation details.

PR #1394, opened 2022-09-20
Reverting change to AllFeatureTest which stopped some tests running.

ldlharper:
When adding some details to the sentence calculation endpoints I changed the configuration of the integration tests to allow me to debug locally. This was accidentally committed and merged in this PR.

https://github.com/ministryofjustice/prison-api/pull/1286

I have reverted this change in this PR. However this resulted in a few failing tests. I have narrowed down the cause of these failures. 

Firstly there were some failing tests in `features/users.feature`. This is because an endpoint was removed in the PR 
https://github.com/ministryofjustice/prison-api/pull/1334
I have removed these tests.

The only other failure was in `features/booking_details.feature`. I don't know what the cause of this is as I'm unfamiliar with this area of prison api, but I can see it was most likely introduced in https://github.com/ministryofjustice/prison-api/pull/1357

I've added a @broken annotation and will chase for someone to follow it up.


petergphillips (approved):
LGTM

PR #1395, opened 2022-09-20
SDI-388 wire up new xtag visit event to Offender events

lbennett-moj:


src/main/java/uk/gov/justice/hmpps/prison/service/transformers/OffenderEventsTransformer.java
```diff
@@ -964,6 +965,19 @@ private OffenderEvent iepUpdatedEventOf(final Xtag xtag) {
             .build();
     }
 
+    private OffenderEvent visitCancelledEventOf(final Xtag xtag) {
+        return OffenderEvent.builder()
+            .eventType("VISIT_CANCELLED")
```
petergphillips, line 970: why convert the update event into a cancelled event?  Will there ever be an update event instead?

PR #1396, opened 2022-09-21
NN-4508 remove prisoner from test as data is setup incorrectly

timoj58:
data setup for the prisoner A1234AD has the following
* calc level - LOW
* review level - null
* override level - MED

The new rules state that this will no longer be selected first, as it either needs a review level, or calculated as long as override is not present.

Given this data may be used elsewhere its easier to simply exclude it from the test set, given its now invalid.  Also removed from the non CSRA test as this uses the same step definition data.  The other tests provide enough coverage for the non CSRA one.


petergphillips (approved):
LGTM

PR #1403, opened 2022-10-03
SDI-266 create offender advanced validation integration tests

andymarke:
Require prior to refactoring


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceNewOffenderTest.kt
```diff
@@ -178,75 +287,154 @@ class OffenderResourceNewOffenderTest : ResourceTest() {
               .isEqualTo("Field: middleName2 - Value is too long: max length is 35")
           }
         }
-      }
 
-      @Nested
```
andymarke, line 183: some of these were moved, they were accidentally outside the main nested test

mikehalmamoj (approved):

PR #1407, opened 2022-10-05
See what it looks like without the nullable Result

mikehalmamoj:


src/main/java/uk/gov/justice/hmpps/prison/service/receiveandtransfer/PrisonerCreationService.kt
```diff
@@ -95,28 +95,28 @@ class PrisonerCreationService(
...
+  private fun suffix(code: String?): Result<Suffix?> = suffixRepository.getOptionalReference(code, Suffix.SUFFIX)
+
+  private fun ethnicity(code: String?): Result<Ethnicity?> = ethnicityRepository.getOptionalReference(code, Ethnicity.ETHNICITY)
+
+  private inline fun <reified T: ReferenceCode> ReferenceCodeRepository<T>.getMandatoryReference(code: String, pk: String): Result<T> =
+    findByIdOrNull(ReferenceCode.Pk(pk, code))
+      ?.let { success(it) }
+      ?: failure(EntityNotFoundException.withMessage("${T::class.java} $code not found"))
+
+  private inline fun <reified T: ReferenceCode> ReferenceCodeRepository<T>.getOptionalReference(code: String?, pk: String): Result<T?> =
```
mikehalmamoj, line 111: These extension functions could be pushed up to ReferenceCode to make them available everywhere. 

You could probably also do something clever with the pk and derive it from T. Maybe make ReferenceCode a sealed class and have a static function mapping type to PK. Or summat like that.

src/main/java/uk/gov/justice/hmpps/prison/service/receiveandtransfer/PrisonerCreationService.kt
```diff
@@ -95,28 +95,28 @@ class PrisonerCreationService(
...
+    code?.takeIf { it.isNotBlank() }
+      ?.let {
+        findByIdOrNull(ReferenceCode.Pk(pk, code))
+          ?.let { success(it) }
+          ?: failure(EntityNotFoundException.withMessage("${T::class.java} $code not found"))
+      }
+      ?: success(null)
```
mikehalmamoj, line 118: Just wanted to see what it looks like the other way around. I think I prefer this as it is explicitly says null is a success, and the extra question mark on the caller kept making me think twice. But it's 6 and two 3s really.

PR #1415, opened 2022-10-11
SDI-416 Switched primary db

andymarke:
Switched primary db (via non-readonly transaction) for 2 endpoints used by ES, this is a temporary fix for issue caused by replica being out of date when a xtag trigger fires.

Tested locally by upping logging level which shows the "master" database is selected with the annotation on


petergphillips (approved):
LGTM

PR #1416, opened 2022-10-12
SDI-266 Added integration test for recall prior to refactor

andymarke:


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceRecallTest.kt
```diff
@@ -0,0 +1,868 @@
...
+          .uri("/api/offenders/{offenderNo}", offenderNo)
+          .headers(
+            setAuthorisation(
+              listOf("ROLE_SYSTEM_USER")
+            )
+          )
+          .header("Content-Type", MediaType.APPLICATION_JSON_VALUE)
+          .accept(MediaType.APPLICATION_JSON)
+          .exchange()
+          .expectStatus().isOk
```
mikehalmamoj, line 104: Should this assert the result is empty?

andymarke: I guess you mean assert that the test offender we just created has no bookings? If so, yes, sounds like a good idea. Next PR

mikehalmamoj (approved):

PR #1418, opened 2022-10-13
MDWFP-23 adding some endpoints to save digital warrant data to NOMIS.…

ldlharper:
… Not production code, can only be run with a role not given to real users.


src/main/java/uk/gov/justice/hmpps/prison/api/resource/DigitalWarrantResource.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @ApiResponses({
+        @ApiResponse(responseCode = "201", description = "Court case created.", content = {@Content(mediaType = "application/json", schema = @Schema(implementation = Long.class))}),
+        @ApiResponse(responseCode = "404", description = "Requested resource not found.", content = {@Content(mediaType = "application/json", schema = @Schema(implementation = ErrorResponse.class))})})
+    @Operation(summary = "Create a court case")
+    @PostMapping("/booking/{bookingId}/court-case")
+    @PreAuthorize("hasRole('MANAGE_DIGITAL_WARRANT') and hasAuthority('SCOPE_write')")
+    public ResponseEntity<Long> createCourtCase(@PathVariable("bookingId") @Parameter(description = "The required booking id (mandatory)", required = true) final Long bookingId,
+                                               @RequestBody final CourtCase courtCase) {
+        return ResponseEntity.status(HttpStatus.CREATED).body(digitalWarrantService.createCourtCase(bookingId, courtCase));
```
mikehalmamoj, line 50: I think you get a ResponseEntity by default so you don't have to be explicit... though the default http status is 200 and is overriden with `@ResponseStatus`.

src/main/java/uk/gov/justice/hmpps/prison/service/digitalwarrant/DigitalWarrantService.java
```diff
@@ -0,0 +1,144 @@
...
+
+    @Transactional
+    public Integer createOffenderSentence(Long bookingId, Sentence sentence) {
+        OffenderCourtCase courtCase = offenderCourtCaseRepository.findById(sentence.getCourtCaseId()).orElseThrow(EntityNotFoundException.withId(sentence.getCourtCaseId()));
+        OffenderBooking booking = offenderBookingRepository.findByBookingId(bookingId).orElseThrow(EntityNotFoundException.withId(bookingId));
+        OffenderCharge offenderCharge = offenderChargeRepository.findById(sentence.getOffenderChargeId()).orElseThrow(EntityNotFoundException.withId(sentence.getOffenderChargeId()));
+        SentenceCalcType sentenceCalcType = sentenceCalcTypeRepository.findById(new SentenceCalcType.PK(sentence.getSentenceType(), sentence.getSentenceCategory())).orElseThrow(EntityNotFoundException.withId(sentence.getSentenceType() + " " + sentence.getSentenceCategory()));
+
+        //TODO is this the correct way to set sequence?
+        Integer sequence = offenderSentenceRepository.findByOffenderBooking_BookingId(bookingId).stream().max(Comparator.comparing(OffenderSentence::getSequence)).map(os -> os.getSequence() + 1).orElse(1);
```
mikehalmamoj, line 112: I can't find the source code for this table so not sure if this is right or not. Would probably be quicker with a dedicated query for max sequence? Which would reduce the chance of a race condition actually happening - though it's probably already very very small. Anyway, worth revisiting post POC.

mikehalmamoj (approved):

src/main/java/uk/gov/justice/hmpps/prison/api/resource/DigitalWarrantResource.java
ldlharper: Yea I added the ResponseEntity to use the correct http status. I didn't know about @ResponseStatus, I'll change it thanks.

src/main/java/uk/gov/justice/hmpps/prison/service/digitalwarrant/DigitalWarrantService.java
ldlharper: Yea I think that would be optimised. This is working for now with an integration test. I wanted to just try a POC before digging into how its done in the NOMIS DB. Its very much work in progress!

I've got a ticket for improvements so I'll add this suggestion.

src/main/java/uk/gov/justice/hmpps/prison/service/digitalwarrant/DigitalWarrantService.java
```diff
@@ -0,0 +1,144 @@
...
+    private ReferenceCodeRepository<LegalCaseType> legalCaseTypeReferenceCodeRepository;
+    @Autowired
+    private OffenceResultRepository offenceResultRepository;
+
+    @Autowired
+    private SentenceTermRepository sentenceTermRepository;
+
+    @Transactional
+    public Long createCourtCase(Long bookingId, CourtCase courtCase) {
+        AgencyLocation agency = agencyLocationRepository.findById(courtCase.getAgencyId()).orElseThrow(EntityNotFoundException.withId(courtCase.getAgencyId()));
```
andymarke, line 66: Vey minor style observation: I think we would prefer uses of `var`

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/DigitalWarrantResourceImplIntTest.java
```diff
@@ -0,0 +1,239 @@
...
+import java.time.LocalDate;
+import java.util.List;
+import java.util.Objects;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+public class DigitalWarrantResourceImplIntTest extends ResourceTest {
+    @Autowired
+    private OffenderCourtCaseRepository offenderCourtCaseRepository;
```
andymarke, line 28: Minor observations; we would prefer to use webTestClient going forward for new test. Also if possible try creating a offender / booking "on the fly" rather than using seed data. Tests tend to read better that way

andymarke (approved):
lgtm

src/main/java/uk/gov/justice/hmpps/prison/service/digitalwarrant/DigitalWarrantService.java
ldlharper: Thanks I'll make a note of this on the improvement ticket.

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/DigitalWarrantResourceImplIntTest.java
ldlharper: Thanks I'll make a note of this on the improvement ticket.

PR #1421, opened 2022-10-14
INC-865 Return Default level for IEP from Prison API endpoint

Mjwillis:
- INC-865 Return Default level for IEP from Prison API endpoint


src/main/java/uk/gov/justice/hmpps/prison/api/model/IepLevel.java
```diff
@@ -26,4 +26,7 @@ public class IepLevel {
 
     @Schema(description = "Sequence Order of IEP", example = "1")
     private Integer sequence;
+
+    @Schema(description = "Indicates this is the default level for this prison", example = "false", type = "Boolean")
+    private boolean defaultLevel;
```
xoen, line 31: Hi, would be possible to name this field `default` please 🙏 ☺️

Mjwillis: no

xoen: Looks like `default` is reserved word so it's not possible, ignore me :)

andymarke (approved):
LGTM

PR #1427, opened 2022-10-21
SDI-266 Transfer by TAP should copy escort code to new movement

andymarke:


src/main/java/uk/gov/justice/hmpps/prison/service/receiveandtransfer/ExternalMovementTransferService.kt
```diff
@@ -173,6 +176,7 @@ class ExternalMovementTransferService(
           /* toAgency = */ lastMovement.fromAgency,
           /* active = */ true,
           /* escortText = */ null,
+          /* escortCode = */ lastMovement.escortCode,
           /* commentText = */ commentText,
```
andymarke, line 180: only copy for TAP returns back to same prison

lbennett-moj (approved):

PR #1433, opened 2022-10-25
SDI-429 discharge case note

mikehalmamoj:
When discharging to hospital via Restricted Patients we change the case note text to say "Transfer" instead of "Release" because the offender is still under prison care. Release is generally interpreted as released into the community.


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceIntTest.java
```diff
@@ -636,6 +655,103 @@ public void testCanMovePrisonerFromPrisonToHospital() {
...
+            "/api/bookings/{bookingId}/movement/{sequenceNumber}",
+            GET,
+            createHttpEntity(token, null),
+            new ParameterizedTypeReference<Movement>() {
+            },
+            bookingId, 2
+        );
+
+        assertThat(lastMovement.getBody().getFromAgency()).isEqualTo("SYI");
+        assertThat(lastMovement.getBody().getToAgency()).isEqualTo("OUT");
```
mikehalmamoj, line 736: The key part of this test is that it is not possible to enter a hospital if releasing in NOMIS so the movement's to location is out. Therefore we don't want the updated text on the case note as we don't know which hospital the offender was moved to.

andymarke (approved):
LGTM

PR #1436, opened 2022-10-27
NN-4465 filter on scheduled only

timoj58:
as per https://api-dev.prison.service.justice.gov.uk/swagger-ui/index.html#/schedules/getEventsByLocationId both endpoints used by whereabouts.


neilmendum (approved):
LGTM

PR #1438, opened 2022-10-27
SDI-429 Set the audit module regardless of user

mikehalmamoj:
We've seen cases where the AUDIT_MODULE_NAME in the context appears to be corrupt - we're still producing events from NOMIS though we think the MERGE context has been set.

This closes a loophole where the AUDIT_MODULE_NAME is not set and might fix the issue.


src/test/java/uk/gov/justice/hmpps/prison/aop/connectionproxy/OracleConnectionAspectTest.java
```diff
@@ -33,56 +40,105 @@ class OracleConnectionAspectTest {
...
-        assertThat(sqlCaptor.getAllValues().get(0)).contains("ALTER SESSION SET CURRENT_SCHEMA");
+                ArgumentCaptor<String> sqlCaptor = ArgumentCaptor.forClass(String.class);
+                verify(pooledConnection, times(2)).prepareStatement(sqlCaptor.capture());
+                assertThat(sqlCaptor.getAllValues().get(0)).contains("nomis_context.set_client_nomis_context");
+                assertThat(sqlCaptor.getAllValues().get(1)).contains("ALTER SESSION SET CURRENT_SCHEMA");
+            }
+        }
+
+        @Nested
+        class NotNomisOrProxyUser {
```
mikehalmamoj, line 111: GitHub struggled with the diff for the unit tests 😞 But essentially the new tests are in the class.

steverendell (approved):
lgtm

src/main/java/uk/gov/justice/hmpps/prison/aop/connectionproxy/OracleConnectionAspect.java
```diff
@@ -50,6 +50,8 @@ protected Connection openProxySessionIfIdentifiedAuthentication(final Connection
             } catch (SQLException e) {
                 log.warn("Failed to set context for proxy user {}", MDC.get(PROXY_USER));
             }
+        } else {
+            setContextAuditModuleOnly(pooledConnection);
```
Mjwillis, line 54: think this will need a try/catch to handle the read replica

Mjwillis: @mikehalmamoj this might error in production

PR #1452, opened 2022-11-02
SDI-429 Don't clear the session - it's then broken forever

mikehalmamoj:
We think we've found a bug where clearing the session when closing a Proxy connection then breaks the session context thereafter. Any further calls to set_context don't work.

This needs fixing properly but in the meantime remove the clear session recently added to all update connections which can only make things worse.


andymarke (approved):
lgtm

PR #1453, opened 2022-11-04
MDWFP-23 imprisonment status and serious offence flag.

ldlharper:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/Offence.java
```diff
@@ -87,4 +90,9 @@ public static class PK implements Serializable {
     @BatchSize(size = 25)
     private List<OffenceIndicator> offenceIndicators = new ArrayList<>();
 
+    public boolean isMoreSeriousThan(Offence other) {
+        return isNumeric(this.getSeverityRanking()) && isNumeric(other.getSeverityRanking()) &&
```
mikehalmamoj, line 94: Why would the severity ranking not be numeric? Just wondering if this could cause problems as `offence1.isMoreSeriousThan(offence2) == false` and `offence2.isMoreSeriousThan(offence1) == false`.

src/main/java/uk/gov/justice/hmpps/prison/service/digitalwarrant/DigitalWarrantService.java
```diff
@@ -160,7 +168,29 @@ public Integer createOffenderSentence(Long bookingId, Sentence sentence) {
...
-        var courtEvent = courtCase.getCourtEvents().get(0);
+
+        var movementReason = movementReasonReferenceCodeRepository.findById(MovementReason.SENTENCING).orElseThrow(EntityNotFoundException.withIdAndClass(MovementReason.SENTENCING.getCode(), MovementReason.class));
+        var eventStatus = eventStatusReferenceCodeRepository.findById(EventStatus.COMPLETED).orElseThrow(EntityNotFoundException.withIdAndClass(EventStatus.COMPLETED.getCode(), EventStatus.class));
+        var result = offenceResultRepository.findById(OffenceResultRepository.IMPRISONMENT).orElseThrow(EntityNotFoundException.withIdAndClass(OffenceResultRepository.IMPRISONMENT, OffenceResult.class));
+        var courtEvent = CourtEvent.builder()
+            .offenderBooking(booking)
+            .courtEventType(movementReason)
+            .eventStatus(eventStatus)
+            .startTime(sentence.getSentenceDate().atTime(10, 0))
```
mikehalmamoj, line 179: 10 seems a bit random 😄

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/Offence.java
ldlharper: So as far as i can see every entry in the dev database is a number. The column is defined as a VARCHAR though. The only way you could have false being returned in both cases as above would be if the column was NULL. I didn't want to change the definition in JPA to be an Integer incase there was something I was missing.

src/main/java/uk/gov/justice/hmpps/prison/service/digitalwarrant/DigitalWarrantService.java
ldlharper: Yea! I'm finding things that are missing from the frontend so this will need to be added to the API at some point. I'm keeping track of the things missing from the first UI design.

PR #1455, opened 2022-11-08
SDI-433 Improve test coverage of Nomis connections

mikehalmamoj:
We currently support 4 different ways of connecting to NOMIS - NOMIS proxy user, other proxy user, anonynous user suppressing XTag events, anonymous user. The tests around `OracleConnectionAspect` have been improved to capture the various side effects for all of these types of user to give confidence before refactoring this class.


steverendell (approved):
lgtm

PR #1460, opened 2022-11-10
MDWFP-23 Adjustments endpoint

ldlharper:


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/DigitalWarrantResourceImplIntTest.java
```diff
@@ -164,6 +178,77 @@ public void sentence_success() {
...
+            Adjustment.builder()
+                .days(10)
+                .from(LocalDate.of(2022, 10, 1))
+                .to(LocalDate.of(2022, 11, 1))
+                .days(30)
+                .sequence(1)
+                .type(SentenceAdjustmentType.REMAND.name())
+                .build()
+        );
+
```
andymarke, line 197: very minor, but going forward webTestClient is preferred - hopefully one day we can get rid of the rest template

andymarke (approved):
LGTM

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/DigitalWarrantResourceImplIntTest.java
ldlharper: Okay I'll look at replacing that. I think all this code will be removed from prison-api at some point, but I'm just testing how this data looks and functions in the NOMIS UI

PR #1461, opened 2022-11-15
SDI-430 Add @SlowReportQuery annotation to prison-api

steverendell:
Remove annotation for endpoints relied on by event message clients


petergphillips (approved):
LGTM

PR #1466, opened 2022-11-22
SDI-444: ✨ Add in new endpoint to return split of housing locations

petergphillips:
- SDI-444: ✨ Add in living unit reference code
- SDI-444: ✨ Add in new endpoint to return split of housing locations


lbennett-moj (approved):
LGTM

PR #1471, opened 2022-11-25
SDI-464 Tidy up logging around NOMIS connections

mikehalmamoj:
We know that the NOMIS connections are being configured correctly now so we can remove some of the extra logging. I think all we need to know is:
* when a proxy user is used
* when a special connection is opened
* when a special connection is closed

We can assume that everything works as any exceptions will be logged in App Insights as will the failing request.

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/aop/connectionproxy/AbstractConnectionAspect.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/aop/connectionproxy/AbstractConnectionAspect.java
@@ -25,7 +25,7 @@ public abstract class AbstractConnectionAspect {
         }
         final var pooledConnection = (Connection) joinPoint.proceed();
         try {
-            final var connectionToReturn = openProxySessionIfIdentifiedAuthentication(pooledConnection);
+            final var connectionToReturn = configureNomisConnection(pooledConnection);
 
             if (log.isTraceEnabled() && MdcUtility.isLoggingAllowed()) {
                 log.trace(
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/justice/hmpps/prison/aop/connectionproxy/AbstractConnectionAspect.java
```diff
@@ -50,5 +50,5 @@ public Object connectionAround(final ProceedingJoinPoint joinPoint) throws Throw
         }
     }
 
-    protected abstract Connection openProxySessionIfIdentifiedAuthentication(final Connection pooledConnection) throws SQLException;
+    protected abstract Connection configureNomisConnection(final Connection pooledConnection) throws SQLException;
```
mikehalmamoj, line 53: The method does more now so the old name isn't correct.

src/main/java/uk/gov/justice/hmpps/prison/aop/connectionproxy/OracleConnectionAspect.kt
```diff
@@ -40,7 +40,7 @@ class OracleConnectionAspect(
 
   @Throws(SQLException::class)
   private fun Connection.openProxySessionConnection(): Connection {
-    log.info("Configuring Oracle Proxy Session for NOMIS user {}", this)
```
mikehalmamoj, line 43: This was actually logging the object id of the `Connection` instance - no value that I can see, we just want to know which kind of connection this request is using.

andymarke (approved):
LGTM

PR #1484, opened 2022-12-22
INC-993 Remove IEP endpoints and generation on receive and transfer

Mjwillis:


src/main/java/uk/gov/justice/hmpps/prison/service/InmateService.java
```diff
@@ -240,7 +244,7 @@ private InmateDetail getOffenderDetails(final InmateDetail inmate, final boolean
             }
             if (extraInfo) {
                 inmate.setAliases(repository.findInmateAliases(bookingId, "createDate", Order.ASC, 0, 100).getItems());
-                inmate.setPrivilegeSummary(bookingService.getBookingIEPSummary(bookingId, false));
```
Mjwillis, line 243: Like to remove this code next

Mjwillis: Don't think anyone looks at IEP through this endpoint but need to check

src/main/java/uk/gov/justice/hmpps/prison/service/InmateService.java
```diff
@@ -240,7 +244,7 @@ private InmateDetail getOffenderDetails(final InmateDetail inmate, final boolean
...
+                inmate.setPrivilegeSummary(getBookingIEPSummary(bookingId));
```
andymarke, line 247: do you know who uses this?

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/service/InmateService.java
Mjwillis: Don't think anyone does

PR #1486, opened 2022-12-22
INC-1009 Remove all references in Prison API to Prisoners Incentives (except V1 apis)

Mjwillis:
Blocked - see slack thread https://mojdt.slack.com/archives/C69NWE339/p1671793804497189


andymarke (approved):
LGTM, but how can you know nobody is using this data?

Mjwillis:
> LGTM, but how can you know nobody is using this data?

@andymarke the only endpoint where the data is still surfaced is in the `/offenders/{noms}` endpoint.  It would mean that systems would be reading the IEP from Prison API still - looking at github there are some mappings in other projects so I'll hold off on this.  I might actually be they are not actually using the field.  See https://github.com/search?p=4&q=org%3Aministryofjustice+privilegeSummary&type=Code

andymarke (approved):
Lgtm

PR #1490, opened 2023-01-10
SDI-559 Ensure ES reindex order is by new offender records

andymarke:


src/main/java/uk/gov/justice/hmpps/prison/repository/sql/PrisonerRepositorySql.kt
```diff
@@ -48,7 +48,7 @@ enum class PrisonerRepositorySql(val sql: String) {
 
   LIST_ALL_OFFENDERS(
     """
-        SELECT OFFENDER_ID_DISPLAY AS OFFENDER_NUMBER, max(OFFENDER_ID) AS mo FROM OFFENDERS GROUP BY OFFENDER_ID_DISPLAY ORDER BY mo ASC
+        SELECT OFFENDER_ID_DISPLAY AS OFFENDER_NUMBER, max(ROOT_OFFENDER_ID) AS mo FROM OFFENDERS GROUP BY OFFENDER_ID_DISPLAY ORDER BY mo ASC
     """
```
andymarke, line 52: new offenders will always be at the end of the list rather then new offenders and ones where an alias was created on an old offender

lbennett-moj (approved):

PR #1492, opened 2023-01-16
NN-4750 oic charge id required for scheduling hearings

timoj58:
converted to draft as should not be accepted before Wednesday


neilmendum (approved):
LGTM

PR #1495, opened 2023-01-24
MDWFP-29 Adding endpoint to get court events dates.

ldlharper:
Adding an endpoint to retrieve all court dates and result codes for a given prisoner. This data is required for the relevant remand calculation.


src/main/java/uk/gov/justice/hmpps/prison/service/digitalwarrant/DigitalWarrantService.kt
```diff
@@ -221,4 +222,30 @@ class DigitalWarrantService(
       ).id
     }
   }
+  @Transactional
```
steverendell, line 225: The convention is to mark the service class as @Transactional(readonly=true) and override update/create functions with @Transactional

src/main/java/uk/gov/justice/hmpps/prison/api/resource/DigitalWarrantResource.java
```diff
@@ -81,4 +85,15 @@ public Long createAdjustment(@PathVariable("bookingId") @Parameter(description =
                                   @RequestBody final Adjustment adjustment) {
         return digitalWarrantService.createAdjustment(bookingId, adjustment);
     }
+
+    @ApiResponses({
+        @ApiResponse(responseCode = "200", description = "The court date results.", content = {@Content(mediaType = "application/json", schema = @Schema(implementation = Integer.class))}),
```
steverendell, line 90: this class doent look right

ldlharper: Done thanks.

steverendell (approved):
lgtm

PR #1497, opened 2023-01-27
SDI-596: 🐛 Always try to set a prison entry date for timeline

petergphillips:


src/test/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderBookingTest.kt
```diff
@@ -232,255 +225,117 @@ class OffenderBookingTest {
         .build().also {
           offender.addBooking(it)
 
+          // booking has a single release - no admission
           it.addExternalMovement(
             ExternalMovement.builder()
-              .movementType(MovementType("ADM", "Admission"))
-              .movementDirection(MovementDirection.IN)
-              .movementReason(MovementReason("B", "Recall"))
-              .movementTime(LocalDateTime.of(2019, 1, 4, 9, 30))
```
mikehalmamoj, line 240: Bit confused why the old test was called 'cope with bad data and no admission' when it has an admission 🤷

mikehalmamoj (approved):

src/test/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderBookingTest.kt
petergphillips: in the previous PR I converted the code from java to kotlin and accidentally added the two test methods by copying them from the previous test and changing their name

PR #1500, opened 2023-02-03
MDWFP-37 adding disposition code to court results.

ldlharper:
Adding disposition code to court results. This will be used to calculate relevant remand for non-sentencing outcomes.


andymarke (approved):
LGTM

PR #1503, opened 2023-02-14
MDWFP-53 fixing clashing names with digital warrant api models.

ldlharper:
Renaming warrant folder api classes to avoid documentation clashes.


andrewrlee (approved):
LGTM

PR #1504, opened 2023-02-15
INC-805 :sparkles: Behaviour since last review case note endpoint support

Mjwillis:
What does this do?:
- provides an endpoint which can count the occurrences of different case note types for a list of prisoners (booking Ids) but with different from dates.

Why?:
Supports the user need: How many positive and negative case note (behaviour) have been raised since a specific date - in this use case its the time since the last incentive review occurred.


src/main/java/uk/gov/justice/hmpps/prison/repository/sql/CaseNoteRepositorySql.kt
```diff
@@ -35,6 +35,21 @@ enum class CaseNoteRepositorySql(val sql: String) {
...
 
+  GROUP_BY_TYPES_FOR_BOOKING_FROM_DATE(
+    """
+        SELECT CASE_NOTE_TYPE,
+        CASE_NOTE_SUB_TYPE,
+        OFFENDER_BOOK_ID  BOOKING_ID,
+        COUNT(*)          NUM_CASE_NOTES,
+        MAX(CONTACT_TIME) LATEST_CASE_NOTE
+        FROM OFFENDER_CASE_NOTES OCS
+        WHERE AUDIT_TIMESTAMP >= :fromDate
```
mikehalmamoj, line 46: Are you sure AUDIT_TIMESTAMP is the correct date for this query? Not CREATE_DATETIME?

mikehalmamoj (approved):

src/main/java/uk/gov/justice/hmpps/prison/repository/sql/CaseNoteRepositorySql.kt
Mjwillis: Yes its the one with the index on it

PR #1513, opened 2023-02-22
Adding a test to detect duplicate OpenAPI schemas

andrewrlee:
Duplicate schemas do not render in the swagger docs - only one version will be displayed everywhere

A couple of duplicates were fixed a couple of weeks ago and this changes another and ignores an instance that affects display of the V1 APIs

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/api/model/adjudications/AdjudicationDetail.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/api/model/adjudications/AdjudicationDetail.java
@@ -12,7 +12,7 @@ import java.util.List;
 
 import static com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL;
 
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/justice/hmpps/prison/api/model/adjudications/AdjudicationDetail.java
```diff
@@ -12,7 +12,7 @@
 
 import static com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL;
 
-@Schema(description = "Detail about an individual Adjudication")
+@Schema(name = "IndividualAnnotation", description = "Detail about an individual Adjudication")
```
petergphillips, line 15: should this be `IndividualAdjudication` instead?

petergphillips (approved):
LGTM

PR #1516, opened 2023-02-24
MOP-43 offences.ho_code is not a mandatory field on the Offences table - correct the entity definition

yasinmustafa:


petergphillips (approved):
LGTM

PR #1518, opened 2023-02-28
SAA-406: Add endpoint for bulk lookup of education levels

smcveigh941:


mikehalmamoj (approved):

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderEducationService.java
```diff
@@ -24,4 +27,12 @@ public class OffenderEducationService {
     public Page<Education> getOffenderEducations(@NotNull final String nomisId, final PageRequest pageRequest) {
         return repository.findAllByNomisId(nomisId, pageRequest).map(transformer::convert);
     }
+
+    @VerifyOffenderAccess(overrideRoles = {"SYSTEM_USER", "GLOBAL_SEARCH", "VIEW_PRISONER_DATA"})
+    public List<Education> getOffenderEducations(@NotNull final Set<String> nomisIds) {
+        return repository.findAllByNomisIdIn(nomisIds)
```
petergphillips, line 33: this should really be batched otherwise you will get an SQL error if you call oracle with too many parameters.  See https://github.com/ministryofjustice/prison-api/blob/b6f55c3ffe3f872abd0c70d4d86638b4a34eab38/src/main/java/uk/gov/justice/hmpps/prison/service/AdjudicationsService.java#L252 for an example.

PR #1519, opened 2023-02-28
SDI-598: 🐛 Fix caching of null items after upgrade to spring boot 3

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/service/ReferenceDomainService.java
```diff
@@ -132,11 +132,13 @@ public Optional<ReferenceCode> getReferenceCodeByDomainAndCode(final String doma
     }
 
     private void verifyReferenceDomain(final String domain) {
+        if (StringUtils.isBlank(domain)) throw new EntityNotFoundException("Reference domain not specified");
```
petergphillips, line 135: decided it was safer to throw the same exception as thrown later in code rather than a different one

src/main/java/uk/gov/justice/hmpps/prison/service/AgencyService.java
```diff
@@ -181,6 +181,7 @@ public Page<Agency> getAgencies(final long offset, final long limit) {
     }
 
     public List<Agency> findAgenciesByUsername(final String username) {
+        if (StringUtils.isBlank(username)) return Collections.emptyList();
```
petergphillips, line 184: this was coming through as `null` under some circumstances causing
```
Invalid key type, expected : java.lang.String but was : org.springframework.cache.interceptor.SimpleKey
```
exception in the logs

mikehalmamoj (approved):

PR #1522, opened 2023-03-03
SDIT-513: 🐛 Remove old javax libs

petergphillips:


petergphillips:
we were getting an odd
> #14 299.7 > Entry BOOT-INF/lib/jaxb-core-4.0.2.jar is a duplicate but no duplicate handling strategy has been set. Please refer to https://docs.gradle.org/7.6/dsl/org.gradle.api.tasks.Copy.html#org.gradle.api.tasks.Copy:duplicatesStrategy for details.
error during building the boot jar, this will fix it

build.gradle.kts
```diff
@@ -42,12 +42,6 @@ dependencies {
 
   implementation("org.springframework.boot:spring-boot-starter-oauth2-resource-server")
 
-  implementation("javax.annotation:javax.annotation-api:1.3.2")
-  implementation("javax.xml.bind:jaxb-api:2.3.1")
```
petergphillips, line 46: not sure how I didn't notice these old javax libraries still hanging around, but not needed

src/main/java/uk/gov/justice/hmpps/prison/service/v1/NomisApiV1Service.java
```diff
@@ -258,7 +258,7 @@ private String convertToJsonString(final EventSP e) {
     public Image getOffenderImage(final String nomsId) {
         final var imageBytes = offenderV1Repository.getPhoto(nomsId).orElseThrow(EntityNotFoundException.withId(nomsId));
 
-        return Image.builder().image(DatatypeConverter.printBase64Binary(imageBytes)).build();
+        return Image.builder().image(Base64.getEncoder().encodeToString(imageBytes)).build();
```
petergphillips, line 261: googling found https://stackoverflow.com/questions/38563609/datatypeconverter-vs-base64 and to use `Base64` instead

mikehalmamoj (approved):

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/ExternalMovement.java
```diff
@@ -12,7 +12,6 @@
 import org.hibernate.type.YesNoConverter;
 import jakarta.persistence.Convert;
 
-import javax.annotation.Nullable;
```
petergphillips, line 15: this is an odd one.  `@Nullable` is a suggestion and isn't enforced.  However it is used by the kotlin null checker and an exception is thrown if a field is not marked as `Nullable`, so the tests then fail.  Reading https://youtrack.jetbrains.com/issue/KT-54205/Support-jakarta-Nullability-annotations aren't intending to support the jakarta annotation, but do currently support the jetbrains one so I've switched to that one instead.  Looks like https://jspecify.dev/ hasn't got off the ground yet so can't switch to that one either.

steverendell (approved):
lgtm

PR #1524, opened 2023-03-08
MDWFP-84 filtering out bad data with the court event charges table.

ldlharper:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/CourtEventChargeRepository.java
```diff
@@ -14,7 +14,8 @@ public interface CourtEventChargeRepository extends CrudRepository<CourtEventCha
         "join cec.eventAndCharge.offenderCharge oc " +
         "join oc.offenderBooking ob " +
         "join ob.offender o " +
-        "where o.nomsId = :offenderId"
+        "where o.nomsId = :offenderId and " +
```
petergphillips, line 17: `and` should really be on the same line as `exists`

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/CourtEventChargeRepository.java
```diff
@@ -14,7 +14,8 @@ public interface CourtEventChargeRepository extends CrudRepository<CourtEventCha
...
+        "exists (select id from CourtEvent where id = cec.eventAndCharge.courtEvent.id)"
```
petergphillips, line 18: Would be good to have a comment here to explain what is going on - as someone in future will just look at the line and think it isn't doing anything since the court event must exist.
Is there an existing repository test for this class - would be good to have a test for this too...

ldlharper: I'll add a comment here. 
I looked at adding a test but it was difficult to create the invalid data because the constraint exists in the schema we run tests against.

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/CourtEventChargeRepository.java
ldlharper: Done

andymarke (approved):
lgtm

PR #1526, opened 2023-03-09
SDIT-645: 🐛 Attempt to prevent duplicate bookings by grabbing lock on select

petergphillips:

PR #1531, opened 2023-03-10
SDIT-645: 🐛 Use selected bookings rather than old bookings on offender

petergphillips:
- Revert "Revert "SDIT-645: 🐛 Attempt to prevent duplicate bookings by grabbing lock on select (#1526)" (#1528)"
- SDIT-645: 🐛 Use selected bookings rather than old bookings on offender

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/service/receiveandtransfer/BookingIntoPrisonService.kt
+++ b/src/main/java/uk/gov/justice/hmpps/prison/service/receiveandtransfer/BookingIntoPrisonService.kt
@@ -76,6 +76,9 @@ class BookingIntoPrisonService(
     return newBooking(offender(prisonerIdentifier).getOrThrow(), requestForNewBooking)
   }
   fun newBooking(offender: Offender, requestForNewBooking: RequestForNewBooking): InmateDetail {
+    // ensure that we can get a select for update on the bookings before we start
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/justice/hmpps/prison/service/receiveandtransfer/BookingIntoPrisonService.kt
```diff
@@ -76,7 +76,10 @@ class BookingIntoPrisonService(
     return newBooking(offender(prisonerIdentifier).getOrThrow(), requestForNewBooking)
   }
   fun newBooking(offender: Offender, requestForNewBooking: RequestForNewBooking): InmateDetail {
-    val previousBooking: OffenderBooking? = previousInactiveBooking(offender).getOrThrow()
+    // ensure that we can get a select for update on the bookings before we start
+    val bookings = offenderBookingRepository.findAllByOffenderNomsIdForUpdate(offender.nomsId)
+
+    val previousBooking: OffenderBooking? = previousInactiveBooking(bookings).getOrThrow()
```
petergphillips, line 82: This time I use the bookings that we have just selected in order to find if any are inactive.  What I think was happening previously was that it was grabbing the offender (and thus bookings) and then hanging waiting for the lock on the bookings.  When it then checked for previous inactive bookings it used the offender.bookings rather than the newly selected bookings and thus used stale data.

mikehalmamoj (approved):

PR #1534, opened 2023-03-14
SDIT-645: 🐛 Lock on offender instead

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderBookingRepository.java
```diff
@@ -18,10 +14,6 @@ public interface OffenderBookingRepository extends
     JpaSpecificationExecutor<OffenderBooking>,
     CrudRepository<OffenderBooking, Long> {
 
-    @Lock(LockModeType.PESSIMISTIC_WRITE)
-    @Query("SELECT ob from OffenderBooking ob where ob.offender.nomsId = :nomsId")
```
petergphillips, line 22: locking moved from bookings to offender instead

src/main/java/uk/gov/justice/hmpps/prison/service/receiveandtransfer/BookingIntoPrisonService.kt
```diff
@@ -73,13 +72,14 @@ class BookingIntoPrisonService(
...
   fun newBooking(prisonerIdentifier: String, requestForNewBooking: RequestForNewBooking): InmateDetail {
-    return newBooking(offender(prisonerIdentifier).getOrThrow(), requestForNewBooking)
+    // grab a lock on the offender and their bookings to ensure that no-one else can add a booking to the offender
+    // during our transaction.  Other processes / threads will hang waiting for this to complete.
+    val offender = offenderForUpdate(prisonerIdentifier).getOrThrow()
+    return newBookingWithoutUpdateLock(offender, requestForNewBooking)
   }
-  fun newBooking(offender: Offender, requestForNewBooking: RequestForNewBooking): InmateDetail {
-    // ensure that we can get a select for update on the bookings before we start
-    val bookings = offenderBookingRepository.findAllByOffenderNomsIdForUpdate(offender.nomsId)
```
petergphillips, line 80: by grabbing the bookings separately we ended up with two different collections of bookings - one on the offender and then one here.  This meant that when hibernate the saved it saved both and one ended up overwriting the other one.  By switching to using and locking the offender we only end up with one list of bookings

src/main/java/uk/gov/justice/hmpps/prison/service/receiveandtransfer/BookingIntoPrisonService.kt
```diff
@@ -73,13 +72,14 @@ class BookingIntoPrisonService(
...
 
-    val previousBooking: OffenderBooking? = previousInactiveBooking(bookings).getOrThrow()
+  fun newBookingWithoutUpdateLock(offender: Offender, requestForNewBooking: RequestForNewBooking): InmateDetail {
```
petergphillips, line 81: renamed to make it clear that we don't lock in this method.

mikehalmamoj (approved):

PR #1550, opened 2023-03-31
MDWFP-73 adding sentence date to court results in order to calc relev…

ldlharper:
…ant remand when a sentence is intersecting.


steverendell (approved):
lgtm

PR #1555, opened 2023-04-06
SDIT-734: ✨ Also return previous permanent cell location

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/service/OffenderLocationService.kt
```diff
@@ -4,29 +4,54 @@ import com.fasterxml.jackson.annotation.JsonIgnore
...
-    val location = agencyInternalLocationRepository.findOneByLocationId(summary.internalLocationId.toLong())
+    val currentLocationId = summary.internalLocationId.toLong()
+    val location = agencyInternalLocationRepository.findOneByLocationId(currentLocationId)
       .orElseThrow { RuntimeException() }
 
-    return OffenderLocation(levels = location.getLocationWithParents())
+    val lastPermanentLocation = if (isTemporaryLocation(location.locationCode)) {
+      val cells = bedAssignmentHistoriesRepository.findAllByBedAssignmentHistoryPKOffenderBookingId(
+        bookingId,
+        PageRequest.of(0, 10, Sort.by("bedAssignmentHistoryPK.sequence").descending()),
```
petergphillips, line 36: we assume that they won't have moved between 10 different temporary locations so only get the last 10

AledWEvans: That's the same logic I used also,.. still a little uneasy as it's an arbitrary number.. But I do think this can be changed at a later date if need be.

mikehalmamoj (approved):

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderLocationService.kt
```diff
@@ -4,29 +4,54 @@ import com.fasterxml.jackson.annotation.JsonIgnore
...
+      null
+    }
+    return OffenderLocation(
+      levels = location.getLocationWithParents(),
+      lastPermanentLevels = lastPermanentLocation?.getLocationWithParents(),
+    )
   }
+
+  private fun isTemporaryLocation(locationCode: String?): Boolean =
+    setOf("RECP", "RECEP", "RECEPTION", "COURT", "TAP", "ECL", "CSWAP").contains(locationCode)
```
AledWEvans, line 54: is location code always uppercase? would also use a constant for this,... but compilers are fairly clever these days!

AledWEvans (commented):
looks like it's been merged,...

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderLocationService.kt
petergphillips: yes location code is always uppercase

PR #1557, opened 2023-04-11
HAAR-1340 Case Note Performance Improvement - Remove SQL call to group case note types, use in memory grouping

Mjwillis:


src/main/java/uk/gov/justice/hmpps/prison/api/resource/BookingResource.java
```diff
@@ -444,7 +444,7 @@ public List<Assessment> getAssessments(@PathVariable("bookingId") @Parameter(des
         return inmateService.getAssessments(bookingId);
     }
 
-    @Operation(summary = "Offender case notes.", description = "Offender case notes.")
+    @Operation(summary = "Offender case notes.", description = "Offender case notes.", hidden = true)
```
Mjwillis, line 447: These endpoint shouldn't be used by clients (only case note service)

andymarke: Maybe add  a new role that only the case notes service has to restrict access?

Mjwillis: That would make sense - obs not in this PR

andymarke (approved):
LGTM

PR #1561, opened 2023-04-18
SDIT-749 return non associations for past bookings

lbennett-moj:


andymarke (approved):
LGTM
We could do with a better integration test

PR #1565, opened 2023-04-21
SDIT-653: 🐛 Fix controller advice null pointer

petergphillips:
SDIT-653: 🐛 Fix controller advice null pointer


src/main/java/uk/gov/justice/hmpps/prison/web/handler/ControllerAdvice.kt
```diff
@@ -0,0 +1,389 @@
...
+import uk.gov.justice.hmpps.prison.api.model.ErrorResponse
+import uk.gov.justice.hmpps.prison.service.BadRequestException
+import uk.gov.justice.hmpps.prison.service.ConflictingRequestException
+import uk.gov.justice.hmpps.prison.service.EntityAlreadyExistsException
+import uk.gov.justice.hmpps.prison.service.EntityNotFoundException
+import uk.gov.justice.hmpps.prison.service.NoContentException
+import java.sql.SQLException
+import java.util.stream.Collectors
+
+@RestControllerAdvice(basePackages = ["uk.gov.justice.hmpps.prison.api.resource"])
```
petergphillips, line 33: was previously also including `"uk.gov.justice.hmpps.nomis.api.resource.controller"` unnecessarily as we don't have anything under there now

src/main/java/uk/gov/justice/hmpps/prison/web/handler/ControllerAdvice.kt
```diff
@@ -0,0 +1,389 @@
...
+  @ExceptionHandler(jakarta.persistence.EntityNotFoundException::class)
+  fun handleEntityNotFoundException(e: jakarta.persistence.EntityNotFoundException): ResponseEntity<ErrorResponse> =
+    ResponseEntity
+      .status(HttpStatus.NOT_FOUND)
+      .body(
+        ErrorResponse
+          .builder()
+          .userMessage(e.message)
+          .status(HttpStatus.NOT_FOUND.value())
+          .developerMessage(e.cause?.message)
```
petergphillips, line 169: this was the bug - switch to using `?` as could be `null`

src/main/java/uk/gov/justice/hmpps/prison/web/handler/ControllerAdvice.kt
```diff
@@ -0,0 +1,389 @@
...
+          .developerMessage(e.message)
+          .build(),
+      )
+  }
+
+  @ExceptionHandler(JpaSystemException::class)
+  fun handleJpaException(e: JpaSystemException): ResponseEntity<ErrorResponse> {
+    try {
+      if (e.cause is GenericJDBCException && (e.cause as GenericJDBCException).cause is SQLException) {
+        return handleJpaSqlException(e, e.cause!!.cause!!)
```
petergphillips, line 344: pass through `e.cause.cause` as we've already checked they're not null (as they have a type)

src/main/java/uk/gov/justice/hmpps/prison/web/handler/ControllerAdvice.kt
```diff
@@ -0,0 +1,389 @@
...
+      }
+    } catch (ex: Exception) {
+      // we don't know the shape of this exception - handling below
+      log.error("Caught error trying to parse jpa sql exception", ex)
```
petergphillips, line 348: also now log if we can't parse the exception, previously was being ignored

andymarke (approved):
LGTM

PR #1567, opened 2023-04-24
SAA-434 new endpoint to get agency pay profile data

timharrison-moj:
This is to allow the Activities services to retrieve the per-prison pay limits, for example, the min and max half-rate rate, max bonus rate and max piece work rate. At present access is allowed for any valid token.


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/AgencyPrisonerPayProfileIntTest.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+import org.springframework.core.ParameterizedTypeReference;
+import org.springframework.http.HttpMethod;
+import uk.gov.justice.hmpps.prison.api.model.AgyPrisonerPayProfile;
+import uk.gov.justice.hmpps.prison.api.model.ErrorResponse;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class AgencyPrisonerPayProfileIntTest extends ResourceTest {
```
petergphillips, line 15: new classes should be written in kotlin

src/main/java/uk/gov/justice/hmpps/prison/api/model/AgyPrisonerPayProfileDto.kt
```diff
@@ -0,0 +1,34 @@
+package uk.gov.justice.hmpps.prison.api.model
+import java.math.BigDecimal
+import java.math.RoundingMode
+import java.time.LocalDate
+
+data class AgyPrisonerPayProfileDto(
```
petergphillips, line 6: not a great fan of abbreviating `Agency` to `Agency`.  Can we just call it `Agency` please?

src/main/java/uk/gov/justice/hmpps/prison/repository/AgencyRepository.java
```diff
@@ -164,6 +172,23 @@ public List<Location> getAgencyLocationsBooked(final String agencyId, final Loca
         return locations.stream().map(LocationDto::toLocation).collect(toList());
     }
 
+    public Optional<AgyPrisonerPayProfile> getAgyPrisonerPayProfile(final String agencyId) {
```
petergphillips, line 175: since this looks like a very simple table, would have been far better to map using JPA rather than increasing the amount of this type of code that we're trying to gradually remove

src/main/java/uk/gov/justice/hmpps/prison/api/resource/AgencyResource.java
```diff
@@ -417,4 +418,17 @@ public void deleteAgencyAddressPhoneContact(
...
+
+
+    @ApiResponses({
+        @ApiResponse(responseCode = "200", description = "OK"),
+        @ApiResponse(responseCode = "400", description = "Invalid request.", content = {@Content(mediaType = "application/json", schema = @Schema(implementation = ErrorResponse.class))}),
+        @ApiResponse(responseCode = "404", description = "Requested resource not found.", content = {@Content(mediaType = "application/json", schema = @Schema(implementation = ErrorResponse.class))}),
+        @ApiResponse(responseCode = "500", description = "Unrecoverable error occurred whilst processing request.", content = {@Content(mediaType = "application/json", schema = @Schema(implementation = ErrorResponse.class))})})
+    @Operation(summary = "Return the payment profile data for the given Agency.", description = "Each agency can configure its own pay profile and this endpoint provides its key data, such as min/max pay and bonus rates.")
+    @GetMapping("/{agencyId}/pay-profile")
+    @SlowReportQuery
```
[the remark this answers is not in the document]
timharrison-moj: The thinking was that the data doesn't change often so doesn't need to come from the live NOMIS so the read-only replica would do just as well, or caching.

src/main/java/uk/gov/justice/hmpps/prison/api/model/AgyPrisonerPayProfileDto.kt
timharrison-moj: The name of the table in NOMIS is AGY_PRISONER_PAY_PROFILE in this case - abbreviated at source.  Happy either way though.

src/main/java/uk/gov/justice/hmpps/prison/repository/AgencyRepository.java
```diff
@@ -164,6 +172,23 @@ public List<Location> getAgencyLocationsBooked(final String agencyId, final Loca
...
+        final var builder = queryBuilderFactory.getQueryBuilder(initialSql, AGY_PRISONER_PAY_PROFILE_ROW_MAPPER);
+        final var sql = builder.build();
+        final var params = createParams("agencyId", agencyId);
+
+        AgyPrisonerPayProfileDto agyPrisonerPayProfileDto;
+
+        try {
+            agyPrisonerPayProfileDto = jdbcTemplate.queryForObject(sql, params, AGY_PRISONER_PAY_PROFILE_ROW_MAPPER);
+        } catch (DataAccessException ex) {
+            agyPrisonerPayProfileDto = null;
```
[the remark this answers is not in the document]
steverendell: indeed, does this get thrown if there is >1 row found ? Maybe getting a list is better so you can differentiate from no rows

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/AgencyPrisonerPayProfileIntTest.java
timharrison-moj: Ok  - will do.

src/main/java/uk/gov/justice/hmpps/prison/api/model/AgyPrisonerPayProfile.java
```diff
@@ -0,0 +1,82 @@
...
+  private String agencyId;
+
+  @Schema(description = "The start date when this pay profile took effect", example = "2022-10-01")
+  private LocalDate startDate;
+
+  @Schema(description = "The end date when this pay profile will stop taking effect", example = "2027-10-01")
+  private LocalDate endDate;
+
+  @Schema(description = "Whether automatic payments are enabled", example = "Y")
+  private String autoPayFlag;
```
petergphillips, line 26: is it best to map this to a boolean instead?

src/main/java/uk/gov/justice/hmpps/prison/repository/AgencyRepository.java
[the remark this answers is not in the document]
timharrison-moj: Reworking this to use a standalone JPA repo, so this will come out.

src/main/java/uk/gov/justice/hmpps/prison/api/model/AgyPrisonerPayProfile.java
timharrison-moj: Yes, I think you're right. On it.

src/main/java/uk/gov/justice/hmpps/prison/repository/sql/AgencyRepositorySql.kt
```diff
@@ -169,4 +169,26 @@ enum class AgencyRepositorySql(val sql: String) {
         ORDER BY USER_DESCRIPTION
     """,
   ),
+
+  GET_AGY_PRISONER_PAY_PROFILE(
```
Mjwillis, line 173: We normally don't write SQL now but use JPA

src/main/java/uk/gov/justice/hmpps/prison/repository/AgencyRepository.java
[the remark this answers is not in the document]
timharrison-moj: Replaced with a JPA repository, where the method will only find 1 active row, or none if an unknown agency code is supplied.

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/AgencyPrisonerPayProfileIntTest.java
timharrison-moj: Now done.

src/main/java/uk/gov/justice/hmpps/prison/api/model/AgyPrisonerPayProfileDto.kt
timharrison-moj: Refactored the model to use the full name, whilst the entity mirrors the table name in NOMIS.

src/main/java/uk/gov/justice/hmpps/prison/repository/AgencyRepository.java
timharrison-moj: Now done.

src/main/java/uk/gov/justice/hmpps/prison/api/resource/AgencyResource.java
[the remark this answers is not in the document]
timharrison-moj: Removed the SlowReportQuery annotation, but not sure it matters for this - should be a quick look-up from table with so few rows.

src/main/java/uk/gov/justice/hmpps/prison/api/model/AgyPrisonerPayProfile.java
timharrison-moj: Converted to a boolean for the model object, from the Y/N flag in the entity.

src/main/java/uk/gov/justice/hmpps/prison/repository/sql/AgencyRepositorySql.kt
timharrison-moj: Now using JPA.

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/AgencyPrisonerPayProfileIntTest.kt
```diff
@@ -0,0 +1,70 @@
...
+import uk.gov.justice.hmpps.prison.api.model.AgencyPrisonerPayProfile
+import uk.gov.justice.hmpps.prison.api.model.ErrorResponse
+import java.math.BigDecimal
+import java.time.LocalDate
+
+class AgencyPrisonerPayProfileIntTest : ResourceTest() {
+  @Test
+  fun agencyPayProfile_returnsSuccessAndData() {
+    val requestEntity = createHttpEntityWithBearerAuthorisation("ITAG_USER", listOf("ROLE_VIEW_PRISON_DATA"), emptyMap())
+    val responseEntity = testRestTemplate.exchange(
```
petergphillips, line 16: for future reference new tests should be using webclient, not the old rest template

src/main/java/uk/gov/justice/hmpps/prison/api/model/AgencyPrisonerPayProfile.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude
+import io.swagger.v3.oas.annotations.media.Schema
+import java.math.BigDecimal
+import java.time.LocalDate
+
+@Schema(description = "Agency prisoner pay profile")
+@JsonInclude(JsonInclude.Include.NON_NULL)
+data class AgencyPrisonerPayProfile(
+  @Schema(description = "Agency identifier", example = "MDI")
+  val agencyId: String?,
```
petergphillips, line 12: this looks weird - why is `agencyId` here nullable when the domain model class can't be?

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/AgencyPrisonerPayProfileIntTest.kt
timharrison-moj: Noted, thanks.

src/test/java/uk/gov/justice/hmpps/prison/service/AgencyPrisonerPayProfileServiceTest.kt
```diff
@@ -0,0 +1,108 @@
...
+      ),
+    )
+
+    val payProfile = service.getAgencyPrisonerPayProfile("MDI")
+
+    with(payProfile) {
+      assertThat(agencyId).isEqualTo("MDI")
+      assertThat(startDate).isEqualTo(today.minusDays(8))
+      assertThat(endDate).isEqualTo(today)
+    }
```
timharrison-moj, line 72: This test checks the logic for start/end dates when multiple rows exist for an agency.

src/main/java/uk/gov/justice/hmpps/prison/service/AgencyPrisonerPayProfileService.kt
```diff
@@ -12,10 +13,17 @@ class AgencyPrisonerPayProfileService(
...
-      .findAgencyPrisonerPayProfileByAgyLocIdEqualsAndEndDateIsNullAndStartDateIsLessThanEqual(agencyId, LocalDate.now())
-      .orElseThrow(EntityNotFoundException.withId(agencyId))
-    return entityToModel(result)
+    val today = LocalDate.now()
+    val agencyProfilesList = agencyPrisonerPayProfileRepository
+      .findAgencyPrisonerPayProfileByAgyLocId(agencyId)
+      .filter { agy -> !agy.startDate.isAfter(today) && (agy.endDate == null || !agy.endDate.isBefore(today)) }
+
+    if (agencyProfilesList.isEmpty()) {
+      log.error("No AGY_PRISONER_PAY_PROFILES row is active for agency {} on {}", agencyId, today)
```
petergphillips, line 22: don't think this should be `error`.  For the private prisons for example there are no active prisoner pay profiles.

petergphillips (approved):
LGTM

PR #1573, opened 2023-05-02
SDI-777 Fetch author for case notes for `findAll`

andymarke:
Currently `author` is eager loaded as a N + 1 type query, when showing all case notes in DPS this leads to 100s of separate `select` from `staff_members` table. This change means that is all done via a `join` instead.

Avoiding high number of separate SQL select held when latency with database is high (e.g when in AWS)


petergphillips (approved):
LGTM

PR #1575, opened 2023-05-03
SDI-778 Fetch booking/offender for bed histories by prison and date

andymarke:
Calls to this endpoint e.g. /api/cell/NMI/history/2023-04-28
result in ( n * 3)  + 1 SQL for each bed history record, each booking and offender record was loaded by individual and there was a separate repo call for the internal location that was already loaded

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/BedAssignmentHistory.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/BedAssignmentHistory.java
@@ -1,6 +1,8 @@
 package uk.gov.justice.hmpps.prison.repository.jpa.model;
 
-import jakarta.persistence.FetchType;
+import jakarta.persistence.NamedAttributeNode;
+import jakarta.persistence.NamedEntityGraph;
+import jakarta.persistence.NamedSubgraph;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/BedAssignmentHistory.java
```diff
@@ -51,7 +69,8 @@ public static class BedAssignmentHistoryPK implements Serializable {
     @Column(name = "LIVING_UNIT_ID")
     private Long livingUnitId;
 
-    @ManyToOne(fetch = FetchType.LAZY)
+    @NotFound(action = IGNORE)
```
andymarke, line 72: LIVING_UNIT_ID has no foreign key and not all history records have a matching cell due to bad NOMIS data, so this needs to be ignored (which therefore means it can not be LAZY loaded)

src/main/java/uk/gov/justice/hmpps/prison/service/BedAssignmentHistoryService.java
```diff
@@ -106,7 +106,7 @@ public List<BedAssignment> getBedAssignmentsHistoryByDateForAgency(final String
     }
 
     private BedAssignment transform(final BedAssignmentHistory assignment) {
-        final var agencyInternalLocation = locationRepository.findOneByLocationId(assignment.getLivingUnitId());
```
andymarke, line 109: already loaded in the assignment records so no need for the additional repo lookup

petergphillips (approved):
LGTM

PR #1576, opened 2023-05-03
DCS-2091 add endpoint for only active bookings

GurnankCheema:
New endpoint will return active bookings only  - needed by HDC


michalnawrockidj:
The work is postponed now. When we find a time to come back we need to add tests for 
bookingRepository.getOffenderSentenceSummary  in BookingRepositoryTest 
and OffenderSentenceResourceImplIntTest to cover OffenderSentenceResource.getActiveOffenderSentences to finis this story.

PR #1579, opened 2023-05-04
SDI-786 Fetch internal location profiles to avoid n + 1 SQL

andymarke:
For /agencies/{agencyId?/cellsWithCapacity cell profile attributes are currently loaded via a separate repo call which results in a n + 1 query.

For this endpoint only force a join fetch to reduce to a single SQL call to avoid latency issues when Oracle is move to AWS


petergphillips (approved):
LGTM

PR #1580, opened 2023-05-05
SDI-779 Remove n + 1 SQL for deprecated GET /api/agencies/prison

andymarke:
Currently this version of endpoint is returning prison phones numbers and address and phones numbers for each address.

Each prison has a single address and 2 numbers and each address has the the same 2 numbers. There is around 127 prisons.

Convert the n * 3 + 1  query to a single Cartesian Product join - we can do this because we know there is a "small" number of prisons and each has a single address and 2 numbers.

Phones and Addresses need to become a set else Hibernate can not load addresses and phones at same time. There is only one primary address per prison but the impact of this PR is the phone numbers might be returned in a different order (though no clients even read the phones let alone rely on the order rather than the type)

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/AgencyLocation.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/AgencyLocation.java
@@ -106,7 +106,7 @@ public class AgencyLocation extends AuditableEntity {
     @OneToMany(mappedBy = "agency", cascade = CascadeType.ALL, fetch = FetchType.LAZY, orphanRemoval = true)
     @Where(clause = "OWNER_CLASS = '"+AgencyAddress.ADDR_TYPE+"'")
     @Default
-    private List<AgencyAddress> addresses = new ArrayList<>();
+    private Set<AgencyAddress> addresses = new HashSet<>();
 
     @OneToMany(mappedBy = "agency", cascade = CascadeType.ALL, fetch = FetchType.LAZY, orphanRemoval = true)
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/AgencyLocation.java
```diff
@@ -41,6 +46,18 @@
 @Table(name = "AGENCY_LOCATIONS")
 @ToString(of = {"id", "description"})
 @With
+@NamedEntityGraph(
```
andymarke, line 49: This will result in a Cartesian Product but will still perform better than the n * 3 + 1 queries since we know real world data is constrained to a single address and 2 phone numbers so it will read 127 * 1 * 2 * 2 which is diddly squat

src/main/java/uk/gov/justice/hmpps/prison/service/AgencyService.java
```diff
@@ -303,7 +303,7 @@ private PrisonContactDetail getPrisonContactDetail(final uk.gov.justice.hmpps.pr
                 .build();
         } else {
             final var primaryAddress = prison.getAddresses().stream()
-                .filter(a -> "Y".equals(a.getPrimaryFlag())).findFirst().orElse(prison.getAddresses().get(0));
+                .filter(a -> "Y".equals(a.getPrimaryFlag())).findFirst().orElse(prison.getAddresses().iterator().next());
```
andymarke, line 306: really `get(0)` meant in this context "get any" so just use `next` on the set since we don't care which one we get

mikehalmamoj (approved):

PR #1582, opened 2023-05-10
SDI-802 Remove n + 1 SQL for api/movements/{agencyId}/in

andymarke:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/ExternalMovement.java
```diff
@@ -80,11 +104,11 @@ public static class PK implements Serializable {
     @JoinColumn(name = "ARREST_AGENCY_LOC_ID")
     private AgencyLocation arrestAgencyLocation;
 
-    @ManyToOne(optional = false, fetch = FetchType.LAZY)
+    @ManyToOne(fetch = FetchType.LAZY)
     @JoinColumn(name = "FROM_AGY_LOC_ID")
```
andymarke, line 108: these are indeed nullable, movements can be to and from cities and other entities. Making it mandatory would mean this is an inner join when it should be an outer join

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/ExternalMovement.java
```diff
@@ -104,23 +128,23 @@ public static class PK implements Serializable {
 
     @ManyToOne(fetch = FetchType.LAZY)
     @JoinColumnsOrFormulas(value = {
-            @JoinColumnOrFormula(formula = @JoinFormula(value = "'" + CITY + "'", referencedColumnName = "domain")),
```
andymarke, line 107: just formatting

petergphillips (approved):
LGTM

petergphillips (approved):
LGTM

petergphillips (approved):
LGTM

PR #1594, opened 2023-05-18
Expose date and time of image capture and creation

emileswarts:
These 2 new fields use existing columns in the database and add them to the API response. Knowing the time in addition to the date is a requirements from the HMPPS Integration API team.


src/main/java/uk/gov/justice/hmpps/prison/api/model/ImageDetail.java
```diff
@@ -36,6 +37,14 @@ public class ImageDetail {
     @NotNull
     private LocalDate captureDate;
 
+    @Schema(description = "Date and time of image capture", example = "2008-08-27 01:01:01")
+    @NotNull
+    private LocalDateTime captureDateTime;
+
+    @Schema(description = "Date and time of image creation", example = "2008-08-27 01:01:01")
```
petergphillips, line 44: Why do you need this data? It is never exposed in NOMIS to the end user so seems very odd that it is now exposed.  Image creation is an odd term - would be better to call it image upload.

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderImage.java
```diff
@@ -48,6 +48,9 @@ public class OffenderImage extends AuditableEntity {
     @Column(name = "CAPTURE_DATETIME")
     private LocalDateTime captureDateTime;
 
+    @Column(insertable=false, updatable=false, name = "CREATE_DATETIME")
+    private LocalDateTime createDateTime;
```
petergphillips, line 52: unnecessary - this class extends `AuditableEntry` that has the fields already

src/main/java/uk/gov/justice/hmpps/prison/api/model/ImageDetail.java
emileswarts: The requirement isn't to expose this in NOMIS.
We need this information in the HMPPS Integration API. Our consumer (MAPPS) are interested in this data to display in their system.

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderImage.java
emileswarts: Missed that, removed this 👍

src/main/java/uk/gov/justice/hmpps/prison/api/model/ImageDetail.java
emileswarts: We're not going to expose the create date, capture date will be fine

petergphillips (approved):
LGTM

PR #1599, opened 2023-05-25
HIA-360 Add active field to ImageDetail response model

mcoffey-mt:
Expose `active` field to allow clients to select the most relevant images.

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/api/model/ImageDetail.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/api/model/ImageDetail.java
@@ -32,6 +32,10 @@ public class ImageDetail {
     @NotNull
     private Long imageId;
 
+    @Schema(requiredMode = RequiredMode.REQUIRED, description = "Active", example = "false")
+    @NotNull
+    private Boolean active;
+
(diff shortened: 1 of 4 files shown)


mikehalmamoj (approved):

src/main/java/uk/gov/justice/hmpps/prison/api/model/ImageDetail.java
```diff
@@ -33,6 +33,10 @@ public class ImageDetail {
     @NotNull
     private Long imageId;
 
+    @Schema(requiredMode = RequiredMode.REQUIRED, description = "Active", example = "false")
+    @NotNull
+    private Boolean active;
```
petergphillips, line 38: seem odd to annotate with `NonNull` but yet use the object `Boolean` rather than the primitive.  Also `active` doesn't really mean anything to me here.  What does it mean for an image to be `active`?

chubberlisk: Thanks again for the feedback, we've created https://github.com/ministryofjustice/prison-api/pull/1601 to address it!

PR #1601, opened 2023-05-26
HIA-360 Update active property for an image

chubberlisk:
- Use boolean primitive type instead of Boolean.
- Add a more meaningful description for the property.

Follow-up PR to address feedback in #1599.


petergphillips (approved):
LGTM

PR #1602, opened 2023-05-30
SDIT-826 Model agency establishment as embeddedid

mikehalmamoj:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderSubAccount.java
```diff
@@ -36,7 +36,6 @@ public class OffenderSubAccount {
     @Column(name = "TRUST_ACCOUNT_CODE", nullable = false, insertable = false, updatable = false)
     private Long accountCode;
 
-    @Id
```
mikehalmamoj, line 39: Not actually included in the id and causes an error in latest Hibernate verison.

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderTrustAccount.java
```diff
@@ -31,7 +31,6 @@ public class OffenderTrustAccount {
     @Column(name = "OFFENDER_ID", nullable = false, insertable = false, updatable = false)
     private Long offenderId;
 
-    @Id
```
mikehalmamoj, line 34: Not actually included in the id and causes an error in latest Hibernate verison.

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/AssessmentRepository.java
```diff
@@ -2,18 +2,15 @@
...
 
 import java.util.List;
-import java.util.Optional;
 
 public interface AssessmentRepository extends CrudRepository<AssessmentEntry, Long> {
     @Query("SELECT question FROM AssessmentEntry question " +
         "INNER JOIN question.parentAssessment questionSet INNER JOIN questionSet.parentAssessment type " +
         "WHERE type.assessmentId = :assessmentId AND type.parentAssessment is null " +
-        "AND type.cellSharingAlertFlag = 'Y' AND questionSet.assessmentCode != 'COMPLETE' " +
+        "AND type.cellSharingAlertFlag = 'Y' AND questionSet.assessmentCode <> 'COMPLETE' " +
```
mikehalmamoj, line 13: Work around bug in latest Spring data which removed `!=` alias allowed by Hibernate - fixed in version 3.2 which has not been released yet.

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/AgencyEstablishmentType.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import jakarta.persistence.DiscriminatorValue
+import jakarta.persistence.Entity
+import lombok.NoArgsConstructor
+
+@Entity
+@DiscriminatorValue(AgencyEstablishmentType.ESTABLISHMENT_TYPE)
+@NoArgsConstructor
+class AgencyEstablishmentType(code: String?, description: String?) :
+  ReferenceCode(ESTABLISHMENT_TYPE, code, description) {
```
mikehalmamoj, line 11: Model this as a reference code instead of loading manually.

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/AgencyLocationEstablishment.kt
```diff
@@ -0,0 +1,60 @@
...
+
+  @JoinColumn(name = "ESTABLISHMENT_TYPE", nullable = false)
+  val establishmentTypeCode: String,
+) : Serializable
+
+@Entity
+@Table(name = "AGY_LOC_ESTABLISHMENTS")
+data class AgencyLocationEstablishment(
+  @EmbeddedId
+  val id: AgencyLocationEstablishmentId,
```
mikehalmamoj, line 29: Switch to `EmbeddedId` as the `IdClass` wasn't mapping correctly on a `OneToMany` annotation in the parent class `AgencyLocation`.

src/test/java/uk/gov/justice/hmpps/prison/service/BookingServiceTest.java
```diff
@@ -231,7 +230,7 @@ public void verifyCanViewSensitiveBookingInfo() {
 
     @Test
     public void verifyCanViewSensitiveBookingInfo_systemUser() {
-        when(authenticationFacade.isOverrideRole(any())).thenReturn(true);
+        when(authenticationFacade.isOverrideRole(any(String[].class))).thenReturn(true);
```
mikehalmamoj, line 233: Mocks for vararg parameters [changed in Mockito 5](https://github.com/mockito/mockito/releases/tag/v5.0.0) so that you have to either match the correct number of args or instead use an array type.

andymarke (approved):
lgtm

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/AgencyLocationEstablishment.kt
mikehalmamoj: Actually to be more accurate... 

We could have just continued to use `@Id` on the columns as long as we modelled `AgencyLocation` in the child as a Hibernate managed object instead of just the id. This would then fulfil the oppopsite side of the bidirectional relationship and Hibernate would be happy. (and note we could have removed the `@IdClass` as that's only needed for repositories but we don't have one for this class).

BUT as I changed the `establishmentType` to be a reference code this means that we had to use a `JoinFormula` and you can't have both `@Id` and a join formula on the same column (Hibernate throws in that case). Hence we need `@EmbeddedId`.

PR #1603, opened 2023-05-30
NN-5130 Remove Booking Id from Adjudication creation request

Mjwillis:
Also removed booking if from create new adjudication


timoj58:
Need to take bookingId out of the new report creation object

  @Schema(required = true, description = "Booking number", example = "123456")
    @NotNull
    private Long bookingId;

andymarke (approved):
LGTM

PR #1607, opened 2023-06-03
CDPS-114: Converting HsqlConnectionAspect to Kotlin and fixing proxy user check

brightonsbox:
When running Prison API locally with the `nomis-hsqldb` profile and hitting the `/users/me/caseLoads` endpoint, I encountered an error:

```
Caused by: java.lang.NullPointerException: Cannot invoke "String.isBlank()" because "proxyUser" is null
	at uk.gov.justice.hmpps.prison.aop.connectionproxy.HsqlConnectionAspect.configureNomisConnection(HsqlConnectionAspect.java:28)
```

The `HsqlConnectionAspect` implementation doesn't do a null check for the proxyUser, unlike `OracleConnectionAspect`.  The call to set MDC.put(PROXY_USER) is in `ProxyUserAspect` and only happens if the `@ProxyUser` annotation is set on the API endpoint, which is not the case with `/users/me/caseLoads`.  So the `MDC.get(PROXY_USER)` will always return null for this endpoint and would always break. 

This PR introduces a null check as well as the isBlank() check and also converts the class to Kotlin.


src/main/java/uk/gov/justice/hmpps/prison/aop/connectionproxy/HsqlConnectionAspect.kt
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  private fun isNomisProxyUser(): Boolean = isProxyUser() && authSource() == NOMIS
+
+  private fun isProxyUser(): Boolean = !MDC.get(PROXY_USER).isNullOrBlank()
+
+  private fun authSource(): AuthSource = authenticationFacade.authenticationSource
+
+  companion object {
+    val log = LoggerFactory.getLogger(this::class.java)!!
```
steverendell, line 59: normally private, and dont need the !!

steverendell (approved):
lgtm with very minor comment

steverendell (approved):

PR #1624, opened 2023-06-07
SDIT-792: 📝 Start to document roles

petergphillips:


src/main/resources/documentation/service-description.html
```diff
@@ -0,0 +1,105 @@
...
+    service.  This also augments the results from the incentives and restricted patients service.  That service should
+    be used in preference, with this service only then being used if the data is not available there.  Furthermore, the
+    Prisoner Search service also provides key prisoner view that could be suitable.
+  </p>
+  <h2 id="role-restrictions">Role restrictions</h2>
+  The following roles are currently in use in this service:
+  <ul>
+    <li>
+      APPROVE_CATEGORISATION
+    </li>
```
mikehalmamoj, line 26: This seems like a bit of a maintenance headache and will probably drift? Maybe we could create a Gradle task to check that all roles in use are included in this doc, or even add them to this list. Maybe.

mikehalmamoj (approved):

PR #1630, opened 2023-06-09
Improve GET /api/agencies/{agencyId}/cellsWithCapacity

steverendell:
Reduce cartesian join as was still slow

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/AgencyInternalLocation.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/AgencyInternalLocation.java
@@ -1,9 +1,17 @@
 package uk.gov.justice.hmpps.prison.repository.jpa.model;
 
+import jakarta.persistence.Column;
+import jakarta.persistence.Convert;
+import jakarta.persistence.Entity;
+import jakarta.persistence.FetchType;
+import jakarta.persistence.Id;
+import jakarta.persistence.JoinColumn;
+import jakarta.persistence.ManyToOne;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/AgencyInternalLocation.java
```diff
@@ -103,6 +103,7 @@ public class AgencyInternalLocation {
     private Integer capacity;
 
     @OneToMany(fetch = FetchType.LAZY)
+    @Where(clause = "INT_LOC_PROFILE_TYPE = 'HOU_UNIT_ATT' AND INT_LOC_PROFILE_CODE is not NULL")
     @JoinColumn(name = "INTERNAL_LOCATION_ID", referencedColumnName = "INTERNAL_LOCATION_ID")
     private List<AgencyInternalLocationProfile> profiles;
```
andymarke, line 108: so is there some associated Java code that was *always* applying this exact filter in all scenarios ?

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/AgencyInternalLocation.java
steverendell: Yep, filtered on:
 ```
public boolean isAttribute() {
        return profileType.equals(HousingAttributeReferenceCode.DOMAIN) && housingAttributeReferenceCode != null;
}
```

PR #1637, opened 2023-06-13
NN-5202 Allow possibility of returning other locations that have cell capacity

Mjwillis:
This will allow for an area at the top level where prisoners can be placed (even if temporary) e.g. TAP, COURT, RECP, CSWAP


andymarke (approved):
LGTM, see my Slack comment

PR #1649, opened 2023-06-16
SDIT-860 Add release detail to entity graph to fix EntityNotFoundException

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/Offender.java
```diff
@@ -68,6 +68,7 @@
                 @NamedAttributeNode(value = "sentences", subgraph = "sentence-terms"),
                 @NamedAttributeNode(value = "location"),
                 @NamedAttributeNode(value = "assignedLivingUnit", subgraph = "agency-internal-location-details"),
+                @NamedAttributeNode(value = "releaseDetail"),
```
karenmillermoj, line 71: This has to be pulled in as part of the graph, otherwise it attempts to pull it in by its associated bookingId (which can be nullable) - very odd

andymarke (approved):
lgtm

PR #1660, opened 2023-06-22
NN-5231 removing verify offender access and correct existing offender lookup to return 404

timoj58:
this api is hidden, and only used api to api, and therefore this check is not required, and is causing issues with transfers, where the prisoner resides outside of the users caseload.

Correct the bookingId lookup, to return a 404 and use the correct format mask


andymarke (approved):
LGTM

PR #1666, opened 2023-06-26
SDIT-904 Improve GET /api/offenders/{offenderNo}/addresses

steverendell:


src/main/java/uk/gov/justice/hmpps/prison/service/OffenderAddressService.java
```diff
@@ -16,12 +17,13 @@
...
 
     @VerifyOffenderAccess(overrideRoles = {"SYSTEM_USER", "GLOBAL_SEARCH", "VIEW_PRISONER_DATA"})
     public List<AddressDto> getAddressesByOffenderNo(@NotNull final String offenderNo) {
-        final var optionalOffenderBooking = offenderBookingRepository.findByOffenderNomsIdAndActive(offenderNo, true);
-        final var offenderBooking = optionalOffenderBooking.orElseThrow(EntityNotFoundException.withMessage(String.format("No active offender bookings found for offender number %s\n", offenderNo)));
-
-        return AddressTransformer.translate(offenderBooking.getOffender().getRootOffender().getAddresses());
+        final var offenderBooking = offenderBookingRepository.findByOffenderNomsIdAndActive(offenderNo, true)
+            .orElseThrow(EntityNotFoundException.withMessage(String.format("No active offender bookings found for offender number %s\n", offenderNo)));
+        final var addresses = offenderAddressRepository.findByOffenderId(offenderBooking.getOffender().getRootOffender().getId());
```
mikehalmamoj, line 26: If you've switched to getting the addresses directly from the repo then should you remove the addresses from the Offender entity? Otherwise somebody else might use it and we're back to square one...

mikehalmamoj (approved):

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderAddressService.java
steverendell: Now that is a good idea! I didnt think whether anything else used that 1tomany .. nothing does.

mikehalmamoj (approved):

PR #1674, opened 2023-06-28
SDIT-908 Improve GET /api/agencies/{agencyId}/locations/type/{type}

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/AgencyInternalLocation.java
```diff
@@ -93,6 +94,7 @@ public class AgencyInternalLocation {
     @OneToMany(mappedBy = "parentLocation", fetch = FetchType.LAZY, cascade = CascadeType.ALL)
     @Exclude
     @Default
+    @BatchSize(size = 1000)
```
karenmillermoj, line 97: This improves the number of sql calls from 1028 to 4 - Introducing an entity graph only improved this to 3, so I left the call as it was

andymarke (approved):
lgtm

PR #1679, opened 2023-06-29
SDIT-912 Add entity graph and use jpa named queries for offender assessment queries

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderAssessmentRepository.kt
```diff
@@ -1,24 +1,16 @@
...
-          oa FROM OffenderAssessment oa 
-          INNER JOIN oa.offenderBooking booking
-          INNER JOIN booking.offender offender 
-          INNER JOIN oa.assessmentType assessment
-        WHERE
-          offender.nomsId IN (:offenderNos) AND assessment.cellSharingAlertFlag = 'Y'
-        ORDER BY oa.assessmentDate DESC, oa.assessmentSeq DESC
-        """,
-  )
-  fun findByCsraAssessmentAndByOffenderNosOrderByLatestFirst(offenderNos: List<String>): List<OffenderAssessment>
```
karenmillermoj, line 23: This is the significant change to this PR.
The native query now uses the derived named query instead and the new query ("findWithDetails...") also uses the same logic.

PR #1684, opened 2023-07-04
SDIT-915: ♻️  Rename endpoint as deals with bookingIds

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/api/resource/BookingMovementsResource.java
```diff
@@ -51,26 +52,15 @@
 @RequestMapping(value = "${api.base.path}/bookings", produces = "application/json")
 @Validated
 @Slf4j
-public class OffenderMovementsResource {
+@AllArgsConstructor
+public class BookingMovementsResource {
```
petergphillips, line 56: this endpoint is for `/api/bookings` - so makes sense to be a booking movement resource, especially as I want to add an identical endpoint under `/api/offenders` for cell move

mikehalmamoj (approved):

PR #1688, opened 2023-07-04
SDIT-915: ✨ Cell move endpoint that takes an offender instead

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderMovementsResource.kt
```diff
@@ -0,0 +1,66 @@
...
+  )
+  @Operation(
+    summary = "Move the prisoner to the specified cell.",
+    description = """
+      Unilink specific version of /api/bookings/{bookingId}/living-unit/{internalLocationDescription}.
+      Requires either a valid user token or a token with UNILINK role.
+      """,
+  )
+  @PutMapping("/{offenderNo}/living-unit/{internalLocationDescription}")
+  @ProxyUser
```
petergphillips, line 47: hopefully this will work for unilink as the system will detect that it is a client credentials user instead

src/main/java/uk/gov/justice/hmpps/prison/service/MovementUpdateService.java
```diff
@@ -46,7 +46,7 @@ public MovementUpdateService(
     }
 
     @Transactional
-    @VerifyBookingAccess
+    @VerifyBookingAccess(overrideRoles = {"ROLE_SYSTEM_USER", "ROLE_UNILINK"})
```
petergphillips, line 49: this is adding a new role, since I've left `SYSTEM_USER` in there it will work for any existing users calling the booking endpoint

.circleci/config.yml
```diff
@@ -10,7 +10,7 @@ executor:
     tag: "19.0"
   environment:
     # Setting max gradle workers to 1 (which means 1 running concurrently, not max 1!) - because although this doesn't stop the ktlint workers from starting it does seem to prevent the test executor from needing as much memory.
-    _JAVA_OPTIONS: -Xmx1280m -XX:ParallelGCThreads=2 -XX:ConcGCThreads=2
+    _JAVA_OPTIONS: -Xmx1536m -XX:ParallelGCThreads=2 -XX:ConcGCThreads=2
```
petergphillips, line 13: had to bump this as integration tests failed twice in a row

mikehalmamoj (approved):

PR #1691, opened 2023-07-05
SDIT-915: ✨ Fix unilink permissions on cell move endpoint

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/service/BedAssignmentHistoryService.java
```diff
@@ -46,7 +46,7 @@ public BedAssignmentHistoryService(
         this.maxBatchSize = maxBatchSize;
     }
 
-    @VerifyBookingAccess
+    @VerifyBookingAccess(overrideRoles = {"SYSTEM_USER", "UNILINK"})
```
petergphillips, line 49: I'm a bit dubious about whether this is needed, but have left it in anyway.  It is only called from `saveAndReturnCellMoveResult`, which in turn is only called from either `moveToCell` or `moveToCellSwap` that are themselves protected

src/main/java/uk/gov/justice/hmpps/prison/service/BookingService.java
```diff
@@ -805,7 +805,7 @@ public void updateLivingUnit(final Long bookingId, final String livingUnitDescri
     }
 
     @Transactional
-    @VerifyBookingAccess
+    @VerifyBookingAccess(overrideRoles = {"SYSTEM_USER", "UNILINK"})
```
petergphillips, line 808: I'm a bit dubious about whether this is needed, but have left it in anyway.  It is only called from `saveAndReturnCellMoveResult`, which in turn is only called from either `moveToCell` or `moveToCellSwap` that are themselves protected

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderMovementsResourceIntTest_moveToCell.kt
```diff
@@ -129,11 +129,11 @@ class OffenderMovementsResourceIntTest_moveToCell : ResourceTest() {
     val response = requestMoveToCell(
       createJwt("unilink", listOf("ROLE_UNILINK")),
       OFFENDER_NO,
-      INITIAL_CELL_DESC,
+      NEW_CELL_DESC,
```
petergphillips, line 132: turns out my test was passing since it didn't then call `saveAndReturnCellMoveResult` so didn't go into the extra restrictions required.  Changed to actually move prisoner cell.

src/test/java/uk/gov/justice/hmpps/prison/util/JwtAuthenticationHelper.kt
```diff
@@ -27,20 +28,41 @@ class JwtAuthenticationHelper {
...
+    internalUser: Boolean = true,
+  ): String {
+    val claims = mutableMapOf<String, Any?>(
+      "client_id" to clientId,
+      "internalUser" to internalUser,
+    ).apply {
+      username?.let { this["user_name"] = username }
+      roles?.let {
+        // ensure that all roles have a ROLE_ prefix
+        this["authorities"] = roles.map { "ROLE_${it.substringAfter("ROLE_")}" }
```
petergphillips, line 57: tidied here to avoid confusion with role prefixes

src/main/java/uk/gov/justice/hmpps/prison/service/MovementUpdateService.java
```diff
@@ -46,7 +45,7 @@ public MovementUpdateService(
     }
 
     @Transactional
-    @VerifyBookingAccess(overrideRoles = {"ROLE_SYSTEM_USER", "ROLE_UNILINK"})
+    @VerifyBookingAccess(overrideRoles = {"SYSTEM_USER", "UNILINK"})
```
petergphillips, line 48: this doesn't make any difference, but removed the `ROLE_` prefix to conform to other usages

mikehalmamoj (approved):

PR #1697, opened 2023-07-10
NN-5281: update existing ADA Sanctions

neilmendum:
https://dsdmoj.atlassian.net/browse/NN-5281


src/main/java/uk/gov/justice/hmpps/prison/service/AdjudicationsService.java
```diff
@@ -530,15 +530,45 @@ private List<Sanction> transform(List<OicSanction> oicSanctions) {
...
+                        oicSanctionRepository.save(adaSanction);
+                        existingOicSanctionsToDeleteAndRecreate.remove(adaSanction);
+
+                        updatedSanctions.addAll(toDto(List.of(adaSanction, linkedSanctionForThisBookId)));
+                    }
+                });
+            }
+        }
+
+       if (!existingOicSanctionsToDeleteAndRecreate.isEmpty()) {
```
neilmendum, line 562: In the scenario where we've updated an existing `adaSanction` then we remove it from existingOicSanctionsToDeleteAndRecreate - [line 554](https://github.com/ministryofjustice/prison-api/pull/1697/files#diff-dadab9a8f82365cf5227ec804663b0d13e0c907b45bc0c110b93e84f21bc91b1R554) - and the collection is empty then we don't need to call `saveToDto`

timoj58 (approved):

PR #1699, opened 2023-07-12
CRS-1477 - Add a new data return value that contains enough information for a booking to have relevant dates calculated (a calculable booking) 

joelstobart-moj:
- Move from a Sentence Summary to a Calculable Sentence Envelope
- Add additional data elements to the Calculable Sentence Envelope


src/main/java/uk/gov/justice/hmpps/prison/api/resource/PrisonResource.kt
```diff
@@ -43,12 +43,12 @@ class PrisonResource(private val bookingService: BookingService) {
   )
   @Operation(summary = "Details of the active offender booking and calculable sentences at a particular establishment")
   @PreAuthorize("hasRole('RELEASE_DATE_MANUAL_COMPARER')")
-  @GetMapping("/{establishmentId}/booking/latest/sentence-summary")
-  fun getActiveOffenderBookingsByEstablishment(
+  @GetMapping("/{establishmentId}/booking/latest/calculable-sentence-envelope")
```
petergphillips, line 46: what is a sentence envelope?  If the endpoint is now changed to return that data instead then would be good to amend the swagger to provide the additional information about what it is

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/SentenceCalcType.java
```diff
@@ -64,4 +66,13 @@ public static class PK implements Serializable {
     @Convert(converter = YesNoConverter.class)
     private boolean active;
 
+    public Boolean isAFine() {
```
petergphillips, line 69: why the nullable result here, especially for an `is` method?  Would prefer
```
"A/FINE".equals(calculationType);
```

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/SentenceCalcType.java
```diff
@@ -64,4 +66,13 @@ public static class PK implements Serializable {
...
+        return Objects.equals(calculationType, "A/FINE");
+    }
+
+    public Boolean isRecallType(){
+        var licenceRecallTypes = new String[]{"LR", "LR_ORA", "LR_YOI_ORA", "LR_SEC91_ORA", "LRSEC250_ORA"};
+        return Arrays.stream(licenceRecallTypes).filter(Objects::nonNull).anyMatch(licenceRecallType -> licenceRecallType.equals(calculationType));
```
petergphillips, line 75: seems quite a convoluted way to find out if it a recall type, especially since `Objects::nonNull` will always be true since you've only just created the list on the previous line.  Would be better to just put the items into a `Set` instead and then you can do a `contains` rather than looping over all the values.

Would be also better to make the values into a constant called `RECALL_TYPES` or something.

src/main/java/uk/gov/justice/hmpps/prison/service/BookingService.java
```diff
@@ -852,15 +868,42 @@ public List<SentenceSummary> getActiveBookingsForEstablishment(String caseLoad){
...
+            offenderBooking.getOffender().getBirthDate()
+        );
+
+        var sentenceAdjustments = offenderBooking.getSentenceAdjustments();
+        var bookingAdjustments = offenderBooking.getBookingAdjustments();
+
+        var containsFine = offenderBooking.getSentences().stream().filter(Objects::nonNull).anyMatch(sentence -> sentence.getCalculationType().isAFine());
+        var containsRecall = offenderBooking.getSentences().stream().filter(Objects::nonNull).anyMatch(sentence -> sentence.getCalculationType().isRecallType());
+
+        List<OffenderFinePaymentDto> offenderFinePaymentDtoList = new ArrayList();
```
petergphillips, line 887: prefer variables to be immutable rather than defining and then redefining

src/main/java/uk/gov/justice/hmpps/prison/api/model/calculation/CalculableSentenceEnvelope.kt
```diff
@@ -0,0 +1,28 @@
...
+import io.swagger.v3.oas.annotations.media.Schema
+import uk.gov.justice.hmpps.prison.api.model.BookingAdjustment
+import uk.gov.justice.hmpps.prison.api.model.FixedTermRecallDetails
+import uk.gov.justice.hmpps.prison.api.model.OffenderFinePaymentDto
+import uk.gov.justice.hmpps.prison.api.model.SentenceAdjustmentValues
+import uk.gov.justice.hmpps.prison.api.model.SentenceSummary
+
+data class CalculableSentenceEnvelope(
+  @Schema(description = "The identifiers of a person necessary for a calculation")
+  var person: Person,
```
petergphillips, line 12: why is this a `var`?

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/SentenceCalcType.java
```diff
@@ -64,4 +66,13 @@ public static class PK implements Serializable {
     @Convert(converter = YesNoConverter.class)
     private boolean active;
 
+    public Boolean isAFine() {
+        return Objects.equals(calculationType, "A/FINE");
+    }
+
+    public Boolean isRecallType(){
```
petergphillips, line 73: again why a nullable `Boolean`?

src/main/java/uk/gov/justice/hmpps/prison/api/resource/PrisonResource.kt
petergphillips: are there any active consumers of this endpoint?

joelstobart-moj: None at present

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/SentenceCalcType.java
joelstobart-moj: Done :)

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/SentenceCalcType.java
joelstobart-moj: I've implemented this

mikehalmamoj (approved):

PR #1705, opened 2023-07-20
MAP-162 allowing to do internal move to reception

michalnawrockidj:


src/main/java/uk/gov/justice/hmpps/prison/api/model/InternalMoveResult.java
```diff
@@ -13,7 +13,7 @@
 @Data
 @AllArgsConstructor
 @NoArgsConstructor
-public class CellMoveResult {
+public class InternalMoveResult {
```
mikehalmamoj, line 16: This would be a breaking change for anyone that generates their client model from the Swagger docs. Have you checked if there is anyone doing this?

michalnawrockidj: I roll back this changes

andymarke (approved):
LGTM

PR #1709, opened 2023-07-24
remove role - used by front end - validation method only

timoj58:


timoj58:
role added to app instead

PR #1715, opened 2023-07-29
NON-155 Incorrect location of NA

Mjwillis:
Incorrect location of NA! - Location of non-association is returning location of linked booking when it should be latest booking


petergphillips (approved):
LGTM

PR #1725, opened 2023-08-24
CDPS-114: Return personal care need IDs

danielburnley:
The new prisoner profile aims to display reasonable adjustments alongside the personal care need it's related to - to accomplish this we need to return the ID for the association


petergphillips (approved):
LGTM

PR #1730, opened 2023-09-05
SDIT-1076 findOffenders can fail to find PNC and CRO

steverendell:

PR #1733, opened 2023-09-06
Revert "SDIT-1076 findOffenders can fail to find PNC and CRO"

steverendell:
Reverts ministryofjustice/prison-api#1730


petergphillips (approved):
LGTM

PR #1735, opened 2023-09-12
[CDPS-101] Return modified at datetimes for alerts

danielburnley:


src/main/java/uk/gov/justice/hmpps/prison/api/model/Alert.java
```diff
@@ -81,6 +82,10 @@ public class Alert {
     @JsonProperty("dateExpires")
     private LocalDate dateExpires;
 
+    @Schema(description = "Datetime the alert was last modified", example = "2020-08-20T12:34:56")
+    @JsonProperty("dateModified")
+    private LocalDateTime dateModified;
```
petergphillips, line 87: should really be called `dateTimeModified` since it doesn't just hold the date.  I'm not a great fan of prefixing the name with the type either, but that seems to be the way the others are done in this class.  Should really be `modifiedDateTime`.  Also would be good to include more information about the format in the `description` - see [CourtHearingDateAmendment](https://github.com/ministryofjustice/prison-api/blob/253475ab2362a7bc6452ca25391dc80931e21560/src/main/java/uk/gov/justice/hmpps/prison/api/model/CourtHearingDateAmendment.java#L22) for an example

src/test/java/uk/gov/justice/hmpps/prison/executablespecification/BookingStepDefinitions.java
```diff
@@ -499,4 +501,10 @@ private LocalDate convertTodayAndParse(final String dateCreated) {
         if (dateCreated.equals("today")) return LocalDate.now();
         return LocalDate.parse(dateCreated);
     }
+
+    private LocalDateTime convertTodayAndParseDateTime(final String dateTimeCreated) {
```
petergphillips, line 505: slightly odd field name here since it isn't always a creation date time.  Agree that you've copied the code above, but I'd prefer just a `dateTime` or even better `dateTimeAsString`

src/main/java/uk/gov/justice/hmpps/prison/api/model/Alert.java
```diff
@@ -81,6 +82,10 @@ public class Alert {
     @JsonProperty("dateExpires")
     private LocalDate dateExpires;
 
+    @Schema(description = "Datetime the alert was last modified", example = "2020-08-20T12:34:56")
```
petergphillips, line 85: not sure if this should be `RequiredMode.NOT_REQUIRED` too - can the field be `null`?

src/test/resources/features/booking_alerts.feature
```diff
@@ -51,13 +51,13 @@ Feature: Booking Alerts
...
+      | offenderNo | bookingId | alertId | alertType | alertTypeDescription | alertCode | alertCodeDescription    | comment               | dateCreated | dateExpires | expired | active | dateModified        |
+      | A1234AF    | -6        | 1       | P         | MAPPP Case           | P1        | MAPPA Level 1 Case      | Alert Text 6          | today       |             | false   | true   | 2006-12-10T03:52:25 |
+      | A1234AF    | -6        | 2       | X         | Security             | XTACT     | XTACT                   | Alert XTACT 6         | today       |             | false   | true   | 2006-12-10T03:52:25 |
+      | A1234AA    | -1        | 1       | X         | Security             | XA        | Arsonist                | Alert Text 1-1        | 2020-06-01  |             | false   | true   | 2006-12-10T03:52:25 |
+      | A1234AA    | -1        | 2       | H         | Self Harm            | HC        | Self Harm - Custody     | Alert Text 1-2        | 2020-06-01  |             | false   | true   | 2006-12-10T03:52:25 |
```
petergphillips, line 58: would be good to add some `null` entries if field is not required?

petergphillips (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/repository/sql/InmateAlertRepositorySql.kt
```diff
@@ -10,6 +10,7 @@ enum class InmateAlertRepositorySql(val sql: String) {
         ALERT_CODE,
         COALESCE(alCode.DESCRIPTION, ALERT_CODE) as ALERT_CODE_DESC,
         EXPIRY_DATE,
+        MODIFY_DATETIME,
```
petergphillips, line 13: this column in ambiguous and needs the `OA.` prefix - `STAFF_USER_ACCOUNTS` also has a column called `MODIFY_DATETIME`

src/main/java/uk/gov/justice/hmpps/prison/repository/sql/InmateAlertRepositorySql.kt
```diff
@@ -42,6 +43,7 @@ enum class InmateAlertRepositorySql(val sql: String) {
...
```
petergphillips, line 46: same here

PR #1736, opened 2023-09-14
ADJUST1-172 adding offence statute to offence resource.

ldlharper:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderCharge.java
```diff
@@ -149,6 +149,7 @@ public OffenderOffence getOffenceDetail() {
             .offenderChargeId(id)
             .offenceStartDate(dateOfOffence)
             .offenceEndDate(endDate)
+            .offenceStatute(offence.getStatute().getCode())
```
petergphillips, line 152: can this ever be null?

ldlharper: No, the statute forms part of the offence primary key.

petergphillips (approved):
LGTM

PR #1742, opened 2023-09-27
SDIT-1105 Add GLOBAL_SEARCH and VIEW_PRISONER_DATA as override roles to match other calls

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/service/InmateService.java
```diff
@@ -415,7 +415,7 @@ public List<OffenderIdentifier> getOffenderIdentifiers(final Long bookingId, @Nu
                 .toList();
     }
 
-    @VerifyBookingAccess(overrideRoles = {"SYSTEM_USER"})
+    @VerifyBookingAccess(overrideRoles = {"SYSTEM_USER", "GLOBAL_SEARCH", "VIEW_PRISONER_DATA"})
```
karenmillermoj, line 418: This is needed to allow us to remove the SYSTEM_USER role from our endpoints and client creds

mikehalmamoj (approved):

PR #1744, opened 2023-10-02
SDIT-1109 Remove default overrideRole of SYSTEM_USER

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/security/AuthenticationFacade.java
```diff
@@ -55,8 +56,7 @@ public static boolean hasRoles(final String... allowedRoles) {
     }
 
     public boolean isOverrideRole(final String... overrideRoles) {
```
karenmillermoj, line 58: We no longer allow default of SYSTEM_USER.  If no overrideRoles are passed in, nothing will match

src/main/java/uk/gov/justice/hmpps/prison/service/SearchOffenderService.java
```diff
@@ -51,7 +51,7 @@ public Page<OffenderBooking> findOffenders(final SearchOffenderRequest request)
         Objects.requireNonNull(request.getLocationPrefix(), "locationPrefix is a required parameter");
         log.info("Searching for offenders, criteria: {}", request);
 
-        final Set<String> caseloads = securityUtils.isOverrideRole() ? Set.of() : userService.getCaseLoadIds(request.getUsername());
+        final Set<String> caseloads = userService.getCaseLoadIds(request.getUsername());
```
karenmillermoj, line 54: 5 clients call the endpoint (/api/locations/description/{locationPrefix}/inmates) that calls this method.  Only categorisation-tool client has the SYSTEM_USER role, and it has not made a call to this endpoint since Jan 2023

andymarke (approved):
LGTM

PR #1746, opened 2023-10-02
SDIT-1120 Override check of SYSTEM_USER role is now deprecated

karenmillermoj:
This is called from POST /api/bookings/offenders endpoint

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/service/InmateService.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/service/InmateService.java
@@ -164,7 +164,7 @@ public class InmateService {
     }
 
     private boolean isViewAllOffenders() {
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/justice/hmpps/prison/service/InmateService.java
```diff
@@ -164,7 +164,7 @@ public List<InmateBasicDetails> getBasicInmateDetailsForOffenders(final Set<Stri
     }
 
     private boolean isViewAllOffenders() {
-        return authenticationFacade.isOverrideRole("SYSTEM_USER", "GLOBAL_SEARCH", "VIEW_PRISONER_DATA");
```
karenmillermoj, line 167: isViewAllOffenders() has been called by 6 clients in the last year.  5 do not have the SYSTEM_USER role and the 6th client (categorisation-tool) passes through the USER details.

mikehalmamoj (approved):

PR #1753, opened 2023-10-04
SDIT-1125 Remove SYSTEM_USER from first set of VerifyOffenderAccess methods

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/api/resource/EducationResource.java
```diff
@@ -43,7 +43,7 @@ public class EducationResource {
         @ApiResponse(responseCode = "500", description = "Unrecoverable error occurred whilst processing request.", content = {@Content(mediaType = "application/json", schema = @Schema(implementation = ErrorResponse.class))})})
     @Operation(summary = "A list of offender educations.", description = "A list of offender educations.")
     @GetMapping("/prisoner/{offenderNo}")
-    @PreAuthorize("hasAnyRole('SYSTEM_USER','GLOBAL_SEARCH', 'VIEW_PRISONER_DATA')")
```
karenmillermoj, line 46: Removed from these 3 resource files as they call the services that have had the SYSTEM_USER role removed from the overrideRole.

PR #1766, opened 2023-10-16
SDIT-1148 Throw 403 error if call using client credentials only fails due to incorrect access

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/service/BookingService.java
```diff
@@ -609,6 +610,7 @@ public void verifyBookingAccess(final Long bookingId, final String... rolesAllow
         if (agencyIds.isEmpty()) {
             if (authenticationFacade.isClientOnly()) {
                 logClientUnauthorisedAccess(bookingId, rolesAllowed);
+                throw new AccessDeniedException(format("Client not authorised to access booking with id %d.", bookingId));
```
karenmillermoj, line 613: I originally created a new exception for this, but I think this is fine (especially as I have left in logging/customEvent)

andymarke (approved):
lgtm

PR #1769, opened 2023-10-17
CRS-1500 added get calculable sentence envelope by offender nos

carloveo-moj:
This is the second way a bulk calculate release dates comparison happens. This is modelled the same as https://github.com/ministryofjustice/prison-api/blob/main/src/main/java/uk/gov/justice/hmpps/prison/api/resource/PrisonResource.kt#L47 but retrieving active bookings by a set of offender numbers instead of establishment id

### DIFF
--- a/src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/BookingResourceIntTest_getCalculableSentenceEnvelope.kt
+++ b/src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/BookingResourceIntTest_getCalculableSentenceEnvelope.kt
@@ -1,0 +1,73 @@
+@file:Suppress("ClassName")
+
+package uk.gov.justice.hmpps.prison.api.resource.impl
+
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import com.fasterxml.jackson.module.kotlin.readValue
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/BookingResourceIntTest_getCalculableSentenceEnvelope.kt
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.core.ParameterizedTypeReference
+import org.springframework.http.MediaType
+import uk.gov.justice.hmpps.prison.api.model.calculation.CalculableSentenceEnvelope
+
+class BookingResourceIntTest_getCalculableSentenceEnvelope : ResourceTest() {
+
+  @Autowired
+  val objectMapper = jacksonObjectMapper()
```
mikehalmamoj, line 16: I guess this isn't actually autowired so maybe drop the annotation? Surprised it works tbh 😄

mikehalmamoj (approved):

mikehalmamoj (dismissed):

mikehalmamoj (approved):

PR #1771, opened 2023-10-18
SDIT-1125 Refactor update case note tests

karenmillermoj:


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/BookingCaseNotesResourceIntTest.kt
```diff
@@ -0,0 +1,311 @@
...
+  inner class GetCaseNotes {
+    @Test
+    fun testCanRetrieveCaseNotes() {
+      webTestClient.get()
+        .uri("/api/bookings/-2/caseNotes")
+        .headers(setAuthorisation("ITAG_USER", listOf("")))
+        .exchange()
+        .expectStatus().isOk
+        .expectBody()
+        .json("case_notes_offender_1.json".readFile())
```
mikehalmamoj, line 34: Has this been converted from Java? Just wondering if this would be a good opportunity to to switch to `jsonPath` instead of comparing with a json file?

andymarke (approved):
lgtm

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/BookingCaseNotesResourceIntTest.kt
karenmillermoj: It has been converted from java - and I agree - this should be updated too to pull out the bits we want to check

PR #1772, opened 2023-10-19
SDIT-1158 Add internal location count to establishment roll api

steverendell:


src/main/java/uk/gov/justice/hmpps/prison/api/model/RollCount.java
```diff
@@ -117,6 +125,16 @@ public void setCurrentlyInCell(final Integer currentlyInCell) {
         this.currentlyInCell = currentlyInCell;
     }
 
+    @Schema(requiredMode = REQUIRED, description = "No of residential prisoners in internal locations")
```
petergphillips, line 128: is this description correct - doesn't seem to correspond to the field name

petergphillips (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/api/model/RollCount.java
steverendell: yep, its just another way of putting it.

PR #1782, opened 2023-10-20
SDIT-1125 Remove VerifyBookingAccess from service when getting casenote as already on controller

karenmillermoj:


andymarke (approved):
LGTM

PR #1785, opened 2023-10-20
SDIT-1125 Remove unused endpoint

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/api/resource/ImageResource.java
```diff
@@ -78,19 +78,6 @@ public List<ImageDetail> getImagesByOffender(@PathVariable("offenderNo") final S
         return imageService.findOffenderImagesFor(offenderNo);
     }
 
-    @GetMapping("/offenders")
```
karenmillermoj, line 81: This endpoint was used by the dps data compliance team and hasn't been accessed since July 2023.  The client no longer exists in prod and checked with Jon Brighton who was happy for the endpoint to be removed

petergphillips (approved):
LGTM

PR #1794, opened 2023-10-24
SDIT-1125 Remove override role SYSTEM_USER from endpoint as only user caseload access required

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/service/InmateAlertService.java
```diff
@@ -79,7 +79,7 @@ public Page<Alert> getInmateAlerts(final Long bookingId, final String orderBy, f
         return alerts;
     }
 
-    @VerifyBookingAccess(overrideRoles = {"SYSTEM_USER"})
```
karenmillermoj, line 82: empty override roles is now possible.  The only client using this is prison-staff-hub which does not have SYSTEM_USER and so we can remove the override

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/BookingResourceIntTest_getOffenderAlert.kt
```diff
@@ -0,0 +1,109 @@
+@file:Suppress("ClassName")
+
+package uk.gov.justice.hmpps.prison.api.resource.impl
+
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+
+@DisplayName("GET /api/bookings/{bookingId}/alerts/{alertId}")
+class BookingResourceIntTest_getOffenderAlert : ResourceTest() {
```
karenmillermoj, line 10: Most tests copied from cucumber tests

petergphillips (approved):
LGTM

PR #1795, opened 2023-10-25
SDIT-1109 Remove SYSTEM_USER role from endpoint only called with client without that role

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/api/resource/BookingMovementsResource.java
```diff
@@ -113,7 +113,7 @@ public CourtHearings getCourtHearings(@PathVariable("bookingId") @Parameter(desc
             @ApiResponse(responseCode = "500", description = "Unrecoverable error occurred whilst processing request.", content = {@Content(mediaType = "application/json", schema = @Schema(implementation = ErrorResponse.class))})})
     @PutMapping("/{bookingId}/living-unit/{internalLocationDescription}")
     @ProxyUser
-    @VerifyBookingAccess(overrideRoles = {"SYSTEM_USER", "MAINTAIN_CELL_MOVEMENTS"})
```
karenmillermoj, line 116: This endpoint is only called by whereabouts-api-client which does not have SYSTEM_USER role

steverendell (approved):
lgtm

PR #1798, opened 2023-10-25
SDIT-1180 OffenderActivitiesResource

steverendell:


src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderActivitiesResource.java
```diff
@@ -30,6 +31,7 @@
 @Tag(name = "offender-activities")
 @Validated
 @RequestMapping(value = "${api.base.path}/offender-activities", produces = "application/json")
+@PreAuthorize("hasRole('NOMIS_ACTIVITIES')")
```
steverendell, line 34: I thought Id reuse this role name as it is accessing similar information

petergphillips (approved):
LGTM

PR #1807, opened 2023-10-30
CRS-1678 added alerts to calculable sentence envelope

carloveo-moj:


src/test/resources/uk/gov/justice/hmpps/prison/api/resource/impl/prison_resource_fixed_recall_calculable_sentence_envelope.json
```diff
@@ -1,7 +1,35 @@
 {
   "person": {
     "prisonerNumber": "A1234AE",
-    "dateOfBirth": "1956-02-28"
+    "dateOfBirth": "1956-02-28",
+    "alerts": [{
```
mikehalmamoj, line 5: I guess there's no change to any tests because they are asserting against this json?

mikehalmamoj (approved):

src/test/resources/uk/gov/justice/hmpps/prison/api/resource/impl/prison_resource_fixed_recall_calculable_sentence_envelope.json
carloveo-moj: that's right, the test can be found https://github.com/ministryofjustice/prison-api/blob/main/src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/PrisonResourceTest.kt#L16-L38

PR #1809, opened 2023-10-31
SDIT-1195 PrisonerStatusController and /offenders/ids

steverendell:


src/main/java/uk/gov/justice/hmpps/prison/api/resource/PrisonerStatusController.java
```diff
@@ -46,6 +35,7 @@ public class PrisonerStatusController {
             @ApiResponse(responseCode = "200", description = "OK", content = {@Content(mediaType = "application/json", schema = @Schema(implementation = PrisonerInformation.class))}),
             @ApiResponse(responseCode = "401", description = "Unauthorized.", content = {@Content(mediaType = "application/json", schema = @Schema(implementation = ErrorResponse.class))}),
             @ApiResponse(responseCode = "404", description = "User not found.", content = {@Content(mediaType = "application/json", schema = @Schema(implementation = ErrorResponse.class))})})
+    @VerifyOffenderAccess(overrideRoles = {"SYSTEM_USER", "GLOBAL_SEARCH", "VIEW_PRISONER_DATA"})
```
mikehalmamoj, line 38: Should we be adding SYSTEM_USER here? I thought it was deprecated...

karenmillermoj: We should only be adding it if we're moving it (as in this case).  The only alternative would be to remove the role if the endpoint has been checked and all associated calling clients don't have that role

mikehalmamoj: Oh sorry it's move from the service. I see, thanks 😄

mikehalmamoj (approved):

src/main/java/uk/gov/justice/hmpps/prison/api/resource/PrisonerStatusController.java
steverendell: Im just leaving it as it is (or was , on the service layer)

PR #1812, opened 2023-10-31
SDIT-1201 Provide alternative override role for /api/staff/roles/{agencyId}/role/{role} and refactor associated cucumber tests

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/service/AgencyService.java
```diff
@@ -211,11 +215,24 @@ public void verifyAgencyAccess(final String agencyId) {
         if (AuthenticationFacade.hasRoles("INACTIVE_BOOKINGS")) {
             agencyIds.addAll(Set.of("OUT", "TRN"));
         }
+        if (agencyIds.isEmpty()) {
+            if (authenticationFacade.isClientOnly()) {
+                logClientUnauthorisedAccess(agencyId);
+            }
+            throw EntityNotFoundException.withId(agencyId);
```
karenmillermoj, line 222: This is similar to the one for bookings.  The overrideRoles is not passed into this method, but if we need it - we can easily pass this through for logging

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/StaffResourceIntTest.kt
```diff
@@ -0,0 +1,186 @@
...
+
+  @SpyBean
+  protected lateinit var telemetryClient: TelemetryClient
+
+  @Nested
+  @DisplayName("GET /api/staff/roles/{agencyId}/role/{role}")
+  inner class StaffRoles {
+    @Test
+    fun `should return not found if agency does not exist for user`() {
+      webTestClient.get()
```
karenmillermoj, line 25: The majority of tests were copied over from the cucumber tests - with additional ones for the role checks

steverendell (approved):
lgtm!

PR #1814, opened 2023-11-01
SDIT-1203 Remove SYSTEM_USER override role from /api/schedules/{agencyId}/events-by-location-ids

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/service/SchedulesService.java
```diff
@@ -72,7 +72,7 @@ public SchedulesService(final LocationService locationService,
         this.maxBatchSize = maxBatchSize;
     }
 
-    @VerifyAgencyAccess(overrideRoles = {"SYSTEM_USER"})
```
karenmillermoj, line 75: The endpoint that calls this method is only ever called by prisonstaffhub which does not have SYSTEM_USER

andymarke (approved):
lgtm

PR #1816, opened 2023-11-01
SDIT-1204 Remove unnecessary VerifyBookingAccess from POST /api/bookings//{bookingId}/personal-care-needs

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/service/HealthService.java
```diff
@@ -109,7 +109,6 @@ public List<PersonalCareCounterDto> countPersonalCareNeedsByOffenderNoAndProblem
     }
 
     @Transactional
-    @VerifyBookingAccess(overrideRoles = {"SYSTEM_USER"})
```
karenmillermoj, line 112: Only `welcome-people-into-prison-client` calls the POST api/bookings/bookingId/personal-care-needs endpoint The endpoint has PreAuthorize with role ROLE_MAINTAIN_HEALTH_PROBLEMS and VerifyBookingAccess with override role SYSTEM_USER. Therefore VerifyBookingAccess always returns true, so the PreAuthorize is ensuring access is valid and SYSTEM_USER is superfluous - so can be removed

mikehalmamoj (approved):

PR #1821, opened 2023-11-03
SDIT-1211 Remove unused override role for /api/cell/{agencyId}/history/{assignmentDate}

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/service/BedAssignmentHistoryService.java
```diff
@@ -87,7 +87,7 @@ public List<BedAssignment> getBedAssignmentsHistory(final long livingUnitId, fin
             .toList();
     }
 
-    @VerifyAgencyAccess(overrideRoles = {"SYSTEM_USER"})
```
karenmillermoj, line 90: The endpoint which calls this service is only called by prison-staff-hub which uses AuthorizationCode - so the overrideRoles can safely be removed

mikehalmamoj (approved):

PR #1843, opened 2023-11-10
SDIT-1228 Add alternative roles for InmateService categorisation calls

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/service/InmateService.java
```diff
@@ -359,6 +359,7 @@ private void setAlertsFields(final InmateDetail inmate) {
      * @param bookingId tacit
      * @return latest assessment of each code for the offender
      */
+    // @VerifyBookingAccess(overrideRoles = {"SYSTEM_USER", "VIEW_ASSESSMENTS"})
```
karenmillermoj, line 362: Added placeholders ready for update

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderAssessmentResourceIntTest.kt
```diff
@@ -149,9 +151,92 @@ class OffenderAssessmentResourceIntTest : ResourceTest() {
   }
 
   @Nested
-  @DisplayName("POST /api/offender-assessments/category/LEI?latest=false\"")
-  inner class Category {
+  @DisplayName("GET /api/offender-assessments/category/{agencyId}")
+  inner class GetOffenderCategorisations {
 
+    @Nested
```
karenmillermoj, line 157: Added authorisation tests for changed endpoint roles

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderAssessmentResourceIntTest.kt
```diff
@@ -363,6 +448,149 @@ class OffenderAssessmentResourceIntTest : ResourceTest() {
...
+            }
+            """,
+          )
+          .exchange()
+          .expectStatus().isUnauthorized
+      }
+
+      @Test
+      fun `should return 403 if does not have authorised role`() {
+        webTestClient.post().uri("/api/offender-assessments/category/categorise")
```
mikehalmamoj, line 474: At line 168 there's an endpoint that returns a 404 if no authorised role. Just wondering what the difference is with this endpoint whcih returns 403?

mikehalmamoj (approved):

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderAssessmentResourceIntTest.kt
karenmillermoj: Good question - there is another PR coming thru - this is because it is using VerifyAgencyAccess which hasn't yet been promoted/updated to change the EntityNotFound to an AccessDenied (only VerifyBookingAccess/VerifyOffenderAccess have been updated so far).  This test will enable me to check for a 403 once the update has been made.

PR #1853, opened 2023-11-14
SDIT-1245 VIEW_INCIDENTS now authorised to access /api/offenders/{offenderNo}/incidents

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/service/IncidentService.java
```diff
@@ -11,19 +11,16 @@
...
         this.repository = repository;
-        this.bookingService = bookingService;
     }
 
     public IncidentCase getIncidentCase(@NotNull final long incidentCaseId) {
         return repository.getIncidentCases(List.of(incidentCaseId)).stream().findFirst().orElseThrow(EntityNotFoundException.withId(incidentCaseId));
     }
 
     public List<IncidentCase> getIncidentCasesByOffenderNo(@NotNull final String offenderNo, final List<String> incidentTypes, final List<String> participationRoles) {
-        bookingService.getOffenderIdentifiers(offenderNo, "SYSTEM_USER");
```
karenmillermoj, line 26: Preauthorize - with SYSTEM_USER and VIEW_INCIDENTS on controller satisfies access, so this is not required here (only offender-risk-profiler calls this which is a client-creds only call)

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceIntTest.kt
```diff
@@ -0,0 +1,757 @@
...
+        httpEntity,
+        object : ParameterizedTypeReference<String?>() {},
+        "A1234AI",
+      )
+      assertThatJsonFileAndStatus(response, 200, "offender_detail_aliases.json")
+    }
+  }
+
+  @Nested
+  @DisplayName("GET api/offenders/{offenderNo}/incidents")
```
karenmillermoj, line 250: OffenderIncidents class tests access to the updated endpoint (all others are just pulled in from the java class)

petergphillips (approved):
LGTM

PR #1862, opened 2023-11-15
SDIT-1148 Throw 403 error if call using client credentials only fails due to incorrect agency access

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/service/AgencyService.java
```diff
@@ -218,10 +220,12 @@ public void verifyAgencyAccess(final String agencyId) {
         if (agencyIds.isEmpty()) {
             if (authenticationFacade.isClientOnly()) {
                 logClientUnauthorisedAccess(agencyId);
+                throw new AccessDeniedException(format("Client not authorised to access agency with id %s.", agencyId));
```
petergphillips, line 223: this message seems slightly odd - clients aren't restricted by `agencyId` anyway?

petergphillips (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/service/AgencyService.java
karenmillermoj: If they don't have the override role then the client isn't allowed to access the method if it has VerifyAgencyAccess

petergphillips (approved):
LGTM

PR #1865, opened 2023-11-16
SDIT-1244 Remove unused endpoint /api/bookings/{bookingId}/activities/today

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/service/BookingService.java
```diff
@@ -331,7 +331,6 @@ private List<ScheduledEvent> getBookingActivities(final Collection<Long> booking
         return bookingRepository.getBookingActivities(bookingIds, fromDate, toDate, sortFields, sortOrder);
     }
 
-    @VerifyBookingAccess(overrideRoles = {"SYSTEM_USER"})
```
karenmillermoj, line 334: This method can not be removed as it is called by two other methods in BookingService which already have @VerifyBookingAccess set, therefore the annotation on this method can be removed

petergphillips (approved):
LGTM

PR #1871, opened 2023-11-17
CDPS-413: Add scheduled transfers endpoint

davidmiddletonmoj:


src/main/java/uk/gov/justice/hmpps/prison/api/resource/ScheduleResource.java
```diff
@@ -203,4 +203,11 @@ public List<PrisonerSchedule> getCourtEvents(@PathVariable("agencyId") @Paramete
     public List<PrisonerSchedule> getExternalTransfers(@PathVariable("agencyId") @Parameter(required = true) final String agencyId, @RequestBody @Parameter(description = "The required offender numbers (mandatory)", required = true) final List<String> body, @RequestParam("date") @DateTimeFormat(iso = ISO.DATE) @Parameter(description = "Date of scheduled transfer") final LocalDate date) {
         return schedulesService.getExternalTransfers(agencyId, body, date);
     }
+
+    @Operation
+    @GetMapping("/{prisonerNumber}/scheduledTransfers")
+    @SlowReportQuery
+    public List<PrisonerSchedule> getScheduledTransfersForPrisoner(@PathVariable("prisonerNumber") @Parameter(required = true) final String prisonerNumber) {
```
andymarke, line 210: It doesn't look like this endpoint is protected by any roles, am I missing some code somewhere to protect this, I suspect schedule transfers would be *very* sensitive data in high security prisons.

(I'm struggling to see how any of these endpoints are protected on this resource so I must be missing something )

src/main/java/uk/gov/justice/hmpps/prison/repository/ScheduleRepository.kt
```diff
@@ -206,6 +206,20 @@ class ScheduleRepository : RepositoryBase() {
     return schedules.toPrisonerSchedules()
   }
 
+  fun getScheduledTransfersForPrisoner(
+    prisonerNumber: String?,
+  ): List<PrisonerSchedule> {
```
andymarke, line 211: very minor but you may as well drop the optional ? given it can't get here an be null

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/ScheduleResourceTest.kt
```diff
@@ -365,6 +365,25 @@ class ScheduleResourceTest : ResourceTest() {
...
   }
 
+  @Nested
+  @DisplayName("GET /api/schedules/{prisonerNumber}/scheduledTransfers")
+  inner class ScheduledTransfers {
+
+    @Test
+    fun testThatScheduledTransfer_IsReturned() {
+      val token = authTokenHelper.getToken(NORMAL_USER)
+      val response: ResponseEntity<List<PrisonerSchedule>> = testRestTemplate.exchange(
```
andymarke, line 375: prefer web test client over deprecated testRestTemplate - but I understand if you chosen this to be consistent with the other tests for this resource

andymarke (dismissed):
LGTM other than the possible role issue

src/main/java/uk/gov/justice/hmpps/prison/api/resource/ScheduleResource.java
davidmiddletonmoj: That's a fair point and other than some references to `@VerifyAgencyAccess` in the service, doesn't appear to be any protection. This is my first time updating PrisonAPI, so not sure where else to look for confirmation this is correct. I basically copied the similar `getExternalTransfers`

andymarke (approved):
LGTM

PR #1874, opened 2023-11-20
SDIT-1244 Add tests and convert test file to kotlin

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/service/BookingService.java
```diff
@@ -534,7 +534,6 @@ public OffenderBookingIdSeq getOffenderIdentifiers(final String offenderNo, fina
         return offenderIdentifier;
     }
 
-    @VerifyBookingAccess(overrideRoles = {"SYSTEM_USER"})
```
karenmillermoj, line 537: Not required as this is now only called from 2 places which both have the VerifyBookingAccess annotation

andymarke (approved):
lgtm

src/main/java/uk/gov/justice/hmpps/prison/repository/BookingRepository.java
```diff
@@ -515,25 +515,6 @@ private static class OffenderBookingSeq {
         private Integer bookingSeq;
     }
 
-    public Page<ScheduledEvent> getBookingAppointments(final Long bookingId, final LocalDate fromDate, final LocalDate toDate, final long offset, final long limit, final String orderByFields, final Order order) {
```
andymarke, line 518: so where has this code gone?

karenmillermoj: It's was only called from endpoints that are not used.  It was called from my last PR (no longer used) but I forgot to remove it

PR #1881, opened 2023-11-21
SDIT-1258 Remove override role when updating case notes and add tests

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/service/CaseNoteService.java
```diff
@@ -179,7 +179,7 @@ public CaseNote createCaseNote(String offenderNo, @NotNull @Valid @CaseNoteTypeS
     }
 
     @Transactional
-    @VerifyOffenderAccess(overrideRoles = {"SYSTEM_USER"})
```
karenmillermoj, line 182: This is only called by prison-staff-hub and whereabouts-api which do not have SYSTEM_USER role, so therefore can be removed.

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceIntTest.kt
```diff
@@ -714,6 +723,211 @@ class OffenderResourceIntTest : ResourceTest() {
...
+      webTestClient.put().uri("/api/offenders/A1234AP/case-notes/34")
+        .headers(setAuthorisation("ITAG_USER", listOf("")))
+        .header("Content-Type", APPLICATION_JSON_VALUE)
+        .body(BodyInserters.fromValue(caseNoteUpdate))
+        .exchange()
+        .expectStatus().isNotFound
+        .expectBody().jsonPath("userMessage").isEqualTo("Resource with id [A1234AP] not found.")
+    }
+
+    @Test
```
karenmillermoj, line 768: Pulled in these tests from the old update case note by bookingId (which has now been removed as it is unused)

PR #1885, opened 2023-11-23
SDIT-549 Discharge prisoner integration tests

mikehalmamoj:


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceIntTest_dischargeToHospital.kt
```diff
@@ -0,0 +1,727 @@
...
+
+    @Nested
+    @DisplayName("when discharge to hospital is a success")
+    @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+    inner class SideEffects {
+
+      @ParameterizedTest
+      @MethodSource("uk.gov.justice.hmpps.prison.api.resource.impl.OffenderResourceIntTest_dischargeToHospital#getOffenderTypes")
+      fun `should update booking`(offenderType: OffenderType) {
+        createOffender(offenderType)
```
mikehalmamoj, line 271: There is no equivalent to `@MethodSource` on `@BeforeEach` so the setup has to be done in each test separately.

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceIntTest_dischargeToHospital.kt
```diff
@@ -0,0 +1,727 @@
...
+
+        // check bed assignments not ended (yet) - not relevant if no booking or released
+        if (listOf(BOOKING, TEMPORARY_ABSENCE).contains(offenderType)) {
+          testDataContext.getBedAssignments(bookingId!!)
+            .maxBy { it.assignmentDate }
+            .also {
+              assertThat(it.assignmentEndDate).isNull()
+            }
+        }
```
mikehalmamoj, line 369: We have a few of these were the pre-checks don't apply to some offender types... which isn't ideal but didn't want to duplicate the various tests if possible.

andymarke (approved):
LGTM - a work or art

PR #1888, opened 2023-11-23
SDIT-1220: 🎨 Switch smoke code to kotlin

petergphillips:


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/SmokeTestHelperResourceIntTest.kt
```diff
@@ -0,0 +1,131 @@
...
+import java.time.temporal.ChronoUnit
+
+class SmokeTestHelperResourceIntTest : ResourceTest() {
+  @Autowired
+  private lateinit var repository: OffenderImprisonmentStatusRepository
+
+  @Test
+  @DisplayName("requires ROLE_SMOKE_TEST")
+  fun requiresCorrectRole() {
+    webTestClient.post()
```
petergphillips, line 28: switch to web test client here too

PR #1901, opened 2023-11-29
CRS-1709 Add new paginated end point to replace existing unpaged version (for retrieving all Calculable Sentences for bookings in a establishment)

yasinmustafa:
This functionality is used by the Calculate Release Dates service, the deprecated endpoint will be removed once the new endpoint has been integrated and tested with Calculate Release Dates.

The new endpoint has the same URL as the deprecated one - except the word `paged` has been added


joelstobart-moj (approved):
LGTM

PR #1902, opened 2023-11-29
SDIT-1287 Switch unilink get-transaction endpoints to primary

steverendell:
Use primary for all GETs involved in the Unilink issue


petergphillips (approved):
LGTM

PR #1908, opened 2023-12-01
SDIT-549 discharge to hospital

mikehalmamoj:


src/main/java/uk/gov/justice/hmpps/prison/api/model/RequestToDischargePrisoner.kt
```diff
@@ -0,0 +1,36 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude
+import io.swagger.v3.oas.annotations.media.Schema
+import jakarta.validation.constraints.NotNull
+import org.hibernate.validator.constraints.Length
+import org.springframework.format.annotation.DateTimeFormat
+import java.time.LocalDateTime
+
+@Schema(description = "Request release of prisoner")
+@JsonInclude(JsonInclude.Include.NON_NULL)
+class RequestToDischargePrisoner(
```
mikehalmamoj, line 12: Apart from conversion to Kotlin the only change to this request object is to drop the comment text - it isn't used anyway.

src/main/java/uk/gov/justice/hmpps/prison/service/enteringandleaving/DischargeToHospitalService.kt
```diff
@@ -48,7 +49,9 @@ class DischargeToHospitalService(
     return if (!lastMovement.isRelease()) {
       dischargeToHospital(offenderNo, dischargeTime, toLocation)
     } else {
+      // this is to support the "Migrate into Restricted Patients" process where an offender was released in NOMIS instead of being moved to hospital in Restricted Patients
       lastMovement.changeToHospitalDischarge(toLocation)
+      caseNoteMovementService.createReleaseNote(offenderBooking, lastMovement)
```
mikehalmamoj, line 54: Create another case note when we are migrating a released prisoner into Restricted Patients - so that it is consistent with other discharge to hospital use cases (and nicer for users because this case note includes the hospital whereas the release case note might not)

src/main/java/uk/gov/justice/hmpps/prison/service/enteringandleaving/CaseNoteMovementService.kt
```diff
@@ -68,7 +68,7 @@ class CaseNoteMovementService(
       .type(type)
       .subType(subType)
       .caseNoteText(note)
-      .agencyLocation(booking.location)
+      .agencyLocation(booking.createLocation)
```
mikehalmamoj, line 71: When creating a case note for a released prisoner being migrated into Restricted patients the location would be OUT - we want the prison instead

steverendell (approved):
lgtm

PR #1936, opened 2023-12-11
SDIT-1298 Update test builders to use new services

andymarke:
Refactored the test builders in the OffenderBookingBuilder unit to use the newer transfer, prison creation and deletion services. This bypasses the webClient calls.

Since we no longer need the webTestClient the builders can now be injected.


src/test/java/uk/gov/justice/hmpps/prison/util/builders/dsl/NomisData.kt
```diff
@@ -1,16 +1,27 @@
...
-  testDataContext: TestDataContext,
-) {
-  private val offenderBuilderFactory: OffenderBuilderFactory = OffenderBuilderFactory(testDataContext = testDataContext)
-  fun build(dsl: NomisData.() -> Unit) = NomisData(
-    offenderBuilderFactory,
-  ).apply(dsl)
+@Component
+class NomisDataBuilder(private val offenderBuilderFactory: OffenderBuilderFactory) {
+  fun build(dsl: NomisData.() -> Unit) {
+    SecurityContextHolder.setContext(
```
andymarke, line 13: this would effectively be done via the webClient call, but since we are bypassing this set the context directly with a hardcoded user (with no roles). For now, this seems "good enough" so long as the test data doesn't require a specific user - or the service makes any further authentication calls.

mikehalmamoj: Nice. I think there will be some user based tests coming for which we could just make user oart if the top level DSL.

mikehalmamoj (approved):

mikehalmamoj (approved):

PR #1942, opened 2023-12-11
CDPS-590 Improve performance of Staff role lookup

Mjwillis:
Switch to jpa entity for this query and remove need to subjoin to get max result in date


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/StaffJobRole.kt
```diff
@@ -0,0 +1,101 @@
...
+  @Id
+  @Column(name = "FROM_DATE", nullable = false)
+  val fromDate: LocalDate,
+
+  @Id
+  @Column(name = "POSITION", nullable = false)
+  val position: String,
+
+  @Id
+  @Column(name = "ROLE", nullable = false)
```
Mjwillis, line 44: Had to map twice as formula not allowed as PK

src/main/java/uk/gov/justice/hmpps/prison/service/StaffJobType.kt
```diff
@@ -0,0 +1,5 @@
+package uk.gov.justice.hmpps.prison.service
+
+enum class StaffJobType {
+  KW, POM
```
Mjwillis, line 4: Currently only two types are actually checked

mikehalmamoj (approved):

src/main/java/uk/gov/justice/hmpps/prison/api/resource/StaffResource.java
```diff
@@ -107,4 +109,24 @@ public List<StaffRole>  getAllRolesForAgency(@PathVariable("staffId") @Parameter
...
+                                 @Parameter(description = "The staff id of the staff member.", required = true, example = "1111111")
+                                 @NotNull final Long staffId,
+                                 @PathVariable("agencyId")
+                                 @Parameter(description = "Agency Id.", required = true, example = "MDI")
+                                 @NotNull final String agencyId,
+                                 @PathVariable("roleType")
+                                 @Parameter(description = "Type of role", required = true, example = "KW")
+                                 @NotNull final StaffJobType roleType
+    ) {
+        return staffService.hasStaffRole(staffId, agencyId, roleType);
```
andymarke, line 128: Looks like this is missing roles

PR #1944, opened 2023-12-12
SDIT-1298 Half of transfer test done with court case, team and hearin…

andymarke:
…g additions to DSL.

### DIFF
--- a/src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceIntTest_transferIn.kt
+++ b/src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceIntTest_transferIn.kt
@@ -23,6 +23,7 @@ import org.springframework.security.test.context.support.WithMockUser
 import org.springframework.test.web.reactive.server.StatusAssertions
 import org.springframework.web.reactive.function.BodyInserters
 import uk.gov.justice.hmpps.prison.api.model.CaseNote
+import uk.gov.justice.hmpps.prison.dsl.NomisDataBuilder
 import uk.gov.justice.hmpps.prison.exception.CustomErrorCodes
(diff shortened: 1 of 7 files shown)


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceIntTest_transferIn.kt
```diff
@@ -287,134 +288,145 @@ class OffendersResourceTransferImpTest : ResourceTest() {
...
-        offenderNo =
-          OffenderBuilder().withBooking(
-            OffenderBookingBuilder(
-              prisonId = "LEI",
-              bookingInTime = LocalDateTime.now().minusDays(1),
-            ),
-          ).save(testDataContext).offenderNo
-      }
+      @Nested
+      inner class WhenInTransit {
```
andymarke, line 292: separated out to 2 nested test

- One for when has been transferred out
- The other when they haven't

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceIntTest_transferIn.kt
```diff
@@ -742,7 +740,7 @@ class OffendersResourceTransferImpTest : ResourceTest() {
                 BedAssignmentHistory::getAssignmentEndDate,
               )
               .containsExactly(
-                tuple("ADM", bookingInTime.toLocalDate(), LocalDate.now()),
+                tuple("ADM", bookingInTime.toLocalDate(), transferOutDateTime.toLocalDate()),
                 tuple("19", transferOutDateTime.toLocalDate(), null),
```
andymarke, line 744: this tests seems wrong, and I think it relied on data by the builder *not* being set correctly, since the above indicated the bed would have been released when they transferred out

mikehalmamoj (approved):

PR #1950, opened 2023-12-14
SDIT-1321 Remove SYSTEM_USER from categorisation endpoints

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/service/InmateService.java
```diff
@@ -516,7 +516,7 @@ public List<OffenderCategorise> getCategory(final String agencyId, final Categor
         };
     }
 
-    @VerifyBookingAccess(overrideRoles = {"SYSTEM_USER", "MAINTAIN_ASSESSMENTS"})
```
karenmillermoj, line 519: These endpoints do not have SYSTEM_USER in the resource file PreAuthorize method so is already tested and can therefore be removed

mikehalmamoj (approved):

PR #1952, opened 2023-12-14
SDIT-1306 Remove SYSTEM_USER from inmateAlerts

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/service/InmateAlertService.java
```diff
@@ -63,7 +63,6 @@ public InmateAlertService(
         this.maxBatchSize = maxBatchSize;
     }
 
-    @VerifyBookingAccess(overrideRoles = {"SYSTEM_USER", "GLOBAL_SEARCH", "VIEW_PRISONER_DATA"})
```
karenmillermoj, line 66: This method is only called as part of offenderDetails - and that endpoint has now been completed (and the overrideRoles added to the Resource file).  This annotation therefore be removed

lbennett-moj (approved):

PR #1975, opened 2024-01-09
CRS-1779: Update NOMIS key dates with reason

dave-virden:
Allows the reason for calculation in NOMIS to be updated with a reason selected by the CRDS user.


steverendell (approved):
lgtm

PR #1979, opened 2024-01-12
CDPS-607: Add more data to CSRA assessment endpoint

danielburnley:
On the new prisoner profile CSRA pages we're looking to include more information which is currently available in NOMIS but not available in DPS for viewing - these fields are the ones added in this PR. 

Some of these fields are conditionally returned already
- the classification codes are returned via the existing `classificationCode` field
- the override reason/approval comment both get returned via `classificationReviewReason`

However as it already has internal logic in the API we were unable to display all of the fields to the user at once if required for a better understanding. Rather than update the logic in those fields (which may be used/depended on elsewhere) i opted to add new fields instead.

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/api/model/AssessmentDetail.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/api/model/AssessmentDetail.java
@@ -9,6 +9,7 @@ import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
 
 import jakarta.validation.constraints.NotNull;
+
 import java.time.LocalDate;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/justice/hmpps/prison/service/OffenderAssessmentService.java
```diff
@@ -64,6 +64,11 @@ public AssessmentDetail getOffenderAssessment(final Long bookingId, final Intege
             .originalClassificationCode((classificationSummary.getOriginalClassification() != null)?classificationSummary.getOriginalClassification().getCode(): null)
             .classificationReviewReason(classificationSummary.getClassificationApprovalReason())
             .questions(getCsraAssessmentQuestionsAndAnswers(assessmentDetails, bookingId, assessmentSeq))
+            .overridingClassificationCode((assessmentDetails.getOverridingClassification() !=null)?assessmentDetails.getOverridingClassification().getCode():null)
+            .calculatedClassificationCode((assessmentDetails.getCalculatedClassification() !=null)?assessmentDetails.getCalculatedClassification().getCode():null)
+            .approvedClassificationCode((assessmentDetails.getReviewedClassification() !=null)?assessmentDetails.getReviewedClassification().getCode():null)
+            .approvalComment((assessmentDetails.getReviewCommitteeComment() !=null)?assessmentDetails.getReviewCommitteeComment():null)
```
petergphillips, line 70: pointless ternary operator

src/main/java/uk/gov/justice/hmpps/prison/api/model/AssessmentDetail.java
```diff
@@ -61,6 +69,21 @@ public AssessmentDetail(final AssessmentSummary summary, final String assessment
     @Schema(description = "The reason for the review of the classification", example = "HI")
     private String classificationReviewReason;
 
+    @Schema(description = "Overriding classification code", example = "HI")
+    private String overridingClassificationCode;
+
+    @Schema(description = "Originally calculated classification code", example = "HI")
```
petergphillips, line 75: this feels confusing.  This is the originally calculated classification code, which is different from the `originalClassificationCode`.  I don't know enough about the CSRA process to know what the better description could be though. Does the calculation happen after review - I noticed that `reviewedClassification` is not being returned here.  Would be good to include more information in the descriptions to help people understand what is going on.

danielburnley: Yeah it's all a bit confusing - The `originalClassificationCode` only gets populated when something has been approved and it was different to the one that was approved and is used in the CSRA Summaries as well. That can be useful for seeing at a glance if it was changed, but for a more detailed CSRA view you'd want to see both the original and approved specifically (hence the adding of the field).

I can update the description for them to the best of my knowledge though yeah - for reference, the `calculatedClassificationCode` is one that's automatically calculated by the system based on the questions answered in the initial review

petergphillips (approved):
LGTM

PR #1980, opened 2024-01-12
SDIT-1126 prison-api startup takes a lot of cpu

steverendell:
Set very low Tiered JIT compilation thresholds

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -1,5 +1,8 @@
 #!/bin/sh
 exec java ${JAVA_OPTS} \
+  -XX:+PrintCompilation -XX:TieredStopAtLevel=4 -XX:Tier3CompileThreshold=1 -XX:Tier4CompileThreshold=1 \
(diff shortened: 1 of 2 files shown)


run.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/sh
 exec java ${JAVA_OPTS} \
+  -XX:+PrintCompilation -XX:TieredStopAtLevel=4 -XX:Tier3CompileThreshold=1 -XX:Tier4CompileThreshold=1 \
+  -XX:Tier3InvocationThreshold=1 -XX:Tier4InvocationThreshold=1 \
+  -XX:Tier3MinInvocationThreshold=1 -XX:Tier4MinInvocationThreshold=1 \
```
steverendell, line 5: This turns on almost immediate JIT compilation. The theory is that most will be complete by the time traffic is let into the pod. Any remaining compilation (in endpoint app level code) will be done much earlier resulting in very few invocations on interpreted code.

mikehalmamoj (approved):

PR #1986, opened 2024-01-16
SDIT-1408: 🎨 Upgrade to latest ktlint

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenceIndicatorRepository.kt
```diff
@@ -4,6 +4,6 @@ import org.springframework.data.repository.CrudRepository
 import uk.gov.justice.hmpps.prison.repository.jpa.model.OffenceIndicator
 
 interface OffenceIndicatorRepository : CrudRepository<OffenceIndicator, Long> {
-  fun deleteByIndicatorCodeAndOffence_Code(indicatorCode: String, offenceCode: String): Long
-  fun existsByIndicatorCodeAndOffence_Code(indicatorCode: String, offenceCode: String): Boolean
+  fun deleteByIndicatorCodeAndOffenceCode(indicatorCode: String, offenceCode: String): Long
```
petergphillips, line 7: lots of these - don't need any of the `_` in the names

mikehalmamoj: I wonder if we should relax this rule? It makes some of the longer repository method names more readable. (though not in this case!)

mikehalmamoj (approved):

PR #1990, opened 2024-01-18
SDIT-1126 prison-api startup takes a lot of cpu

steverendell:
Set one pod per node (5) in prod and tune memory and cpu

### DIFF
--- a/helm_deploy/prison-api/values.yaml
+++ b/helm_deploy/prison-api/values.yaml
@@ -3,7 +3,7 @@
 generic-service:
   nameOverride: prison-api
   productId: DPS060
-  replicaCount: 6
+  replicaCount: 5 # One per node
 
   image:
     repository: quay.io/hmpps/prison-api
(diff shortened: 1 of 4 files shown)


helm_deploy/prison-api/values.yaml
```diff
@@ -3,7 +3,7 @@
 generic-service:
   nameOverride: prison-api
   productId: DPS060
-  replicaCount: 6
+  replicaCount: 5 # One per node
```
steverendell, line 6: The number of nodes available to run pods on has increased from 3 to 5

helm_deploy/prison-api/values.yaml
```diff
@@ -27,14 +27,14 @@ generic-service:
 
   resources:
     requests:
-      cpu: 100m
-      memory: 640Mi
+      cpu: 1000m
+      memory: 2048Mi
```
steverendell, line 31: Make these values more realistic (requests is minimum required)

helm_deploy/values-dev.yaml
```diff
@@ -6,6 +6,14 @@ generic-service:
...
     enabled: true
 
+  resources:
+    requests:
+      cpu: 100m
+      memory: 1024Mi
+    limits:
+      cpu: 2000m
+      memory: 2048Mi
+
```
steverendell, line 16: Convention is for values.yaml to contain production values - so smaller limits go in the env-specific config

petergphillips (approved):
LGTM

PR #1998, opened 2024-01-30
SDIT-1451 Remove SYSTEM_USER from InmateService.getInmatesAssessmentsByCode

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/service/InmateService.java
```diff
@@ -412,9 +412,9 @@ public List<Assessment> getInmatesAssessmentsByCode(final List<String> offenderN
                                                         final boolean mostRecentOnly) {
         final List<Assessment> results = new ArrayList<>();
         if (!CollectionUtils.isEmpty(offenderNos)) {
-            final Set<String> caseLoadIds = authenticationFacade.isOverrideRole("SYSTEM_USER", "VIEW_ASSESSMENTS", "VIEW_PRISONER_DATA")
+            final Set<String> caseLoadIds = authenticationFacade.isOverrideRole("VIEW_ASSESSMENTS", "VIEW_PRISONER_DATA")
                 ? Collections.emptySet()
-                : caseLoadService.getCaseLoadIdsForUser(authenticationFacade.getCurrentUsername(), false);
+                : loadCaseLoadsOrThrow();
```
karenmillermoj, line 417: This change was needed to ensure the following scenarios were captured (as was previously returning data)
- client creds with no user and no override role
- user in context with no user roles
However - this throws a 400 (bad request), I used this as it was being used elsewhere, but wasn't sure if we should be throwing a 403

andymarke (approved):
LGTM

PR #1999, opened 2024-01-31
Updated annotations for from agency

anthony-britton-moj:
From agency can be null, swagger docs indicated it was non-null. This should update the swagger docs to make it clear that the field can be null.


andymarke (approved):
lgtm

PR #2005, opened 2024-02-02
SDIT-1358 Remove SYSTEM_USER from /api/offenders/{offenderNo}/adjudications

karenmillermoj:


src/test/resources/features/offender_adjudications.feature
```diff
@@ -14,7 +14,7 @@ Feature: Offenders Adjudications
     And the associated agencies for this offender are: "LEI,MDI"
 
   Scenario: A staff user views many adjudications for an existing offender
-    Given a user has a token name of "SYSTEM_USER_READ_WRITE"
```
karenmillermoj, line 17: Have just swapped these tests to use an alternative override role (rather than rewriting tests) as this functionality has now been removed from Nomis.  All calling clients will shortly be moving over to the DPS service and so these tests can then be removed.

lbennett-moj (approved):
lgtm

PR #2010, opened 2024-02-12
CMG-19 Add new endpoint to get the next-court-event for an offender (based on passed in bookingId) - protected by the RELEASE_DATES_CALCULATOR role

yasinmustafa:


andymarke (approved):
LGTM

PR #2011, opened 2024-02-14
SDIT-1335 Implement hmpps-kotlin-spring-boot-starter lib

mikehalmamoj:


src/main/java/uk/gov/justice/hmpps/prison/security/AuthenticationFacade.java
```diff
@@ -49,8 +50,10 @@ public Collection<? extends GrantedAuthority> getCurrentRoles() {
     public boolean isClientOnly() {
         final var authentication = getAuthentication();
         if (!(authentication instanceof AuthAwareAuthenticationToken)) return false;
+        final var client = ((AuthAwareAuthenticationToken) authentication).getClientId();
+        final var principal = ((AuthAwareAuthenticationToken) authentication).getPrincipal();
 
-        return ((AuthAwareAuthenticationToken) authentication).isClientOnly();
+        return client.equals(principal);
```
mikehalmamoj, line 56: This is slightly different to the old code in the deleted local version of AuthAwareAuthenticationConverter which checks that the JWT's subject and client_id are the same. We now compare principal and client_id because we know that they are the same only if the userName is null.

src/main/java/uk/gov/justice/hmpps/prison/security/AuthenticationFacade.java
```diff
@@ -60,13 +63,6 @@ public String getClientId() {
         return (String)((AuthAwareAuthenticationToken) authentication).getClientId();
     }
 
-    public String getGrantType() {
-        final var authentication = getAuthentication();
-        if (!(authentication instanceof AuthAwareAuthenticationToken)) return null;
-
-        return (String)((AuthAwareAuthenticationToken) authentication).getGrantType();
-    }
```
mikehalmamoj, line 68: Not used anywhere other than in tests

petergphillips (approved):
LGTM

PR #2018, opened 2024-02-21
MAP-494 Refactor capacity check methods in AgencyInternalLocation

Mjwillis:
This commit removes the boolean parameter 'treatZeroOperationalCapacityAsNull' from the capacity check methods in AgencyInternalLocation class. Now the methods consider operational capacity as zero when it is null. The related tests and class methods that use those methods have also been updated accordingly.


andymarke (approved):
LGTM

PR #2027, opened 2024-02-27
SDIT-1531 Return 403 for GET /api/bookings/{bookingId} and /api/bookings/{bookingId}/cell-history if user does not have access

karenmillermoj:


petergphillips (approved):
LGTM

PR #2029, opened 2024-02-27
SDIT-1536 Return 403 for GET /api/bookings/{bookingId}/caseNotes/{type}/{subType}/count if user does not have access

karenmillermoj:


petergphillips (approved):
LGTM

PR #2034, opened 2024-02-28
CRS-1976: Update calculation history query to return all calculations…

justin-gardiner:
… for an offender. Remove dates from  calculation history return value and add agency details.


src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderDatesResource.java
```diff
@@ -84,7 +84,7 @@ public ResponseEntity<OffenderCalculatedKeyDates> getOffenderKeyDates(@PathVaria
     @GetMapping("/calculations/{nomsId}")
     @PreAuthorize("hasRole('RELEASE_DATES_CALCULATOR')")
     @ProxyUser
-    public ResponseEntity<List<OffenderSentenceCalculation>> getOffenderCalculations(@PathVariable("nomsId") @Parameter(description = "The booking id of offender", required = true) final String nomsId) {
+    public ResponseEntity<List<SentenceCalculationSummary>> getOffenderCalculations(@PathVariable("nomsId") @Parameter(description = "The booking id of offender", required = true) final String nomsId) {
```
justin-gardiner, line 87: this isn't currently so ok to change

PR #2042, opened 2024-03-04
SDIT-1553 Return 403 for POST /api/offenders/{offenderNo}/case-notes if user does not have access

karenmillermoj:


petergphillips (approved):
LGTM

PR #2044, opened 2024-03-04
SDIT-1566: 🐛 Fix offender repository query

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderRepository.kt
```diff
@@ -19,11 +17,8 @@ interface OffenderRepository : JpaRepository<Offender, Long> {
 
   fun findByLastNameAndFirstNameAndBirthDate(lastName: String, firstName: String, dob: LocalDate): List<Offender>
 
-  @Query("select o from Offender o left join fetch o.bookings b WHERE o.nomsId = :nomsId order by b.bookingSequence asc")
-  fun findOffendersByNomsId(@Param("nomsId") nomsId: String, pageable: Pageable): List<Offender>
-
-  fun findOffenderByNomsId(nomsId: String): Optional<Offender> =
-    findOffendersByNomsId(nomsId, PageRequest.of(0, 1)).stream().findFirst()
+  @Query("select o from Offender o left join fetch o.bookings b WHERE o.nomsId = :nomsId and o.id = o.rootOffenderId")
```
petergphillips, line 20: odd one - if we are going to only get one entry then should be the root one.  Same sql as for `findOffenderByNomsIdOrNullForUpdate` below

andymarke: @petergphillips 100% agree - I was talking to karen about the exact thing last week  - I was just too scared to change this in case it broke something

PR #2047, opened 2024-03-04
SDIT-1551 In-cell appointments not working

steverendell:
Filter out locations with no description


src/main/java/uk/gov/justice/hmpps/prison/repository/sql/AgencyRepositorySql.kt
```diff
@@ -127,44 +127,44 @@ enum class AgencyRepositorySql(val sql: String) {
...
                         FROM OFFENDER_VISITS VIS
                         WHERE VIS.START_TIME BETWEEN :periodStart AND :periodEnd
                 AND VIS.VISIT_DATE = TRUNC(:periodStart)
-        AND VIS.AGY_LOC_ID = :agencyId
-        )
-        )
+              AND VIS.AGY_LOC_ID = :agencyId
+           )
+          )
         ORDER BY USER_DESCRIPTION
```
steverendell, line 168: Just correcting formatting. The random indentation really confused what I thought the results of this query should be 🙄

src/main/java/uk/gov/justice/hmpps/prison/service/AgencyService.java
```diff
@@ -297,7 +297,10 @@ private List<Location> getAgencyLocationsOnDayAndPeriod(final String agencyId, @
 
         final var locations = agencyRepository.getAgencyLocationsBooked(agencyId, bookedOnDay, bookedOnPeriod);
         final var processedLocations = LocationProcessor.processLocations(locations, true);
-        return processedLocations.stream().sorted(LOCATION_DESCRIPTION_COMPARATOR).toList();
+        return processedLocations.stream()
+            .filter(location -> StringUtils.isNotBlank(location.getUserDescription()))
+            .sorted(LOCATION_DESCRIPTION_COMPARATOR)
+            .toList();
```
steverendell, line 303: This endpoint returns location ids with just the user description if there is one, in both the userDescription and description (not nullable) fields. If there isnt a user description, the issue is that the client gets a json error because description is missing. So the fix is to just omit these values. They are caused by in-cell bookings because cells have no user description. And it is ok for cell location not to be mentioned in the activities page which uses this.

src/main/java/uk/gov/justice/hmpps/prison/service/AgencyService.java
```diff
@@ -389,24 +392,24 @@ private boolean isBlankAddress(final PrisonContactDetail pcd) {
     public List<OffenderCell> getCellsWithCapacityInAgency(@NotNull final String agencyId, final String attribute) {
         final var cells = agencyInternalLocationRepository.findWithProfilesAgencyInternalLocationsByAgencyIdAndLocationTypeAndActive(agencyId, "CELL", true);
         return cells.stream()
-            .filter((l) -> l.isActiveCellWithSpace())
-            .map(cell -> transform(cell, true))
+            .filter(AgencyInternalLocation::isActiveCellWithSpace)
+            .map(this::transform)
             .filter(cell -> attribute == null || cell.getAttributes().stream().anyMatch((a) -> a.getCode().equals(attribute)))
```
steverendell, line 397: Intellij recommendations

src/main/resources/db/migration/nomis/ddl/V1_402__AGY_INT_LOC_PROFILES.sql
```diff
@@ -12,6 +12,8 @@ CREATE TABLE AGY_INT_LOC_PROFILES (
   AUDIT_CLIENT_USER_ID VARCHAR2(64),
   AUDIT_CLIENT_IP_ADDRESS VARCHAR2(39),
   AUDIT_CLIENT_WORKSTATION_NAME VARCHAR2(64),
-  AUDIT_ADDITIONAL_INFO VARCHAR2(256)
+  AUDIT_ADDITIONAL_INFO VARCHAR2(256),
+  CONSTRAINT AGY_INT_LOC_PROFILES_PK primary key (INTERNAL_LOCATION_ID, INT_LOC_PROFILE_TYPE, INT_LOC_PROFILE_CODE),
+  CONSTRAINT AGY_INT_LOC_PROF_AGY_INT_LOC_F FOREIGN KEY (INTERNAL_LOCATION_ID)  REFERENCES AGENCY_INTERNAL_LOCATIONS (INTERNAL_LOCATION_ID)
```
steverendell, line 17: Retrofitted constraints from the 'real' database

src/main/java/uk/gov/justice/hmpps/prison/service/AgencyService.java
mikehalmamoj: This feels like a bit of a hack. If the location has been used for an event then pretending it doesn't exist because it has a blank user description is surely wrong - it's being used after all? Better to change the OpenAPI specifiation to allow blank description and the client can ignore it if they really want to?

mikehalmamoj (approved):

src/main/java/uk/gov/justice/hmpps/prison/service/AgencyService.java
steverendell: Yes definitely, Ill be raising a ticket to overhaul the endpoint.

PR #2052, opened 2024-03-06
SDIT-1579: 🐛 Fix sentence query and upgrade hibernate

petergphillips:
The original queries had two issues:
1. They were meant to return offender bookings, but these would sometimes be duplicated (if multiple sentences exist) as the query was eager fetching with a left outer join.  See https://developer.jboss.org/docs/DOC-15782#jive_content_id_Hibernate_does_not_return_distinct_results_for_a_query_with_outer_join_fetching_enabled_for_a_collection_even_if_I_use_the_distinct_keyword for more information
2. Pagination was specified with eager fetching.  This means that hibernate had to load the whole prison data into memory before then doing the paging.  See https://stackoverflow.com/questions/11431670/how-can-i-avoid-the-warning-firstresult-maxresults-specified-with-collection-fe

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderBookingId.kt
+++ b/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderBookingId.kt
@@ -1,0 +1,5 @@
+package uk.gov.justice.hmpps.prison.repository.jpa.repository
+
+interface OffenderBookingId {
+  val bookingId: Long
+}
--- a/src/main/java/uk/gov/justice/hmpps/prison/service/BookingService.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/service/BookingService.java
@@ -82,6 +82,7 @@ import uk.gov.justice.hmpps.prison.repository.jpa.model.VisitVisitor;
 import uk.gov.justice.hmpps.prison.repository.jpa.repository.AgencyLocationRepository;
 import uk.gov.justice.hmpps.prison.repository.jpa.repository.CourtEventRepository;
 import uk.gov.justice.hmpps.prison.repository.jpa.repository.OffenderBookingFilter;
+import uk.gov.justice.hmpps.prison.repository.jpa.repository.OffenderBookingId;
 import uk.gov.justice.hmpps.prison.repository.jpa.repository.OffenderBookingRepository;
 import uk.gov.justice.hmpps.prison.repository.jpa.repository.OffenderChargeRepository;
 import uk.gov.justice.hmpps.prison.repository.jpa.repository.OffenderContactPersonsRepository;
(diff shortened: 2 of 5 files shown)


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderBookingId.kt
```diff
@@ -0,0 +1,5 @@
+package uk.gov.justice.hmpps.prison.repository.jpa.repository
+
+interface OffenderBookingId {
```
petergphillips, line 3: this seemed the nicest way to just get the list of booking ids without loading the whole `OffenderBooking` for each record returned

src/main/java/uk/gov/justice/hmpps/prison/service/BookingService.java
```diff
@@ -848,12 +849,13 @@ public Optional<SentenceSummary> getSentenceSummary(final String offenderNo) {
     public Page<CalculableSentenceEnvelope> getCalculableSentenceEnvelopeByEstablishment(String caseLoad, int pageNumber, int pageSize) {
         final var agencyLocation = agencyLocationRepository.getReferenceById(caseLoad);
         PageRequest pageRequest = PageRequest.of(pageNumber, pageSize, Sort.by("bookingId"));
-        final var activeBookings = offenderBookingRepository.findAllOffenderBookingsByActiveTrueAndLocationAndSentences_statusAndSentences_CalculationType_CalculationTypeNotLikeAndSentences_CalculationType_CategoryNot(
+        final var bookingIds = offenderBookingRepository.findDistinctByActiveTrueAndLocationAndSentences_statusAndSentences_CalculationType_CalculationTypeNotLikeAndSentences_CalculationType_CategoryNot(
             agencyLocation, "A", "%AGG%", "LICENCE", pageRequest
         );
+        final var activeBookings = offenderBookingRepository.findAllByBookingIdIn(bookingIds.stream().map(OffenderBookingId::getBookingId).toList());
```
petergphillips, line 855: we split into two - first one to get the paged booking ids and second to retrieve (with eager fetching) each of those bookings.

src/main/java/uk/gov/justice/hmpps/prison/service/BookingService.java
```diff
@@ -848,21 +849,23 @@ public Optional<SentenceSummary> getSentenceSummary(final String offenderNo) {
...
         final var calculableSentenceEnvelopes = activeBookings.stream().map(this::determineCalculableSentenceEnvelope).toList();
 
-        return new PageImpl<>(calculableSentenceEnvelopes, pageRequest, activeBookings.getTotalElements());
+        return new PageImpl<>(calculableSentenceEnvelopes, pageRequest, bookingIds.getTotalElements());
     }
 
     public List<CalculableSentenceEnvelope> getCalculableSentenceEnvelopeByOffenderNumbers(Set<String> offenderNumbers) {
-        final var activeBookings = offenderBookingRepository.findAllOffenderBookingsByActiveTrueAndOffenderNomsIdInAndSentences_statusAndSentences_CalculationType_CalculationTypeNotLikeAndSentences_CalculationType_CategoryNot(
+        final var bookingIds = offenderBookingRepository.findDistinctByActiveTrueAndOffenderNomsIdInAndSentences_statusAndSentences_CalculationType_CalculationTypeNotLikeAndSentences_CalculationType_CategoryNot(
```
petergphillips, line 862: had to split the query into two here - first one without the entity graph to get the booking ids (so don't run into the hibernate bug) and then the second to get the data

src/test/java/uk/gov/justice/hmpps/prison/repository/OffenderBookingRepositoryTest.kt
```diff
@@ -138,19 +137,29 @@ class OffenderBookingRepositoryTest {
...
+    val ids = repository.findDistinctByActiveTrueAndLocationAndSentences_statusAndSentences_CalculationType_CalculationTypeNotLikeAndSentences_CalculationType_CategoryNot(
       agencyLocation,
       "A",
       "BOB",
       "JOE",
-      Pageable.unpaged(),
+      PageRequest.of(0, 2, Sort.by("bookingId")),
     )
+    assertThat(ids.content).hasSize(2)
+    assertThat(ids.totalElements).isGreaterThanOrEqualTo(20)
```
petergphillips, line 160: previously total elements was 24, with new distinct it is 22.  I've done greater than 20 though in case other tests add some more data

mikehalmamoj (approved):

PR #2054, opened 2024-03-07
SDIT-1579: 🎨 Separate out service and convert to kotlin

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/service/BookingService.java
```diff
@@ -826,10 +822,6 @@ public List<OffenderSentenceAndOffences> getSentenceAndOffenceDetails(final Long
     }
 
     public List<OffenderFinePaymentDto> getOffenderFinePayments(final Long bookingId) {
-        return getOffenderFinePaymentsFromRepository(bookingId);
```
petergphillips, line 829: can't see why needed two methods here when the first one doesn't do anything

src/main/java/uk/gov/justice/hmpps/prison/service/SentenceEnvelopeService.kt
```diff
@@ -0,0 +1,122 @@
...
+    val bookingIds =
+      offenderBookingRepository.findDistinctByActiveTrueAndLocationAndSentences_statusAndSentences_CalculationType_CalculationTypeNotLikeAndSentences_CalculationType_CategoryNot(
+        agencyLocation,
+        "A",
+        "%AGG%",
+        "LICENCE",
+        pageRequest,
+      )
+    val activeBookings =
+      offenderBookingRepository.findAllByBookingIdIn(bookingIds.stream().map(OffenderBookingId::bookingId).toList())
```
petergphillips, line 50: I've done a very simple conversion to kotlin - will fix the `.stream()` etc in next PR

mikehalmamoj (approved):

PR #2060, opened 2024-03-14
SDIT-1304 Fix disabled tests and tidy up

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/api/resource/BookingResource.java
```diff
@@ -142,7 +142,6 @@ public class BookingResource {
         @ApiResponse(responseCode = "500", description = "Unrecoverable error occurred whilst processing request.", content = {@Content(mediaType = "application/json", schema = @Schema(implementation = ErrorResponse.class))})})
     @Operation(summary = "Prisoners Booking Summary ", description = "Returns data that is available to the users caseload privileges, at least one attribute of a prisonId, bookingId or offenderNo must be specified")
     @GetMapping("/v2")
-    @VerifyBookingAccess(overrideRoles = {"VIEW_PRISONER_DATA"})
```
karenmillermoj, line 145: This wasn't doing anything - for verifyBookingAccess it needs to pass in  "bookingId".  
bookingIds are not mandatory and "bookingIds" is passed in - a list

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/BookingResourceIntTest.kt
```diff
@@ -189,21 +188,21 @@ class BookingResourceIntTest : ResourceTest() {
...
     @Test
-    @Disabled("this test fails - code update needed")
-    fun `should return 403 if does not have override role`() {
+    fun `should return empty list if client does not have override role`() {
       webTestClient.get().uri("/api/bookings/v2?prisonId=BXI")
         .headers(setClientAuthorisation(listOf()))
         .exchange()
-        .expectStatus().isForbidden
+        .expectStatus().isOk
+        .expectBody().jsonPath("totalElements").isEqualTo(0)
```
karenmillermoj, line 196: Have agreed with Andy M that due to the sql filtering the data - this was acceptable

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderFinePaymentResourceTest.kt
```diff
@@ -2,29 +2,27 @@ package uk.gov.justice.hmpps.prison.api.resource.impl
 
 import org.assertj.core.api.Assertions
 import org.junit.jupiter.api.Test
-import org.springframework.http.MediaType
```
karenmillermoj, line 5: changes in this file are just a tidy up

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderBookingFilter.java
```diff
@@ -62,7 +62,6 @@ public Predicate toPredicate(@NotNull final Root<OffenderBooking> root, @NotNull
 
         Optional
             .ofNullable(caseloadIds)
-            .filter(java.util.function.Predicate.not(List::isEmpty))
```
karenmillermoj, line 65: Removal of this line was needed to allow for a user with no caseloads.  Otherwise a user with no caseloads had the same result as an "override role" which meant everything was returned

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/BookingResourceIntTest.kt
```diff
@@ -212,14 +211,25 @@ class BookingResourceIntTest : ResourceTest() {
         .headers(setClientAuthorisation(listOf("ROLE_VIEW_PRISONER_DATA")))
         .exchange()
         .expectStatus().isOk
+        .expectBody().jsonPath("totalElements").isEqualTo(1)
     }
 
     @Test
-    @Disabled("this test fails - code update needed")
-    fun `returns 403 if prison is not in user caseload`() {
+    fun `returns empty list if prison is not in user caseload`() {
```
mikehalmamoj, line 218: Is this consistent with other endpoints? Wrong caseload and no override role feels like a "not authorised" scenario - 403. Or maybe even a 404 if that's what other endpoints do?

karenmillermoj: yeah - had a long chat with Andy about it - because the work to determine if a user can view an offender (in this case via prisonId or bookingId or offenderNo) is done in the sql (@programmaticAuthorisation annotation), it was agreed this is acceptable.  Otherwise we would be going down a long road to fix all.  I am happy for the team to discuss this as I can see your point, but it's a difficult one... 
should we first test access and only then do the sql call?  If we do go down that route, we probably need to change all @programmaticAuthorisation annotations that Steve added on as part of checking authorisation as they all work in the same way. Looking in prison-api - there are 24 occurrences of this.

mikehalmamoj (approved):

PR #2069, opened 2024-03-26
CDPS-690: Update CSRA Override behaviour

danielburnley:
Following advice from security collegues the way the CSRA asessement is displayed has been changed. Instead of overrides returning a blank CSRA rating until approved - it now displays the higher of the two in case of it being HI until approved, at which point it will display the final result.


steverendell (approved):
lgtm

PR #2071, opened 2024-03-27
SDIT-1610: ✨ Add in repository method to lock bookings

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/service/enteringandleaving/PrisonerCreationService.kt
```diff
@@ -125,8 +125,8 @@ class PrisonerCreationService(
 
   private fun validPncNumber(pncNumber: String?): Result<String>? =
     pncNumber?.takeIf { it.isNotBlank() }?.let { pnc ->
-      return Pnc.getShortPncNumber(pnc).checkForDuplicatePncNumber().map {
-        return Pnc.getLongPncNumber(pnc).checkForDuplicatePncNumber()
+      Pnc.getShortPncNumber(pnc).checkForDuplicatePncNumber().flatMap {
+        Pnc.getLongPncNumber(pnc).checkForDuplicatePncNumber()
```
petergphillips, line 129: tidying - removed unnecessary returns

src/test/java/uk/gov/justice/hmpps/prison/dsl/Alias.kt
```diff
@@ -25,6 +25,7 @@ class AliasBuilderRepository(
     offenderRepository.save(
       Offender.builder()
         .nomsId(it.nomsId)
+        .rootOffenderId(it.rootOffenderId)
```
petergphillips, line 28: the `rootOffender` is mapped as `insertable=false` and `updateable=false`, the mapping is handled by the `rootOffenderId` here instead so needs to be set

steverendell (approved):
lgtm

PR #2076, opened 2024-03-28
SDIT-1542 Return 403 for GET /api/offender-assessments/csra/{bookingId}/assessment/{assessmentSeq} if user does not have access

karenmillermoj:


andymarke (approved):
lgtm

PR #2077, opened 2024-03-28
SDIT-1556 Return 403 for GET /api/bookings/{bookingId}/visits-with-visitors if user does not have access

karenmillermoj:


andymarke (approved):
lgtm

PR #2078, opened 2024-03-28
SDIT-1542 Return 403 for GET /api/offender-sentences/booking/{bookingId}/sentences-and-offences if user does not have access

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/service/BookingService.java
```diff
@@ -790,7 +790,6 @@ public void updateLivingUnit(final Long bookingId, final AgencyInternalLocation
         updateLivingUnit(offenderBooking, location);
     }
 
-    @VerifyBookingAccess(overrideRoles = {"VIEW_PRISONER_DATA"})
     public List<OffenderSentenceAndOffences> getSentenceAndOffenceDetails(final Long bookingId) {
```
steverendell, line 793: afair I couldnt move the @VerifyBookingAccess because the service was called from more than 1 place .. assuming thats not a problem now!

steverendell (approved):
lgtm

PR #2083, opened 2024-04-03
SDIT-1634: 🎨 Remove unnecessary method

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/service/ImageService.java
```diff
@@ -35,7 +35,6 @@ public class ImageService {
     private final OffenderRepository offenderRepository;
 
     public List<ImageDetail> findOffenderImagesFor(final String offenderNumber) {
-        if (offenderRepository.findByNomsId(offenderNumber).isEmpty()) throw EntityNotFoundException.withId(offenderNumber);
```
petergphillips, line 38: the resource has a `VerifyOffenderAccess` on it anyway so this doesn't do anything

andymarke (approved):
LGTM

PR #2086, opened 2024-04-03
SDIT-896: 🐛 Switch to get the latest offender booking

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
```diff
@@ -679,7 +679,7 @@ public CaseNote updateOffenderCaseNote(
     @VerifyOffenderAccess(overrideRoles = {"GLOBAL_SEARCH", "VIEW_PRISONER_DATA"})
     @GetMapping("/{offenderNo}/sentences")
     public OffenderSentenceDetail getOffenderSentenceDetail(@PathVariable("offenderNo") @Parameter(description = "Noms ID or Prisoner number (also called offenderNo)", required = true) final String offenderNo) {
-        return bookingService.getOffenderSentenceDetail(offenderNo).orElseThrow(EntityNotFoundException.withId(offenderNo));
```
petergphillips, line 682: this does a `VerifyOffenderAccess` anyway so don't need to do an extra throw here

andymarke (approved):
lgtm

PR #2088, opened 2024-04-03
SDIT-1559 Add AccessDeniedException (403) to PUT /api/bookings/{bookingId}/move-to-cell-swap

steverendell:


petergphillips (approved):
LGTM

PR #2090, opened 2024-04-04
SDIT-1634: 🎨 More tidying of JPA

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/service/BookingService.java
```diff
@@ -1065,9 +1065,9 @@ private static String forAgency(final String agencyId) {
 
     public InmateDetail getOffender(final String offenderNo) {
         telemetryClient.trackEvent("getOffender 1.1_beta version used");
-        return offenderRepository.findRootOffenderByNomsId(offenderNo)
+        return offenderBookingRepository.findLatestOffenderBookingByNomsId(offenderNo)
```
petergphillips, line 1068: best way to get the booking is from the booking repository instead, otherwise get the root. Don't think this method is used (will find out in a week or so when the telemetry gets recorded), but have fixed it anyway.

src/main/java/uk/gov/justice/hmpps/prison/service/enteringandleaving/BookingIntoPrisonService.kt
```diff
@@ -76,16 +76,16 @@ class BookingIntoPrisonService(
...
-    val offender = offenderBookingRepository.findLatestOffenderBookingByNomsIdForUpdate(prisonerIdentifier)
-      .map { it.offender }
+    val (offender, offenderBooking) = offenderBookingRepository.findLatestOffenderBookingByNomsIdForUpdate(prisonerIdentifier)
+      .map { Pair(it.offender, it) }
       // if there are no bookings then get the root offender instead
-      .orElseGet { rootOffenderForUpdate(prisonerIdentifier).getOrThrow() }
+      .orElseGet { Pair(rootOffenderForUpdate(prisonerIdentifier).getOrThrow(), null) }
 
-    return newBookingWithoutUpdateLock(offender, requestForNewBooking)
+    return newBookingWithoutUpdateLock(offender, offenderBooking, requestForNewBooking)
```
petergphillips, line 84: passing in both then stops us from having to get the latest booking for the prisoner.  We can't just pass in the booking since the prisoner might not have any yet.

src/main/java/uk/gov/justice/hmpps/prison/service/transformers/OffenderTransformer.java
```diff
@@ -31,10 +29,8 @@ public class OffenderTransformer {
 
     private final Clock clock;
 
-    public InmateDetail transform(final Offender offender) {
-        return offender.getLatestBooking()
-            .map(this::transform)
-            .orElse(buildOffender(offender).build());
+    public InmateDetail transformWithoutBooking(final Offender offender) {
```
petergphillips, line 32: this stops us calling `getLatestBooking`, which can only be called on the root offender

src/main/java/uk/gov/justice/hmpps/prison/service/BookingService.java
```diff
@@ -1065,9 +1065,9 @@ private static String forAgency(final String agencyId) {
...
                 .map(offenderTransformer::transform)
-                .orElseThrow(EntityNotFoundException.withId(offenderNo));
+                .orElseGet(() -> offenderTransformer.transformWithoutBooking(offenderRepository.findRootOffenderByNomsId(offenderNo).orElseThrow()));
```
mikehalmamoj, line 1070: Is this a breaking change? Although I guess if you think it's not used then no matter.

mikehalmamoj (approved):

PR #2094, opened 2024-04-05
SDIT-1221: 🐛 Switch smoketest to risley

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/service/SmokeTestHelperService.kt
```diff
@@ -15,21 +14,19 @@ class SmokeTestHelperService(
   private val inmateService: InmateService,
   private val releasePrisonerService: ReleasePrisonerService,
   private val bookingIntoPrisonService: BookingIntoPrisonService,
-  private val offenderRepository: OffenderRepository,
+  private val offenderBookingRepository: OffenderBookingRepository,
 ) {
   companion object {
-    const val SMOKE_TEST_PRISON_ID = "LEI"
+    const val SMOKE_TEST_PRISON_ID = "RSI"
```
petergphillips, line 20: risley

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/SmokeTestHelperResourceIntTest.kt
```diff
@@ -119,7 +123,12 @@ class SmokeTestHelperResourceIntTest : ResourceTest() {
 
     @Test
     fun `will change prisoner name`() {
-      val prisonerNo = createPrisoner()
+      var prisonerNo: String? = null
```
petergphillips, line 126: bit naff here, but the `builder.build` doesn't return anything

andymarke: I think that makes sense, since what would it return? Especially if you created lots of prisoners?

andymarke (approved):
lgtm

karenmillermoj (approved):
lgtm

PR #2095, opened 2024-04-05
SDIT-1221: ✨ Use return from builder

petergphillips:


src/test/java/uk/gov/justice/hmpps/prison/dsl/NomisData.kt
```diff
@@ -66,6 +68,7 @@ class NomisData(
         )
           .also {
             builder.apply(dsl)
+            offenders += it
```
petergphillips, line 71: maintain list of offenders and teams that we've added

andymarke (commented):
Disagree this is a good idea. But lets discuss

PR #2098, opened 2024-04-09
SDIT-1656 Ensure agency location user description formatted correctly

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/service/support/LocationProcessor.java
```diff
@@ -104,34 +104,38 @@ public static Location processLocation(final Location location, final boolean pr
...
-                    stripAgencyId(locationSummary.getDescription(), locationSummary.getAgencyId())))
+            .userDescription(formatUserDescription(locationSummary))
             .description(locationSummary.getDescription())
             .agencyId(locationSummary.getAgencyId())
             .build();
     }
 
+    private static String formatUserDescription(LocationSummary locationSummary) {
+        return StringUtils.defaultIfBlank(formatLocation(locationSummary.getUserDescription()),
+            stripAgencyId(locationSummary.getDescription(), locationSummary.getAgencyId()));
```
petergphillips, line 136: is the user description always set?  Wondering if the `formatLocation` needs to go around the `getDescription` too

karenmillermoj: Well I wondered that but the Description is always stuff like CSI-A-1-001 which doesn't make sense to make lower case

karenmillermoj: fyi asked Eddie, he he happy with the normal description staying as it is (i.e. just the user description is formatted)

petergphillips (approved):
LGTM

PR #2100, opened 2024-04-10
SDIT-1621: 🎨 Convert tests to kotlin

petergphillips:


mikehalmamoj (approved):

src/test/java/uk/gov/justice/hmpps/prison/dsl/NomisData.kt
```diff
@@ -103,7 +103,7 @@ interface NomisDataDsl {
     firstName: String = randomName(),
     middleName1: String? = null,
     middleName2: String? = null,
-    birthDate: LocalDate = LocalDate.of(1965, 7, 19),
+    birthDate: LocalDate = LocalDate.of(1975, 7, 19),
```
petergphillips, line 106: ended up switching the default out of the 10 year period in the test so that we don't get so many offenders in the 1960-69 range.

mikehalmamoj (approved):

PR #2101, opened 2024-04-10
SDIT-1621: ✨ Add new find inmate aliases repo method

petergphillips:
- **SDIT-1621: ✨ Add new find inmate aliases repo method**


src/test/java/uk/gov/justice/hmpps/prison/repository/InmateRepositoryTest.kt
```diff
@@ -1519,80 +1526,76 @@ class InmateRepositoryTest {
     )
   }
 
-  @get:Test
```
petergphillips, line 1522: hadn't spotted this previously

PR #2102, opened 2024-04-10
SDIT-1617 New endpoint dedicated to prisoner search

mikehalmamoj:
This is the first PR in a multi-stage plan to switch to a new endpoint deicated to prisoner search.

- first we've created an endpoint that only includes details required by prisoner search
- for now this is calling the old InmateService to get data, and testing against the same old seed data
- we will then create a test fixture in prisoner search to compare the old and new endpoints with production data to give us confidence that the new endpoint works the same as the old one
- once established we can switch prisoner search to the new endpoint
- we will then be able to add additional data to the new endpoint without affecting other prison-api clients
- then we'll rewrite the tests for the new endpoint such that we don't rely on the seed data
- once we're happy with the test coverage we will write a new service to back the new endpoint
- at this point we're isolated from the old endpoint and service


petergphillips (approved):
LGTM

PR #2111, opened 2024-04-17
SDIT-1617 Start moving towards JPA instead of InmateService

mikehalmamoj:
This is for the new prisoner search endpoint that's not being used **yet** - it's being tested in parallel with the old endpoint to make sure it provides the same details. We don't want to get this wrong so we're testing with real data.

The idea is to move the endpoint away from InmateService (which relies heavily on SQL based repositories) and onto JPA instead. We haven't achieved this yet but this but are now using JPA modelling where it exists (still more to do).

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/service/InmateService.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/service/InmateService.java
@@ -309,7 +309,7 @@ public class InmateService {
         }
     }
 
-    private List<Assessment> getAllAssessmentsOrdered(final Long bookingId) {
+    public List<Assessment> getAllAssessmentsOrdered(final Long bookingId) {
         final var assessmentsDto = repository.findAssessments(Collections.singletonList(bookingId), null, Collections.emptySet());
 
         return assessmentsDto.stream().map(this::createAssessment).toList();
@@ -400,6 +400,10 @@ public class InmateService {
             .stream()
             .filter(i -> identifierType == null || identifierType.equalsIgnoreCase(i.getType()))
             .toList();
+    }
+
+    public List<Alias> getAliases(final Long bookingId) {
+        return repository.findInmateAliasesByBooking(bookingId, "createDate", Order.ASC, 0, 100).getItems().stream().toList();
     }
 
     @Transactional // route to primary in live so that we can get the latest data after a trigger
--- a/src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerSearchService.kt
+++ b/src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerSearchService.kt
@@ -1,13 +1,25 @@
 package uk.gov.justice.hmpps.prison.service
 
+import jakarta.transaction.Transactional
 import org.springframework.stereotype.Component
+import uk.gov.justice.hmpps.prison.api.model.InmateDetail
 import uk.gov.justice.hmpps.prison.api.model.PrisonerSearchDetails
+import uk.gov.justice.hmpps.prison.repository.jpa.repository.OffenderBookingRepository
+import uk.gov.justice.hmpps.prison.repository.jpa.repository.OffenderRepository
+import uk.gov.justice.hmpps.prison.service.transformers.OffenderTransformer
+import java.util.Optional
 
 @Component
-class PrisonerSearchService(private val inmateService: InmateService) {
+class PrisonerSearchService(
+  private val offenderBookingRepository: OffenderBookingRepository,
+  private val offenderRepository: OffenderRepository,
+  private val offenderTransformer: OffenderTransformer,
+  private val inmateService: InmateService,
+) {
 
+  @Transactional
   fun getPrisonerDetails(offenderNo: String): PrisonerSearchDetails =
-    inmateService.findOffender(offenderNo, true, false)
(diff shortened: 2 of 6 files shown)


src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerSearchService.kt
```diff
@@ -1,13 +1,25 @@
...
+class PrisonerSearchService(
+  private val offenderBookingRepository: OffenderBookingRepository,
+  private val offenderRepository: OffenderRepository,
+  private val offenderTransformer: OffenderTransformer,
+  private val inmateService: InmateService,
+) {
 
+  @Transactional
   fun getPrisonerDetails(offenderNo: String): PrisonerSearchDetails =
-    inmateService.findOffender(offenderNo, true, false)
```
mikehalmamoj, line 10: The main change is to stop using this service and load the offender and booking from JPA based repositories.

src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerSearchService.kt
```diff
@@ -45,4 +57,26 @@ class PrisonerSearchService(private val inmateService: InmateService) {
...
+        // TODO These need to be modelled in JPA and set by the OffenderTransformer
+        physicalAttributes = inmateService.getPhysicalAttributes(bookingId)
+        physicalCharacteristics = inmateService.getPhysicalCharacteristics(bookingId)
+        physicalMarks = inmateService.getPhysicalMarks(bookingId)
+        aliases = inmateService.getAliases(bookingId)
+        inmateService.getAllAssessmentsOrdered(bookingId).also {
+          csra = it.firstOrNull { it.isCellSharingAlertFlag }?.classification
+          categoryCode = it.firstOrNull { "CATEGORY" == it.assessmentCode }?.classificationCode
+        }
+      }
```
mikehalmamoj, line 77: We're still relying on InmateService for the details that aren't modelled by JPA.

src/main/java/uk/gov/justice/hmpps/prison/service/transformers/OffenderTransformer.java
```diff
@@ -63,9 +68,18 @@ public InmateDetail transform(final OffenderBooking latestBooking) {
...
                     .resultValue(pd.getCode().getDescription())
                     .build()).toList())
+            .legalStatus(latestBooking.getLegalStatus())
+            .imprisonmentStatus(latestBooking.getActiveImprisonmentStatus().map(ims -> ims.getImprisonmentStatus().getStatus()).orElse(null))
+            .imprisonmentStatusDescription(latestBooking.getActiveImprisonmentStatus().map(ims -> ims.getImprisonmentStatus().getDescription()).orElse(null))
+            .offenceHistory(offenceHistory)
+            .recall(RecallCalc.calculate(latestBooking.getBookingId(), latestBooking.getLegalStatus(), offenceHistory, sentenceTerms))
+            .receptionDate(latestBooking.getBookingBeginDate().toLocalDate())
+            .locationDescription(LocationProcessor.formatLocation(latestBooking.getLocation().getDescription()))
+            .latestLocationId(latestBooking.getLocation().getId())
```
mikehalmamoj, line 78: Add offender data supported by JPA to the offender transformer. This will be expanded once other data has JPA support.

src/main/java/uk/gov/justice/hmpps/prison/service/transformers/OffenderTransformer.java
```diff
@@ -63,9 +68,18 @@ public InmateDetail transform(final OffenderBooking latestBooking) {
...
             .build()
             .deriveStatus()
-            .splitStatusReason();
+            .splitStatusReason()
+            .updateReligion();
```
mikehalmamoj, line 82: Followed the existing pattern for derived data to get the religion - but this will also need moving out of InmateService soon.

src/main/java/uk/gov/justice/hmpps/prison/service/InmateService.java
```diff
@@ -402,6 +402,10 @@ public List<OffenderIdentifier> getOffenderIdentifiers(final Long bookingId, @Nu
             .toList();
     }
 
+    public List<Alias> getAliases(final Long bookingId) {
+        return repository.findInmateAliasesByBooking(bookingId, "createDate", Order.ASC, 0, 100).getItems().stream().toList();
+    }
```
mikehalmamoj, line 407: Put this in front of the repository so we don't get tied to the repo in PrisonerSearchService.

src/test/java/uk/gov/justice/hmpps/prison/service/enteringandleaving/TransferIntoPrisonServiceTest.kt
```diff
@@ -160,6 +162,7 @@ internal class TransferIntoPrisonServiceTest {
           birthDate = LocalDate.now().minusYears(30)
           gender = Gender("M", "MALE")
         }
+        bookingBeginDate = LocalDateTime.now().minusDays(1)
```
mikehalmamoj, line 165: We can't have null booking begin date anymore, which aligns with NOMIS where that column is never null.

andymarke (approved):
lgtm

PR #2112, opened 2024-04-18
SDIT-1669 Ensure jcache used for caching

karenmillermoj:


src/main/resources/application.yml
```diff
@@ -11,6 +11,9 @@ spring:
   application:
     name: prison-api
 
+  cache:
```
karenmillermoj, line 14: This config isn't actually needed.  Introducing the spring-boot-starter-cache and cache-api (both needed) automatically pulls in jcache.  However, I can imagine it would be easy to just remove the two jars and everything would "appear" to work ok, so I have added it in (leaving it in and removing the two jars will stop the application from starting, so is more of a safety net)

build.gradle.kts
```diff
@@ -32,13 +32,15 @@ dependencies {
   implementation("org.springframework.boot:spring-boot-starter-aop")
   implementation("org.springframework.boot:spring-boot-starter-validation")
   implementation("org.springframework.boot:spring-boot-starter-jdbc")
+  implementation("org.springframework.boot:spring-boot-starter-cache")
```
petergphillips, line 35: I think this is pulled in by kotlin lib anyway?

build.gradle.kts
```diff
@@ -32,13 +32,15 @@ dependencies {
...
   implementation("org.springframework.boot:spring-boot-starter-data-jpa")
   developmentOnly("org.springframework.boot:spring-boot-devtools")
 
   implementation("commons-codec:commons-codec:1.16.1")
   implementation("com.github.jsqlparser:jsqlparser:$jsqlParserVersion")
   implementation("org.ehcache:ehcache:3.10.8")
   implementation("com.zaxxer:HikariCP:5.1.0")
+  implementation("javax.cache:cache-api:1.1.1")
```
petergphillips, line 43: interesting, I thought that pulling in ehcache would do the trick

build.gradle.kts
petergphillips: no it's not 🙁

src/main/resources/application.yml
karenmillermoj: CORRECTION - the cache-api is not required

petergphillips (approved):
LGTM

PR #2120, opened 2024-04-23
SDIT-1648 Add title to prisoner search response

mikehalmamoj:


src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerSearchService.kt
```diff
@@ -41,7 +47,7 @@ class PrisonerSearchService(
           categoryCode = it.categoryCode,
           inOutStatus = it.inOutStatus,
           identifiers = it.identifiers?.sortedBy { it.whenCreated },
-          sentenceDetail = it.sentenceDetail,
+          sentenceDetail = it.sentenceDetail.apply { additionalDaysAwarded = booking?.additionalDaysAwarded ?: 0 },
```
mikehalmamoj, line 50: The old endpoint also returned additional days awarded so we should too.

PR #2125, opened 2024-04-25
SDIT-1617 Fix a bug which ignores offender charges 

mikehalmamoj:
It used to try and ignore any with `CREATE_USER_ID='SYS'` and `AUDIT_MODULE_NAME='MERGE'` but it also ignores any with `CREATE_USER_ID='SYS'` and `AUDIT_MODULE_NAME=null`.


petergphillips (approved):
LGTM

PR #2129, opened 2024-04-29
SDIT-1649 Return offender addresses to Prisoner Search

mikehalmamoj:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/Offender.java
```diff
@@ -166,6 +166,11 @@ public class Offender extends AuditableEntity {
     @Column(name = "LAST_NAME_ALPHA_KEY")
     private String lastNameAlphaKey;
 
+    @OneToMany(mappedBy = "offender", cascade = CascadeType.ALL, fetch = FetchType.LAZY)
+    @Default
+    @Exclude
+    private List<OffenderAddress> addresses = new ArrayList<>();
```
petergphillips, line 172: I take it Prison API will return an empty list rather than `null` if there are no addresses?

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/Offender.java
```diff
@@ -166,6 +166,11 @@ public class Offender extends AuditableEntity {
     @Column(name = "LAST_NAME_ALPHA_KEY")
     private String lastNameAlphaKey;
 
+    @OneToMany(mappedBy = "offender", cascade = CascadeType.ALL, fetch = FetchType.LAZY)
```
petergphillips, line 169: not really understanding this mapping.  Does that mean that each offender record can have addresses - is there no mapping by the root offender?

petergphillips (approved):
LGTM

PR #2132, opened 2024-04-30
Creates an endpoint to get latest TUSED dates for an offender.

dave-virden:
* Gets the latest TUSED dates from historical bookings for analysis in Calculate Release Dates
* Updates tests accordingly

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/api/model/LatestTusedData.kt
+++ b/src/main/java/uk/gov/justice/hmpps/prison/api/model/LatestTusedData.kt
@@ -1,0 +1,13 @@
+package uk.gov.justice.hmpps.prison.api.model
+
+import com.fasterxml.jackson.annotation.JsonInclude
+import io.swagger.v3.oas.annotations.media.Schema
+import java.time.LocalDate
+
+@Schema(description = "Latest TUSED dates and associated info for offender")
+data class LatestTusedData(
+  val latestTused: LocalDate?,
+  val latestOverrideTused: LocalDate?,
+  val comment: String?,
+  val offenderNo: String?,
+  )
--- a/src/main/java/uk/gov/justice/hmpps/prison/api/model/LatestTusedData.kt
+++ b/src/main/java/uk/gov/justice/hmpps/prison/api/model/LatestTusedData.kt
@@ -1,6 +1,5 @@
 package uk.gov.justice.hmpps.prison.api.model
 
-import com.fasterxml.jackson.annotation.JsonInclude
 import io.swagger.v3.oas.annotations.media.Schema
 import java.time.LocalDate
 
@@ -10,4 +9,4 @@ data class LatestTusedData(
   val latestOverrideTused: LocalDate?,
   val comment: String?,
   val offenderNo: String?,
-  )
+)
--- a/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderBookingRepository.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderBookingRepository.java
@@ -13,6 +13,7 @@ import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.CrudRepository;
 import org.springframework.data.repository.PagingAndSortingRepository;
 import org.springframework.stereotype.Repository;
+import uk.gov.justice.hmpps.prison.api.model.LatestTusedData;
 import uk.gov.justice.hmpps.prison.repository.jpa.model.AgencyLocation;
 import uk.gov.justice.hmpps.prison.repository.jpa.model.OffenderBooking;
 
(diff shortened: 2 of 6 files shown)


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderDatesResourceTest.java
```diff
@@ -229,4 +230,75 @@ public void testGetOffenderKeyDatesForOffenderSentCalcId() {
         assertEquals(LocalDate.of(2020, 12, 30),nomisCalculations.getHomeDetentionCurfewEligibilityDate());
     }
 
+    @Test
+    public void testGetLatestTusedForPrisoner() {
+        // Given
+        final var request = createEmptyHttpEntity(AuthToken.CRD_USER);
+        final var type = new ParameterizedTypeReference<LatestTusedData>() {};
+        // When
+        final var response = testRestTemplate.exchange(
```
petergphillips, line 239: don't change, but for new code we try and use the webclient instead

src/test/java/uk/gov/justice/hmpps/prison/service/OffenderDatesServiceTest.java
```diff
@@ -381,6 +384,21 @@ private static void assertOffenderCalculatedKeyDates(OffenderCalculatedKeyDates
             .build());
     }
 
+    @Test
+    void findLatestTusedDataIsReturnedCorrectly() {
+        LatestTusedData latestTusedData = new LatestTusedData(LocalDate.of(2023, 1, 3), null, null, "A1234AA");
+        when(offenderBookingRepository.findLatestTusedDataFromNomsId(anyString())).thenReturn(Optional.of(latestTusedData));
+        LatestTusedData returned = service.getLatestTusedDataFromNomsId("A1234AA");
+        assertEquals("A1234AA", returned.getOffenderNo());
```
petergphillips, line 392: should be using `assertThat` rather than `assertEquals`, I tried to remove all of them a while ago but looks like some have snuck back in

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderBookingRepository.java
```diff
@@ -71,4 +72,12 @@ List<OffenderBookingId> findDistinctByActiveTrueAndOffenderNomsIdInAndSentences_
     @Override
     @EntityGraph(type = EntityGraphType.FETCH, value = "booking-with-summary")
     Page<OffenderBooking> findAll(@NotNull Specification<OffenderBooking> filter, @NotNull Pageable pageable);
+
+    @Query("""
+    SELECT new uk.gov.justice.hmpps.prison.api.model.LatestTusedData(sc.tusedCalculatedDate, sc.tusedOverridedDate, sc.comments, o.nomsId)
+    from SentenceCalculation sc
+    inner join sc.offenderBooking.offender o where o.nomsId = :nomsId and coalesce(sc.tusedOverridedDate, sc.tusedCalculatedDate) is not null
```
petergphillips, line 79: seems a slightly odd way of doing it - if the latest booking doesn't have a TUSED date then it will use a previous booking values? Is that correct?

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderBookingRepository.java
```diff
@@ -71,4 +72,12 @@ List<OffenderBookingId> findDistinctByActiveTrueAndOffenderNomsIdInAndSentences_
...
+    order by sc.calculationDate desc limit 1
```
petergphillips, line 80: didn't realise that `limit` was a valid JPQL keyword - what does the SQL translate to?

src/main/java/uk/gov/justice/hmpps/prison/api/model/LatestTusedData.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import io.swagger.v3.oas.annotations.media.Schema
+import java.time.LocalDate
+
+@Schema(description = "Latest TUSED dates and associated info for offender")
+data class LatestTusedData(
+  val latestTused: LocalDate?,
+  val latestOverrideTused: LocalDate?,
+  val comment: String?,
+  val offenderNo: String?,
```
petergphillips, line 11: don't think this needs to be nullable?

src/test/java/uk/gov/justice/hmpps/prison/service/OffenderDatesServiceTest.java
dave-virden: I'll sort them when I address mine

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderBookingRepository.java
dave-virden: Yes, it's peculiar to this sort of sentence that we're looking at, need to consider dates from previous bookings that might still be valid

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderBookingRepository.java
dave-virden: I can't find the reference but I think it's a fairly recent addition

src/main/java/uk/gov/justice/hmpps/prison/api/model/LatestTusedData.kt
dave-virden: I think it needs to be nullable so it can be deserialized correctly

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderBookingRepository.java
petergphillips: probably worth adding something to the endpoint swagger docs describing ☝️ since it could be returning data people aren't expecting

petergphillips (approved):
LGTM

petergphillips (approved):
LGTM

PR #2134, opened 2024-04-30
SDIT-1617 Fix data for prisoner search

mikehalmamoj:


src/main/java/uk/gov/justice/hmpps/prison/service/InmateService.java
```diff
@@ -343,6 +343,7 @@ private void setAlertsFields(final InmateDetail inmate) {
             activeAlertCount.incrementAndGet();
             alertTypes.add(a.getAlertType());
         });
+        items.sort(Comparator.comparing(Alert::getAlertId));
```
mikehalmamoj, line 346: When I tested switching prisoner search to the new endpoint the index refresh job had differences on loads of records because the alerts were in a different sequence to what they used to be.

Therefore I'm planning to do a full rebuild of the indexes using the old endpoint with the alerts in the new order - so we won't publish millions of events when we switch endpoints.

src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerSearchService.kt
```diff
@@ -51,7 +51,7 @@ class PrisonerSearchService(
           inOutStatus = it.inOutStatus,
           identifiers = it.identifiers?.sortedBy { it.whenCreated },
           allIdentifiers = offender.allIdentifiers?.map { oi -> oi.toModel(it.offenderNo) }?.sortedBy { it.whenCreated },
-          sentenceDetail = it.sentenceDetail?.apply { additionalDaysAwarded = booking?.additionalDaysAwarded ?: 0 },
+          sentenceDetail = it.sentenceDetail?.apply { additionalDaysAwarded = booking?.additionalDaysAwarded },
```
mikehalmamoj, line 54: We're going to end up with some diffs here becaues the old endpoint has a mix of zeros and null for ADA. However most seem to be null so make the new endpoint return null to minimise the number of diffs.

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/PrisonerSearchResourceIntTest.kt
```diff
@@ -126,8 +131,8 @@ class PrisonerSearchResourceIntTest : ResourceTest() {
               assertThat(nonDtoReleaseDateType).isEqualTo(NonDtoReleaseDateType.ARD)
               assertThat(confirmedReleaseDate).isEqualTo("2018-04-19") // TODO test where OFFENDER_RELEASE_DETAILS.RELEASE_DATE is null, this should be null and NOT default to AUTO_RELEASE_DATE
               assertThat(releaseDate).isEqualTo("2018-04-19")
-              assertThat(additionalDaysAwarded).isEqualTo(0)
             }
+            assertThat(sentenceDetail?.additionalDaysAwarded).isNull()
```
mikehalmamoj, line 135: Bizarrely `assertThat(sentenceDetail!!.additionalDaysAwarded).isNull()` throws a null pointer exception so I had to move this out of the `with` above.

PR #2135, opened 2024-05-01
SDIT-924 Lock check for PUT /api/bookings/{bookingId}/living-unit/{internalLocationDescription}

steverendell:
Add lockTimeout feature

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderBookingRepository.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderBookingRepository.java
@@ -1,6 +1,7 @@
 package uk.gov.justice.hmpps.prison.repository.jpa.repository;
 
 import jakarta.persistence.LockModeType;
+import jakarta.persistence.QueryHint;
 import jakarta.validation.constraints.NotNull;
 import org.springframework.data.domain.Page;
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderBookingRepository.java
```diff
@@ -68,6 +70,13 @@ List<OffenderBookingId> findDistinctByActiveTrueAndOffenderNomsIdInAndSentences_
 
     Optional<OffenderBooking> findByBookingId(Long bookingId);
 
+    String lockWaitTimeMillis = "10000";
```
karenmillermoj, line 73: Should we pick this up from application yaml (and use this value as the default?)

karenmillermoj (approved):
lgtm

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderBookingRepository.java
steverendell: I was thinking of setting 'jakarta.persistence.lock.timeout' in application.yaml , in which case it applies globally but a custom config value might be better (at least in the short to medium term) as it would only apply for the endpoints we want. Will add this in with the next lock subtask.

PR #2141, opened 2024-05-01
SDIT-924 Lock check for PUT /api/bookings/{bookingId}/living-unit/{internalLocationDescription}

steverendell:
Correct the exception which needs to be caught
Also decrease all lock timeouts to 10s (ensuring they're less than the new UI 20s timeout)

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/repository/BookingRepository.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/repository/BookingRepository.java
@@ -318,7 +318,7 @@ public class BookingRepository extends RepositoryBase {
         return activities.stream().map(ScheduledEventDto::toScheduledEvent).collect(Collectors.toList());
     }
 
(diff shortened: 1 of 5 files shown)


petergphillips (approved):
LGTM

src/main/java/uk/gov/justice/hmpps/prison/repository/BookingRepository.java
```diff
@@ -318,7 +318,7 @@ public List<ScheduledEvent> getBookingActivities(final Collection<Long> bookingI
         return activities.stream().map(ScheduledEventDto::toScheduledEvent).collect(Collectors.toList());
     }
 
-    private final static int lockWaitTime = 25;
+    private final static int lockWaitTime = 10;
```
steverendell, line 321: Ill pull these constants into a single config item when I do the next lock PR

PR #2144, opened 2024-05-07
SDIT-1617 Do not return all offences to prisoner search

mikehalmamoj:


src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerSearchService.kt
```diff
@@ -69,7 +68,6 @@ class PrisonerSearchService(
           addresses = offender.addresses.filter { it.noFixedAddressFlag == "N" }.map(AddressTransformer::translate),
           phones = offender.phones.map(AddressTransformer::translate),
           emailAddresses = offender.emailAddresses.map(AddressTransformer::translate),
-          allOffences = offender.allBookings.map { it.bookingId }.toSet().let { bookingService.getActiveOffencesForBookings(it) }.filter { it.convicted() },
```
mikehalmamoj, line 72: This was wrong - we shouldn't be checking the active flag, adn the attribute should be called `convictedOffences`. Already removed from prisoner search so removing from here and starting again.

petergphillips (approved):
LGTM

PR #2150, opened 2024-05-09
SDIT-1710 Check Agency exists before determining access

karenmillermoj:
This mimics the check for bookings


src/main/java/uk/gov/justice/hmpps/prison/service/AgencyService.java
```diff
@@ -214,7 +215,7 @@ public Set<String> getAgencyIds(boolean allowInactive) {
      * @throws AccessDeniedException   if current user does not have access to this agency and accessDeniedError is true.
      */
     public void verifyAgencyAccess(final String agencyId, boolean accessDeniedError, boolean allowInactive) {
-        Objects.requireNonNull(agencyId, "agencyId is a required parameter");
+        checkAgencyExists(agencyId, allowInactive ? ALL: ACTIVE_ONLY);
```
steverendell, line 218: Not sure about this - does it matter if a 403 is returned when the agency doesnt exist? Just thinking about the extra db calls 🤷

steverendell (approved):
lgtm

src/main/java/uk/gov/justice/hmpps/prison/service/BookingService.java
```diff
@@ -543,15 +543,20 @@ private void validateScheduledEventsRequest(final LocalDate fromDate, final Loca
...
+            return;
+        }
 
         final var agencyIds = agencyService.getAgencyIds(false);
         if (AuthenticationFacade.hasRoles("INACTIVE_BOOKINGS")) {
             agencyIds.addAll(Set.of("OUT", "TRN"));
         }
 
         if (agencyIds.isEmpty()) {
+            checkBookingExists(bookingId);
```
karenmillermoj, line 559: All of the "if" blocks are starting to look a bit unreadable.  This should hopefully start to tidy up once we can remove the checks for different types of exceptions to throw and also we will not need to do explicit client only checks

steverendell (approved):
LGTM!

src/main/java/uk/gov/justice/hmpps/prison/service/AgencyService.java
karenmillermoj: As discussed - move checks into non-happy path

PR #2153, opened 2024-05-15
SDIT-1745 Join on staff members to fix n+1 query

mikehalmamoj:


andymarke (approved):
lgtm

mikehalmamoj:
It looks like there was an attempt to fix this with an subgraph but it doesn't work - better to get the subgraph working than to force an eager load for everyone

PR #2154, opened 2024-05-15
MAP-1130 Improve roll count ordering using Natural Sort Order

Mjwillis:


src/main/java/uk/gov/justice/hmpps/prison/api/model/RollCount.java
```diff
@@ -84,5 +100,9 @@ public class RollCount {
     @NotNull
     private Integer outOfOrder;
 
-
+    @org.jetbrains.annotations.NotNull
```
andymarke, line 103: BTW I thought Kotlin understand all NotNull annotations from different libraries but maybe I'm wrong

src/main/java/uk/gov/justice/hmpps/prison/util/NaturalOrderComparator.kt
```diff
@@ -0,0 +1,82 @@
+package uk.gov.justice.hmpps.prison.util
+
+class NaturalOrderComparator : Comparator<SortAttribute> {
```
andymarke, line 3: What is this doing? What sort order is it trying to achieve?

andymarke (approved):
lgtm

PR #2158, opened 2024-05-16
SDIT-1755 Fix issue where expired new alerts are created active

andymarke:
When an alert is created, and it has already expired the status should be set to INACTIVE


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/BookingResourceIntTest.kt
```diff
@@ -760,6 +769,84 @@ class BookingResourceIntTest : ResourceTest() {
       assertThat(response.body!!.alertId).isGreaterThan(1)
       assertThat(response.statusCode).isEqualTo(HttpStatus.CREATED)
     }
+
```
andymarke, line 772: no point rewriting all these tests since this endpoint will be removed in the next month when the data moves to DPS

steverendell (approved):
lgtm

PR #2160, opened 2024-05-17
SDIT-1751: 🔧 Tidy mod security and enable in production

petergphillips:


helm_deploy/values-prod-cp.yaml
```diff
@@ -10,44 +10,6 @@ generic-service:
     tlsSecretName: prison-api-cert
     modsecurity_enabled: true
 
-    modsecurity_snippet: |
-      SecAuditEngine On
-      SecRuleEngine DetectionOnly
```
petergphillips, line 15: decided not to go with detection only - go with enabled as have checked old azure configuration and we've got all the rules in place as is.  Can always switch back if needed

mikehalmamoj (approved):

PR #2170, opened 2024-05-21
SDIT-1774 Set default exception for VerifyBookingAccess to AccessDenied

karenmillermoj:


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/BookingResourceIntTest.kt
```diff
@@ -2147,34 +2141,46 @@ class BookingResourceIntTest : ResourceTest() {
   inner class GetBookingSentenceDetail {
     @Test
     fun `should return 401 when user does not even have token`() {
-      webTestClient.get().uri("api/bookings/-1/sentenceDetail")
+      webTestClient.get().uri("/api/bookings/-1/sentenceDetail")
```
steverendell, line 2144: Good spot!

PR #2173, opened 2024-05-22
SDIT-1751: 🔧 Tweak connection pool settings to reduce invalid connections

petergphillips:


src/main/resources/application-nomis.yml
```diff
@@ -19,7 +19,7 @@ spring:
     datasource:
       hikari:
         maximum-pool-size: 10
-        connectionTimeout: 5000
+        connectionTimeout: 2500
```
petergphillips, line 22: this was originally 2000, changing to 2500 to match the primary

helm_deploy/prison-api/values.yaml
```diff
@@ -85,7 +85,7 @@ generic-service:
     APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.json
 
     SPRING_DATASOURCE_HIKARI_MAXIMUM_POOL_SIZE: 15
-    SPRING_REPLICA_DATASOURCE_HIKARI_MAXIMUM_POOL_SIZE: 50
+    SPRING_REPLICA_DATASOURCE_HIKARI_MAXIMUM_POOL_SIZE: 25
```
petergphillips, line 88: hasn't gone about 8 in last 24 hours

steverendell: Could go lower even than this. My theory now is that the problematic pool entry was left idle for a very long time and this was why the setNetworkTimeout call failed. So the smaller the pool, the lower the chance of any pool entry being idle for an extended period. And that would mean  setting maxLifeTime would also work. Interestingly the hikari read.me https://github.com/brettwooldridge/HikariCP?tab=readme-ov-file#frequently-used says 'We strongly recommend setting this value'

steverendell (approved):
lgtm

PR #2195, opened 2024-06-05
SDIT-1798 Offender assessments causing database lock and hikari timeouts

steverendell:
Fix after testing in dev


petergphillips (approved):
LGTM

PR #2205, opened 2024-06-13
SDIT-1808: 🎨 Rewrite auth facade into kotlin

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/security/AuthenticationFacade.java
```diff
@@ -1,95 +0,0 @@
...
-    public Authentication getAuthentication() {
-        return SecurityContextHolder.getContext().getAuthentication();
-    }
-
-    public String getCurrentPrincipal() {
-        final String username;
-
-        final var userPrincipal = getUserPrincipal();
-
-        if (userPrincipal instanceof String) {
```
petergphillips, line 33: not previous implementation could't tell what type of principal was being returned - test code has all refactored to now use the WithMockAuthUser so we know it will be a string

mikehalmamoj (approved):

PR #2206, opened 2024-06-14
SDIT-1808: 🎨 Rewrite current username method to reflect purpose

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/security/AuthenticationFacade.java
```diff
@@ -25,7 +25,7 @@ public Authentication getAuthentication() {
         return SecurityContextHolder.getContext().getAuthentication();
     }
 
-    public String getCurrentUsername() {
+    public String getCurrentPrincipal() {
```
petergphillips, line 28: this confused me for quite a while - line 31 is actually getting the principal in the authentication token.  This will be the `username` if set, if not it will be the `clientId` so will never be null.

Most of the usage in Prison API though actually wants the `username` instead since it is then retrieving the active caseload etc. for the user.

mikehalmamoj (approved):

PR #2210, opened 2024-06-17
SDIT-1715 Remove unused 404 exception code when have no access - now superceded by 403

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/service/AgencyService.java
```diff
@@ -223,22 +223,14 @@ public void verifyAgencyAccess(final String agencyId, boolean accessDeniedError,
         if (agencyIds.isEmpty()) {
```
karenmillermoj, line 223: I will do a separate PR to tidy up this method - just realised I missed it (This will be similar to the change for bookingservice)

PR #2212, opened 2024-06-18
MAP-1277 Sort correctly using natural sorting

Mjwillis:
Old Roll count

New Roll count

Sorting incorrect on new


steverendell (approved):
lgtm

PR #2220, opened 2024-06-26
SDIT-1593: 🎨 Start to remove jwt parameters usage

petergphillips:


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/ResourceTest.kt
```diff
@@ -45,11 +41,8 @@ abstract class ResourceTest {
   @Autowired
   protected lateinit var webTestClient: WebTestClient
 
-  @SpyBean
-  protected lateinit var telemetryClient: TelemetryClient
-
   @Autowired
-  protected lateinit var jwtAuthenticationHelper: JwtAuthenticationHelper
+  protected lateinit var jwtAuthenticationHelper: JwtAuthorisationHelper
```
petergphillips, line 45: I tried to keep `JwtAuthenticationHelper` but `JwtAuthorisationHelper` also declares a primary JWT decoder so having both in the same project causes 401 errors.  Had to therefore remove it and do one big PR.

steverendell (approved):
lgtm

PR #2222, opened 2024-06-27
BAVL-213 adds the event id to the response object on the schedules appointment search endpoint so appointments can be explicitly identified.

matt-crowson-moj:
The BLVS service needs to be able to make amendments to appointments in NOMIS.

To do this we need the appointment event ID (which we don't store in our service).  To achieve this we have to search for appointments based on criteria we do hold e.g. prison code, prisoner number, date, etc.


steverendell (approved):
lgtm

PR #2223, opened 2024-07-02
SDIT-1842 Get last modified height and weight for prisoner search

mikehalmamoj:
We previously always took the `OFFENDER_PHYSICAL_ATTRIBUTES` record with `ATTRIBUTE_SEQ=1`. The prisoner profile team asked if we could send them the last modified record that for the sync. to DPS - this is the latest value entered by a user. However this would cause discrepancies with Prisoner Search which still displays the first sequence.

Therefore this PR changes the prisoner search endpoint to return the last modified physical attributes and so will align with DPS.


petergphillips (approved):
LGTM

PR #2226, opened 2024-07-05
SDIT-1882 Include duplicate merge offences for prisoner search

mikehalmamoj:
We previously copied some functionality from the SQL driven endpoints to ignore offences copied during a merge (with create user SYS and audit module MERGE). However we realised this removes the active charge on the new booking rather than the inactive charge on the old booking. This is bad - for example it could corrupt the recall flag if we ignore an active recall charge. So we've decided to remove that functionality entirely and accept the duplicate charge in the "all convicted offences" data. Most users should only be interested in active offences which should be retrieved from the DPS offences API.


andymarke (approved):
LGTM

PR #2227, opened 2024-07-09
SDIT-1593: ✨ Start to switch to authentication holder from kotlin lib

petergphillips:


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/BookingResourceIntTest_getCalculableSentenceEnvelope.kt
```diff
@@ -15,9 +13,6 @@ import java.time.LocalDate
 @DisplayName("GET /api/bookings/latest/calculable-sentence-envelope")
 class BookingResourceIntTest_getCalculableSentenceEnvelope : ResourceTest() {
 
-  @Autowired
-  lateinit var objectMapper: ObjectMapper
```
petergphillips, line 19: moved up a level

src/main/java/uk/gov/justice/hmpps/prison/aop/connectionproxy/OracleConnectionAspect.kt
```diff
@@ -56,7 +56,7 @@ class OracleConnectionAspect(
 
   @Throws(SQLException::class)
   private fun Connection.openProxySessionForCurrentUsername(): OracleConnection {
-    val currentUsername = authenticationFacade.currentPrincipal
+    val currentUsername = authenticationFacade.principal
```
petergphillips, line 59: by the time we get here we've already discovered it is a nomis so no danger of the authentication not being set, in fact better to thrown an exception if it isn't set.

mikehalmamoj (dismissed):

mikehalmamoj (approved):

PR #2236, opened 2024-07-19
MAP-1416 Add filtering of schedules by location paths

Mjwillis:


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/ScheduleResourceTest.kt
```diff
@@ -126,135 +125,6 @@ class ScheduleResourceTest : ResourceTest() {
     }
   }
 
-  @Nested
```
Mjwillis, line 129: Move to its own test class

steverendell (approved):
lgtm

PR #2240, opened 2024-08-02
SDIT-1927: 🔨 Add in allow listing for dev too

petergphillips:


helm_deploy/prison-api/values.yaml
```diff
@@ -10,11 +10,10 @@ generic-service:
     tag: latest
     port: 8080
 
-  service:
```
petergphillips, line 13: bit of tidying up now we don't deploy to azure as well.  Configuration moved here instead of each of the `values-*.yaml` files

andymarke (approved):

PR #2242, opened 2024-08-05
SDIT-1925 Refactor GET /api/agencies/type/{type} tests

andymarke:
Refactor before making change to add court type description


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/AgencyResourceIntTest.kt
```diff
@@ -189,4 +189,114 @@ class AgencyResourceIntTest : ResourceTest() {
...
   }
+
+  @Nested
+  @DisplayName("GET /api/agencies/type/{type}")
+  inner class GetAgencyByType {
+    @Nested
+    inner class Security {
+      @Test
+      fun `should return 401 when user does not even have token`() {
+        webTestClient.get().uri("/api/agencies/type/{type}", "INST")
```
mikehalmamoj, line 200: I never use parameter substituion with WebTestClient - just hardcode the String e.g. `"/api/agencies/type/INST". Is there any advantage to use substituion?

mikehalmamoj (approved):

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/AgencyResourceIntTest.kt
andymarke: No. I will change in the next PR - thanks @mikehalmamoj

PR #2246, opened 2024-08-07
Adjst 786 return latest booking calculations

ldlharper:


src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderDatesResource.java
```diff
@@ -102,9 +103,10 @@ public ResponseEntity<OffenderCalculatedKeyDates> getOffenderKeyDatesForOffender
     @GetMapping("/calculations/{nomsId}")
     @PreAuthorize("hasRole('RELEASE_DATES_CALCULATOR')")
     @ProxyUser
-    public ResponseEntity<List<SentenceCalculationSummary>> getOffenderCalculations(@PathVariable("nomsId") @Parameter(description = "The booking id of offender", required = true) final String nomsId) {
+    public ResponseEntity<List<SentenceCalculationSummary>> getOffenderCalculations(@PathVariable("nomsId") @Parameter(description = "The booking id of offender", required = true) final String nomsId,
+            @Parameter(name = "activeOnly", description = "return only calculations from active bookings") @RequestParam(value = "activeOnly", required = false, defaultValue = "true") final Boolean activeOnly) {
```
andymarke, line 107: I think `activeOnly` should be renamed `latestOnly`

booking sequence of 1 is just the latest booking - which itself might be inactive

andymarke (approved):
LGTM but might be best to rename "active"

src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderDatesResource.java
ldlharper: Ah thats a good point. I hadn't thought about if the booking was active or not. The query already filters for active only bookings, so I'd named it completely wrong.

However what I want to do is return all calculations from all bookings. So would `activeOnly` now be okay? Or `latestAndActiveOnly`

andymarke (dismissed):

andymarke (approved):

PR #2248, opened 2024-08-09
SDIT-1593: ✨ More switching to authentication holder

petergphillips:
- **SDIT-1593: ✨ More switching to authentication holder**
- **SDIT-1593: ✨ More switching to authentication holder**

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/api/resource/BookingResource.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/api/resource/BookingResource.java
@@ -35,6 +35,7 @@ import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.ResponseStatus;
 import org.springframework.web.bind.annotation.RestController;
+import uk.gov.justice.hmpps.kotlin.auth.HmppsAuthenticationHolder;
 import uk.gov.justice.hmpps.prison.api.model.Account;
 import uk.gov.justice.hmpps.prison.api.model.Alert;
 import uk.gov.justice.hmpps.prison.api.model.AlertChanges;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/justice/hmpps/prison/api/resource/BookingResource.java
```diff
@@ -935,7 +935,7 @@ public List<PrisonDetails> getBookingVisitsPrisons(
     public ScheduledEvent postBookingsBookingIdAppointments(
         @PathVariable("bookingId") @Parameter(description = "The offender booking id", required = true) final Long bookingId,
         @RequestBody @Parameter(required = true) final NewAppointment newAppointment) {
-        return appointmentsService.createBookingAppointment(bookingId, authenticationFacade.getCurrentPrincipal(), newAppointment);
+        return appointmentsService.createBookingAppointment(bookingId, hmppsAuthenticationHolder.getUsername(), newAppointment);
```
petergphillips, line 938: calls through to get case loads for user so can be replaced

src/main/java/uk/gov/justice/hmpps/prison/service/AppointmentsService.kt
```diff
@@ -125,7 +124,7 @@ class AppointmentsService(
   @Transactional
   fun createBookingAppointment(
     bookingId: Long,
-    username: String,
+    username: String?,
```
petergphillips, line 127: previously `username` would actually hold the principal, hence couldn't be `null`.  Since now holds the `username` it needs to be nullable.

src/main/java/uk/gov/justice/hmpps/prison/service/validation/CaseNoteTypeSubTypeValidator.java
```diff
@@ -37,7 +37,7 @@ public boolean isValid(final NewCaseNote value, final ConstraintValidatorContext
         if (value == null) return true;
 
         // This should be ok as it is cached:
-        final var caseLoad = caseLoadService.getWorkingCaseLoadForUser(authenticationFacade.getCurrentPrincipal());
+        final var caseLoad = caseLoadService.getWorkingCaseLoadForUser(hmppsAuthenticationHolder.getUsername());
```
petergphillips, line 40: switch to `username` call as we check the caseload for the user

src/test/java/uk/gov/justice/hmpps/prison/service/CaseNoteMovementServiceImplIntTest.java
```diff
@@ -16,6 +17,7 @@
 
 @ActiveProfiles("test")
 @SpringBootTest
+@WithMockAuthUser("")
```
petergphillips, line 20: this test then started failing as no authentication in the session (invalid with new authentication holder).  Using `""` then ensures that the `username` is not set and `clientId` is set, which is a more appropriate test

andymarke (approved):

PR #2253, opened 2024-08-12
Revert "Revert "SDIT-1593: ✨ Start to switch to authentication holder from kotlin lib…" (#2229)"

petergphillips:
This reverts commit 783e1d53d80aa94429ef565b746e1be5d4421e02 and PR #2229


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/ResourceTest.kt
```diff
@@ -47,6 +48,9 @@ abstract class ResourceTest {
   @Autowired
   protected lateinit var authTokenHelper: AuthTokenHelper
 
+  @Autowired
+  protected lateinit var objectMapper: ObjectMapper
```
petergphillips, line 52: moved up a level

src/main/java/uk/gov/justice/hmpps/prison/aop/ProxyUserAspect.java
```diff
@@ -30,18 +30,18 @@ public void proxyUserPointcut() {
     @Around("proxyUserPointcut()")
     public Object controllerCall(final ProceedingJoinPoint joinPoint) throws Throwable {
 
-        var proxyUser = authenticationFacade.getCurrentPrincipal();
+        var authentication = authenticationFacade.getAuthenticationOrNull();
```
petergphillips, line 33: in the new holder `authentication` can be null if no authentication for call so check that instead

Mjwillis (approved):
LGTM

PR #2254, opened 2024-08-13
SDIT-1933 Fix OffenderBooking.offenderPhysicalAttributes mapping

andymarke:
OrderColumn would fill list will nulls if first sequence was greater than 1 causing a NPE

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderBooking.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderBooking.java
@@ -309,8 +309,6 @@ public class OffenderBooking extends AuditableEntity {
     private ReleaseDetail releaseDetail;
 
     @OneToMany(mappedBy = "id.offenderBooking", cascade = CascadeType.ALL)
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/OffenderBooking.java
```diff
@@ -799,13 +797,15 @@ private String getLastMovementTypeCode() {
...
     public OffenderPhysicalAttributes getLatestPhysicalAttributes() {
-        return offenderPhysicalAttributes.stream().max(Comparator.comparing((pa) -> {
-            if (pa.getModifyDatetime() != null) {
-                return pa.getModifyDatetime();
-            } else {
-                return pa.getCreateDatetime();
-            }
-        })).orElse(null);
+        return offenderPhysicalAttributes.stream()
+            .sorted(Comparator.comparing((pa) -> pa.getId().getSequence()))
```
andymarke, line 801: get in the right order before doing max in case more than one attribute share the same modifyDateTime

karenmillermoj (approved):
lgtm

PR #2257, opened 2024-08-14
SDIT-1593: ✨ More switching to authentication holder

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/api/resource/LocationResource.java
```diff
@@ -72,7 +68,7 @@ public ResponseEntity<List<OffenderBooking>> getOffendersAtLocationDescription(
             @RequestHeader(value = "Sort-Fields", defaultValue = "lastName,firstName,bookingId", required = false) @Parameter(description = "Comma separated list of one or more of the following fields - <b>bookingNo, bookingId, offenderNo, firstName, lastName, agencyId, or assignedLivingUnitId</b>") final String sortFields,
             @RequestHeader(value = "Sort-Order", defaultValue = "ASC", required = false) @Parameter(description = "Sort order (ASC or DESC) - defaults to ASC.") final Order sortOrder) {
         final var request = SearchOffenderRequest.builder()
-                .username(authenticationFacade.getCurrentPrincipal())
+                .username(hmppsAuthenticationHolder.getPrincipal())
```
petergphillips, line 71: couldn't work out whether we could get away with `username` here, so left it as `principal`

src/main/java/uk/gov/justice/hmpps/prison/api/resource/UserResource.java
```diff
@@ -71,7 +61,7 @@ public UserResource(final AuthenticationFacade authenticationFacade,
     @ProgrammaticAuthorisation("Returns information about the current user only")
     @GetMapping("/me")
     public UserDetail getMyUserInformation() {
-        return userService.getUserByUsername(authenticationFacade.getCurrentPrincipal());
+        return userService.getUserByUsername(hmppsAuthenticationHolder.getUsername());
```
petergphillips, line 64: this is all about the current user information, so fine to replace with `getUsername`

mikehalmamoj (approved):

PR #2280, opened 2024-09-30
SDIT-2056: 🥅 Rethrow proxy session user errors

petergphillips:
- **SDIT-2056: 🥅 Rethrow proxy session user errors**

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/aop/connectionproxy/OracleConnectionAspect.kt
+++ b/src/main/java/uk/gov/justice/hmpps/prison/aop/connectionproxy/OracleConnectionAspect.kt
@@ -6,6 +6,7 @@ import org.slf4j.LoggerFactory
 import org.slf4j.MDC
 import org.springframework.context.annotation.Profile
 import org.springframework.stereotype.Component
+import org.springframework.transaction.TransactionException
 import uk.gov.justice.hmpps.kotlin.auth.AuthSource
 import uk.gov.justice.hmpps.kotlin.auth.AuthSource.NOMIS
 import uk.gov.justice.hmpps.kotlin.auth.HmppsAuthenticationHolder
@@ -15,6 +16,7 @@ import uk.gov.justice.hmpps.prison.util.MdcUtility.REQUEST_URI
 import uk.gov.justice.hmpps.prison.util.MdcUtility.SUPPRESS_XTAG_EVENTS
 import uk.gov.justice.hmpps.prison.util.MdcUtility.USER_ID_HEADER
 import uk.gov.justice.hmpps.prison.web.config.RoutingDataSource
+import java.lang.Exception
 import java.sql.Connection
 import java.sql.SQLException
 import java.util.Properties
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/justice/hmpps/prison/aop/connectionproxy/OracleConnectionAspect.kt
```diff
@@ -62,8 +64,14 @@ class OracleConnectionAspect(
     }
 
     return (this.unwrap(Connection::class.java) as OracleConnection)
-      .also { oracleConnection ->
-        oracleConnection.openProxySession(OracleConnection.PROXYTYPE_USER_NAME, info)
+      .apply {
```
petergphillips, line 67: didn't like this as an `also` since it is not a side effect

src/main/java/uk/gov/justice/hmpps/prison/aop/connectionproxy/OracleConnectionAspect.kt
```diff
@@ -108,3 +116,5 @@ class OracleConnectionAspect(
 
 @Throws(SQLException::class)
 internal fun Connection.run(sql: String) = prepareStatement(sql).use { ps -> ps.execute() }
+
+class ProxyConnectionTransactionException(message: String, e: Exception) : TransactionException(message, e)
```
petergphillips, line 120: throwing `TransactionException` here so that it is rethrown as is and [not wrapped](https://github.com/spring-projects/spring-framework/blob/30dc86810e2a14357bce4f2e5a6fd2faa0c0dac2/spring-orm/src/main/java/org/springframework/orm/jpa/JpaTransactionManager.java#L460-L463) in a `CannotCreateTransactionException` that is always worrying when we see it in the logs.

src/main/java/uk/gov/justice/hmpps/prison/aop/connectionproxy/OracleConnectionAspect.kt
```diff
@@ -62,8 +64,14 @@ class OracleConnectionAspect(
...
 
     return (this.unwrap(Connection::class.java) as OracleConnection)
-      .also { oracleConnection ->
-        oracleConnection.openProxySession(OracleConnection.PROXYTYPE_USER_NAME, info)
+      .apply {
+        try {
+          openProxySession(OracleConnection.PROXYTYPE_USER_NAME, info)
+        } catch (e: SQLException) {
+          if (e.errorCode == 1017) throw ProxyConnectionTransactionException("Could not open proxy session as $currentUsername - account doesn't support proxy connections", e)
+          if (e.errorCode == 28000) throw ProxyConnectionTransactionException("Could not open proxy session as $currentUsername - account is locked", e)
```
petergphillips, line 72: this comes through as
```
Caused by: java.sql.SQLException: ORA-28000: the account is locked
```
which means that the Oracle account is locked so we can't proxy as the user

andymarke (approved):
lgtm

PR #2289, opened 2024-10-23
SDIT-2123: 🔊 Change logging to info level and alter logging some logging to debug instead

petergphillips:


applicationinsights.dev.json
```diff
@@ -7,7 +7,7 @@
   },
   "instrumentation": {
     "logging": {
-      "level": "DEBUG"
+      "level": "INFO"
```
petergphillips, line 10: default level now `INFO`

src/main/java/uk/gov/justice/hmpps/prison/api/resource/EducationResource.java
```diff
@@ -49,7 +49,7 @@ public Page<Education> getPrisonerEducations(
         @RequestParam(value = "page", defaultValue = "0", required = false) @Parameter(description = "The page number of the paged results") final Integer page,
         @RequestParam(value = "size", defaultValue = "10", required = false) @Parameter(description = "Requested limit to number of results returned.") final Integer size
     ) {
-        log.info("get prisoner educations for offenderNo: {}", offenderNo);
+        log.debug("get prisoner educations for offenderNo: {}", offenderNo);
```
petergphillips, line 52: most of these are unnecessary anyway since they'll be in `AppRequests`

andymarke (approved):
lgtm

PR #2301, opened 2024-11-27
SAA-2217: Handle null in weekly_absence_limit column

munslowl:


src/main/java/uk/gov/justice/hmpps/prison/api/model/AgencyPrisonerPayProfile.kt
```diff
@@ -24,7 +24,7 @@ data class AgencyPrisonerPayProfile(
   val payFrequency: Int?,
 
   @Schema(description = "The number of absences that are acceptable within one week", example = "5")
-  val weeklyAbsenceLimit: Int?,
+  val weeklyAbsenceLimit: Int? = null,
```
mikehalmamoj, line 27: I've checked the prod data and noticed that we also have nulls in minHalfDayRate, maxHalfDayRate, maxPieceWorkRate, maxBonusRate and defaultPayBandCode - would you mind making these nullable too while you're in here? Thanks.

munslowl: Sorted @mikehalmamoj

mikehalmamoj (approved):

PR #2304, opened 2024-12-03
ADJST-884 Retrieve the history of imprisonment status for an offender

ldlharper:


src/main/java/uk/gov/justice/hmpps/prison/api/resource/ImprisonmentStatusHistoryResource.kt
```diff
@@ -0,0 +1,62 @@
...
+    ApiResponse(
+      responseCode = "404",
+      description = "Requested resource not found.",
+      content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+    ),
+  )
+  @Operation(summary = "Returns the details of all the historic imprisonment statuses for an offender.")
+  @GetMapping("/{offenderNo}")
+  @VerifyOffenderAccess(overrideRoles = ["GLOBAL_SEARCH", "VIEW_PRISONER_DATA"])
+  @SlowReportQuery
```
ldlharper, line 53: Is this required? It was added to another similar endpoint I'd created.

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/ImprisonmentStatusHistoryResourceImplTest.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.List;
+
+public class ImprisonmentStatusHistoryResourceImplTest extends ResourceTest {
+    @Test
+    public void courtDateResults() {
+        final var requestEntity = createHttpEntityWithBearerAuthorisation("ITAG_USER", List.of("ROLE_VIEW_PRISON_DATA"), null);
+
+        final var responseEntity = testRestTemplate
+            .exchange(
+                "/api/inmate-status-history/A1180HL",
```
ldlharper, line 16: Is it okay to use existing test data here?

petergphillips: yes fine, all new files should be in kotlin though

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/ImprisonmentStatusHistoryResourceImplTest.java
```diff
@@ -0,0 +1,24 @@
...
+import org.springframework.http.HttpStatus;
+
+import java.util.List;
+
+public class ImprisonmentStatusHistoryResourceImplTest extends ResourceTest {
+    @Test
+    public void courtDateResults() {
+        final var requestEntity = createHttpEntityWithBearerAuthorisation("ITAG_USER", List.of("ROLE_VIEW_PRISON_DATA"), null);
+
+        final var responseEntity = testRestTemplate
```
petergphillips, line 14: new files should use the web client instead

src/main/java/uk/gov/justice/hmpps/prison/api/resource/ImprisonmentStatusHistoryResource.kt
petergphillips: normally only required for endpoints that are slow / computationally expensive which this one doesn't look like it will be

petergphillips (approved):
LGTM

petergphillips (approved):
LGTM

PR #2311, opened 2024-12-13
CDPS-1074 Add edit nationality endpoint

scottrowley:


src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerProfileUpdateService.kt
```diff
@@ -44,7 +51,61 @@ class PrisonerProfileUpdateService(
...
 
-  private fun Optional<Offender>.orElseThrowNotFound(message: String, prisonerNumber: String) =
+  @Transactional
+  fun updateNationalityOfCurrentBooking(prisonerNumber: String, nationality: String?) {
+    val profileType: ProfileType = profileTypeRepository.nationalityProfile().getOrThrow()
+    val profileCode: ProfileCode? =
+      nationality?.uppercase()?.let { profileCodeRepository.profile(profileType, it).getOrThrow() }
+
+    try {
+      val currentBooking = currentBooking(prisonerNumber)
```
brightonsbox, line 61: Meh.. I think 'latest' over 'current' would be preferred since if the prisoner has been released, the latest booking is inactive

src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerProfileUpdateService.kt
```diff
@@ -44,7 +51,61 @@ class PrisonerProfileUpdateService(
...
+          currentBooking.profileDetails.add(
+            OffenderProfileDetail.builder()
+              .id(OffenderProfileDetail.PK(currentBooking, profileType, 1))
+              .caseloadType("INST")
+              .code(profileCode)
+              .listSequence(profileType.listSequence)
+              .build(),
+          )
+        }
+      }
```
brightonsbox, line 80: Could this be simplified by doing:

```
if (currentNationality != null) {
    currentBooking.profileDetails.remove(currentNationality)
}

if (nationality != null) {
    currentBooking.profileDetails.add(
            OffenderProfileDetail.builder()
              .id(OffenderProfileDetail.PK(currentBooking, profileType, 1))
              .caseloadType("INST")
              .code(profileCode)
              .listSequence(profileType.listSequence)
              .build(),
          )
}
```

i.e. not worrying about setting the code on the current row, just get rid of the row and add a new one?

brightonsbox: No... this causes issues with JPA thinking there's a conflict: `a different object with the same identifier value was already associated with the session`, so ignore this

src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerProfileUpdateService.kt
```diff
@@ -44,7 +51,61 @@ class PrisonerProfileUpdateService(
...
+    } catch (e: CannotAcquireLockException) {
+      throw processLockError(e, prisonerNumber)
+    }
+  }
+
+  private fun currentBooking(prisonerNumber: String) =
+    (
+      offenderRepository.findLinkedToLatestBookingForUpdate(prisonerNumber)
```
brightonsbox, line 88: As discussed, I think we need to do the locking on the actual OFFENDER_PROFILE_DETAILS (NAT) record so that we get the exception thrown if someone's trying to edit nationality in NOMIS at the same time.  I think this should do it: https://github.com/ministryofjustice/prison-api/pull/2313

brightonsbox (approved):
LGTM

PR #2314, opened 2024-12-18
CDPS-1075 Add endpoint to allow retrieval of both reference and profile codes

scottrowley:


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/ReferenceDataResourceTest.kt
```diff
@@ -0,0 +1,372 @@
+package uk.gov.justice.hmpps.prison.api.resource.impl
```
brightonsbox, line 1: 👍 for Kotlin conversion

brightonsbox (approved):
LGTM

PR #2315, opened 2024-12-18
PRC-321: Add town, city and country code to address

james-baker-moj:
This supports copying an address from a prisoner to a contact for hmpps-contacts-ui.


steverendell (approved):
lgtm

PR #2316, opened 2024-12-20
CDPS-1069 Add edit religion endpoint

scottrowley:


src/main/java/uk/gov/justice/hmpps/prison/api/model/UpdateReligion.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import io.swagger.v3.oas.annotations.media.Schema
+import jakarta.validation.constraints.Size
+
+@Schema(description = "Update to prisoner religion")
+data class UpdateReligion(
+  @Schema(description = "Religion", example = "ZORO", required = true, nullable = true)
+  val religion: String?,
+  @Schema(description = "Comment", example = "Some information", required = false)
+  @Size(max = 4000, message = "Comment text must be a maximum of 4000 characters")
```
petergphillips, line 11: IIRC this is normally 4000 bytes, not characters.  If you pass through 4000 emojis for instance then you'll get an oracle error on save.  Worth therefore validating that the comment is not too long, or alternatively truncating to 4000 bytes.  There is existing code for this around comments in case notes etc.

src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerProfileUpdateService.kt
```diff
@@ -57,25 +67,56 @@ class PrisonerProfileUpdateService(
...
+      if (profileCode == null && latestProfileEntryOfType != null) {
+        booking.profileDetails.remove(latestProfileEntryOfType)
       } else {
-        if (latestNationality != null) {
-          latestNationality.setProfileCode(profileCode)
+        if (latestProfileEntryOfType != null) {
+          latestProfileEntryOfType.setProfileCode(profileCode)
         } else {
-          latestBooking.profileDetails.add(
+          booking.profileDetails.add(
```
petergphillips, line 117: not sure this is wise.  There is already a `OFFENDER_BELIEFS_T1` trigger on the `OFFENDER_BELIEFS` table that will update or insert into `offender_profile_details` table.

scottrowley: Thanks, I hadn't realised there was a trigger for this. I've stopped doing it for religion, but have left it in place for updating nationality as I don't think there are any triggers around that

src/main/java/uk/gov/justice/hmpps/prison/api/model/UpdateReligion.kt
scottrowley: Thanks, I'd copied this from elsewhere but I've looked at the case notes and added similar validation

src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderBeliefRepository.kt
```diff
@@ -13,7 +15,7 @@ interface OffenderBeliefRepository : CrudRepository<OffenderBelief, Long> {
     FROM OffenderBelief belief
     WHERE belief.booking.offender.nomsId = :prisonerNumber
     AND (:bookingId IS NULL OR belief.booking.bookingId = :bookingId)
-    ORDER BY belief.startDate DESC
+    ORDER BY belief.startDate DESC, belief.createDatetime DESC
```
scottrowley, line 18: Added as part of mapping startDate from LocalDateTime to LocalDate - it wasn't correctly ordered when making multiple updates on the same day.

src/test/java/uk/gov/justice/hmpps/prison/service/MovementsServiceImplTest.java
```diff
@@ -706,7 +706,7 @@ public void invalidDateRange() {
             @Test
             public void dateRangeIsGreaterThan24Hours() {
                 final var from = LocalDateTime.now();
-                final var to = LocalDateTime.now().plusDays(1);
+                final var to = from.plusDays(1).plusSeconds(1);
```
scottrowley, line 709: This test was flaky for me - this seems to fix it.

brightonsbox (approved):
LGTM

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderResourceImplIntTest_updateReligion.kt
```diff
@@ -0,0 +1,274 @@
...
+        .header("Content-Type", APPLICATION_JSON_VALUE)
+        .bodyValue(VALID_RELIGION_UPDATE)
+        .exchange()
+        .expectStatus().isNotFound
+        .expectBody().jsonPath("userMessage")
+        .isEqualTo("Staff user account with provided username not found")
+    }
+
+    @Test
+    fun `returns status 423 (locked) when database row lock times out`() {
```
brightonsbox, line 220: I think this can be removed? We now don't expect that the locking condition will happen

brightonsbox (approved):
LGTM

PR #2322, opened 2025-01-08
SDIT-2391: 🐛 Correct swagger and throw 409 for duplicate

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/api/resource/FinanceController.java
```diff
@@ -51,7 +51,7 @@ public class FinanceController {
     @Tag(name = "integration-api")
     @Tag(name = "unilink")
     @ApiResponses({
-            @ApiResponse(responseCode = "201", description = "Transaction Created", content = {@Content(mediaType = "application/json", schema = @Schema(implementation = TransferTransaction.class))}),
+            @ApiResponse(responseCode = "200", description = "Transaction Created"),
```
petergphillips, line 54: removed so will use default instead

Mjwillis (approved):
LGTM

PR #2328, opened 2025-01-21
MAP-1935 removed unused endpoint

GurnankCheema:
the GET /api/agencies/{agencyId}/locations/type/{type} is no longer being called by any app


Mjwillis (approved):
LGTM

PR #2334, opened 2025-01-23
SDIT-2406 Include last movement time in the prisoner search API

mikehalmamoj:


src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerSearchService.kt
```diff
@@ -56,8 +57,9 @@ class PrisonerSearchService(
           indeterminateSentence = it.sentenceTerms?.any { st -> st.lifeSentence && it.bookingId == st.bookingId },
           aliases = it.aliases,
           status = it.status,
-          lastMovementTypeCode = it.lastMovementTypeCode,
-          lastMovementReasonCode = it.lastMovementReasonCode,
+          lastMovementTypeCode = booking?.lastMovement?.getOrNull()?.movementType?.code,
+          lastMovementReasonCode = booking?.lastMovement?.getOrNull()?.movementReason?.code,
+          lastMovementTime = booking?.lastMovement?.getOrNull()?.movementTime,
```
mikehalmamoj, line 62: We were previously deriving the last movement details from the `STATUS_REASON` column. As we now pick up the last movement for its time, it makes more sense to also get the type and reason directly from the movement too.

petergphillips: I'd be very careful about this though - we've had instances in the past where the movement doesn't reflect the latest code on the booking

mikehalmamoj: Good point. I spot checked a few and it seemed to be a "rule" but I've checked again and there are many discrepancies.

But why were we setting last movement type/reason from the booking's STATUS_REASON column anyway? If that's not always the last movement why do we  say it is? If anything this change fixes a bug?

petergphillips (approved):
LGTM

PR #2335, opened 2025-01-24
SDIT-2406 - allow creation of prisoners with same name when required.

andymarke:
Expected to only be used for testing when needing prisoners to do a merge.


karenmillermoj (approved):
lgtm

PR #2336, opened 2025-01-24
CDPS-1107: New endpoints to support editing military history

davidatmoj:


src/main/java/uk/gov/justice/hmpps/prison/service/OffenderMilitaryRecordService.kt
```diff
@@ -0,0 +1,108 @@
...
+  fun createMilitaryRecord(offenderNo: String, militaryRecord: MilitaryRecord) {
+    val booking = getLatestOffenderBooking(offenderNo)
+    val nextMilitarySeq = booking.militaryRecords.size + 1
+
+    val offenderMilitaryRecord = OffenderMilitaryRecord.builder()
+      .bookingAndSequence(OffenderMilitaryRecord.BookingAndSequence(booking, nextMilitarySeq))
+      .warZone(militaryRecord.warZoneCode?.let { WarZone(it, militaryRecord.warZoneDescription) })
+      .startDate(militaryRecord.startDate)
+      .endDate(militaryRecord.endDate)
+      .militaryDischarge(militaryRecord.militaryDischargeCode?.let { MilitaryDischarge(it, militaryRecord.militaryDischargeDescription) })
```
andymarke, line 45: I think it would make more sense to lookup the reference data from the code use a repository rather than just setting blindly - since the lookup implicitly validates the code, the you drop the description from the DTO

davidatmoj: Thanks @andymarke that makes sense. Will update.

andymarke (dismissed):
lgtm

src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
```diff
@@ -760,18 +763,49 @@ public ResponseEntity<List<OffenderTransactionHistoryDto>> getTransactionsHistor
...
-            return bookingService.getMilitaryRecords(booking.getBookingId());
-        } catch (EntityNotFoundException e) {
-            // rethrow against the offender number rather than the booking id
-            throw EntityNotFoundException.withId(offenderNo);
-        }
+        return offenderMilitaryRecordService.getMilitaryRecords(offenderNo);
+    }
+
+    @ApiResponses({
+        @ApiResponse(responseCode = "200", description = "OK"),
```
brightonsbox, line 775: Should this also be a 204?

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderMilitaryRecordService.kt
```diff
@@ -0,0 +1,161 @@
...
+          )
+        },
+      )
+      .build()
+
+    repository.save(offenderMilitaryRecord)
+  }
+
+  fun updateMilitaryRecord(militaryRecord: MilitaryRecord) {
+    val record = repository.findByBookingIdAndMilitarySeqWithLock(militaryRecord.bookingId, militaryRecord.militarySeq)
```
brightonsbox, line 96: I think this needs surrounding with a `try... catch` to catch the `CannotAcquireLockException` and throw the `DatabaseRowLockedException` which gets handled with a 423 status code (see https://github.com/ministryofjustice/prison-api/blob/main/src/main/java/uk/gov/justice/hmpps/prison/service/PrisonerProfileUpdateService.kt#L52)

src/main/java/uk/gov/justice/hmpps/prison/api/resource/BookingResource.java
```diff
@@ -567,10 +566,10 @@ public void addPersonalCareNeed(
         @ApiResponse(responseCode = "404", description = "Requested resource not found.", content = {@Content(mediaType = "application/json", schema = @Schema(implementation = ErrorResponse.class))}),
         @ApiResponse(responseCode = "500", description = "Unrecoverable error occurred whilst processing request.", content = {@Content(mediaType = "application/json", schema = @Schema(implementation = ErrorResponse.class))})})
     @Operation(summary = "Military Records", description = "Requires booking to be in caseload, or role VIEW_PRISONER_DATA")
-    @ProgrammaticAuthorisation("Checked in service for legacy reasons")
```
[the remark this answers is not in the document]
davidatmoj: It was basically just deferring the auth check to BookingService.getMilitaryRecords which has now been deleted
image

src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
davidatmoj: Probably should be 201, so will update that.

src/main/java/uk/gov/justice/hmpps/prison/service/OffenderMilitaryRecordService.kt
brightonsbox: At some point we need to propagate the 423 through the proxy and handle the database lock in the profile...

davidatmoj: I'm moving on to the proxy next, then using it all in the profile, so I will look into that as part of that work.

brightonsbox (approved):
LGTM

PR #2341, opened 2025-01-28
ADJST-1010 Add new adjustments (TCA & TSA)

dave-virden:
Adds the two new adjustments, time spent in custody abroad and time spent as an appeal applicant not to count, that have been added to DPS to preserve the two way sync.


andymarke (approved):
Lgtm

PR #2347, opened 2025-02-10
CDPS-1140: Military history

davidatmoj:
New models for CreateMilitaryRecord and UpdateMilitaryRecord


src/main/java/uk/gov/justice/hmpps/prison/api/model/UpdateMilitaryRecord.kt
```diff
@@ -0,0 +1,94 @@
...
+import java.time.LocalDate
+
+@JsonInclude(Include.NON_NULL)
+@Schema(description = "Update military record.")
+data class UpdateMilitaryRecord(
+  @Schema(
+    description = "Sequence number of the military record within the booking.",
+    example = "1",
+  )
+  val militarySeq: Int,
```
brightonsbox, line 15: Would it not be better to add this to the URL instead to avoid having to (almost) duplicate the request object?

So you would have :
`POST /{offenderNo}/military-records` for creation
and `PUT /{offenderNo}/military-records/{ militarySeq }` for update

davidatmoj: Yeah, that does make sense

brightonsbox (approved):
LGTM

PR #2349, opened 2025-02-10
SDIT-2562: 👽️ Attempt to unpin hibernate version

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/repository/converter/YesNoToBooleanConverter.java
```diff
@@ -1,22 +0,0 @@
-package uk.gov.justice.hmpps.prison.repository.converter;
```
petergphillips, line 1: use the hibernate one instead

PR #2354, opened 2025-02-17
SDIT-2564 Separate movement reason code into separate field to avoid using value from reference_codes table

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/model/ExternalMovement.java
```diff
@@ -140,13 +141,18 @@ public static class PK implements Serializable {
     })
     private City fromCity;
 
+    // TODO SDIT-2595 Change mapping to use MOVEMENT_REASONS table (not REFERENCE_CODES table)
```
karenmillermoj, line 144: Have raised a SDIT-2595 and added to backlog to fix this issue

andymarke (approved):

PR #2360, opened 2025-02-21
CDPS-1125 Update distinguishing marks response format

scottrowley:


src/main/java/uk/gov/justice/hmpps/prison/service/DistinguishingMarkService.kt
```diff
@@ -121,13 +122,47 @@ class DistinguishingMarkService(
     imageRepository.save(newImage)
   }
 
+  private fun OffenderIdentifyingMark.toDistinguishingMark(): DistinguishingMark {
+    val latestImageId = images.stream()
```
davidatmoj, line 126: You don't need to use Java streams as this is a Kotlin extension (even though the target is Java)
e.g.
```
    val latestImageId = images
        .filter { it.isActive }
        .maxByOrNull { it.id }?.id ?: -1L
```

davidatmoj (approved):

brightonsbox (approved):
LGTM

PR #2369, opened 2025-03-06
CDPS-1274: Update working name alias

brightonsbox:


src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java
```diff
@@ -1048,4 +1052,30 @@ public void updateSmokerStatus(
...
         );
     }
+
+    @ApiResponses({
+        @ApiResponse(responseCode = "204", description = "The prisoner's working name has been updated."),
+        @ApiResponse(responseCode = "304", description = "The prisoner's working name has not been changed (the supplied name was the same as the current working name)."),
+        @ApiResponse(responseCode = "400", description = "Invalid request.", content = {@Content(mediaType = "application/json", schema = @Schema(implementation = ErrorResponse.class))}),
+        @ApiResponse(responseCode = "404", description = "Requested resource not found.", content = {@Content(mediaType = "application/json", schema = @Schema(implementation = ErrorResponse.class))}),
+        @ApiResponse(responseCode = "500", description = "Unrecoverable error occurred whilst processing request.", content = {@Content(mediaType = "application/json", schema = @Schema(implementation = ErrorResponse.class))})})
+    @Operation(summary = "Update the prisoner's working name. This will generate a new alias unless `forceAliasCreation` " +
```
davidatmoj, line 1062: Being picky, but if the default for `forceAliasCreation` is `false`, might make more sense to reword this as something like
"This will generate a new alias if `forceAliasCreation` is `true` (default: `false`), otherwise only changes to the surname will cause a new alias to be created."

davidatmoj (approved):
LGTM

PR #2378, opened 2025-03-13
HMAI-303 - Add tags to endpoints used by HMPPS Integration API

tomwatkins1994:
This PR is to add tags to all endpoints currently used in the HMPPS Integration API as the tags are missing from many of the endpoints that we are using. 

This also reverts a temporary permission we added in #2361 in order to create prisoners in dev for the Milsike prison


petergphillips (approved):
LGTM

PR #2383, opened 2025-03-14
CAS-1509 - Add local data for new offender (required by CAS team)

gregkhawkins:
**Context**
* As a team member in `Community Accommodation Services (CAS)` our API is a consumer of `Prison API`
* We were getting problems when locally our service calls `GET /api/offenders/A1237AI`
* This was because this `A1237AI` is an offender number that is associated with a user (that exists in `AP and Delius` service) that we hand picked to help us test our new `User with No Oasys data` case on our side

**PR includes:**
* New migration data required for the offender `A1237AI` on our side to start behaving so we get back a 200 from your `GET /api/offenders/A1237AI` endpoint


petergphillips (approved):
LGTM

PR #2390, opened 2025-03-24
MAP-2303 only return a distinct list of locations associated with activities - Currently its returning duplicates

Mjwillis:


src/main/java/uk/gov/justice/hmpps/prison/service/AgencyService.java
```diff
@@ -255,7 +255,7 @@ public List<Location> getAgencyEventLocations(final String agencyId, final Strin
 
         final var rawLocations = agencyRepository.getAgencyLocations(agencyId, allEventLocationUsages, orderBy, order);
 
-        return LocationProcessor.processLocations(rawLocations);
+        return LocationProcessor.processLocations(rawLocations).stream().distinct().toList();
```
Mjwillis, line 258: Because the processLocation removes fields there are now duplicates for locations with multiple activity usages.  This is to just make sure the list is distinct

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/AgencyResourceTest.kt
```diff
@@ -736,4 +738,20 @@ class AgencyResourceTest : ResourceTest() {
       AgencyEstablishmentType.builder().code("IM").description("Closed Young Offender Institute (Male)").build(),
     )
   }
+
+  @Test
+  fun testGetEventLocationsForAPrison() {
+    val token = authTokenHelper.getToken(AuthToken.REF_DATA_MAINTAINER)
```
Mjwillis, line 744: There wasn't even a test for this endpoint, I've added a very simple one, ideally it could do with more.

petergphillips (approved):
LGTM

PR #2394, opened 2025-03-26
CDPS-1320: Add an endpoint to get list of alias details

brightonsbox:
I'm aware that there are existing ways of getting most of this data, but this provides an efficient query to get everything that we need to pass to the CPR proxy API.


mtac50 (approved):
LGTM

PR #2408, opened 2025-04-24
BAVL-778: Add the appointment event status to the response

timharrison-moj:
The endpoint `GET /api/schedules/{agencyId}/appointments` returns both completed, scheduled and cancelled appointments from NOMIS. However, it does not indicate which appointments are cancelled. This PR adds the event_status (COMP, CANC, SCH, EXP) to the response so that consumers could determine which are cancelled. BVLS needs to know this for prisons where A&A is not yet rolled out, so that bookings do not clash with other prison appointments in the video rooms. 

This endpoint looks to be being used by 3 clients in production:

* hmpps-book-a-video-link-api (my team's product)
* hmpps-video-daily-schedule (my teams's product)
* whereabouts-api (i can test this in dev to ensure the whereabouts view appointments forms still work)


petergphillips (approved):
LGTM

PR #2420, opened 2025-05-22
SDIT-2737: 👷 Switch to github actions

petergphillips:


.github/workflows/gradle_verify.yml
```diff
@@ -0,0 +1,49 @@
...
+        default: '21'
+      java-options:
+        type: string
+        default: ''
+        required: false
+      gradle-command:
+        type: string
+        default: './gradlew check'
+        required: false
+      artifact-name:
```
petergphillips, line 17: temporary local version so can specify the artifact name so doesn't conflict

steverendell (approved):
lgtm

PR #2423, opened 2025-05-27
SDIT-2774: 🔇 Prevent flyway logging database params

petergphillips:


petergphillips:
Not needed as flyway not run

PR #2438, opened 2025-06-20
CDPS-1361 Allow identifier update when value has not changed

scottrowley:
This allows updating the issuedAuthorityText without changing the identifier value. This was previously being rejected as a duplicate value.


src/test/java/uk/gov/justice/hmpps/prison/service/OffenderIdentityServiceTest.kt
```diff
@@ -81,179 +91,344 @@ class OffenderIdentityServiceTest {
...
-    val identifiers = offenderIdentifierService.getOffenderIdentifiers("ABC123", false)
-    assertThat(identifiers).hasSize(1)
-    assertThat(identifiers[0].offenderId).isEqualTo(123)
-  }
+      val identifiers = offenderIdentifierService.getOffenderIdentifiers("ABC123", false)
+      assertThat(identifiers).hasSize(1)
+      assertThat(identifiers[0].offenderId).isEqualTo(123)
+    }
 
-  @Test
```
scottrowley, line 95: Most of these changes are restructuring to use nested tests, but I did add a tests against duplicate values.

brightonsbox (approved):
LGTM

PR #2457, opened 2025-07-25
SDIT-2671: ♻️  Fix offender identifier model and convert to kotlin

petergphillips:


andymarke (approved):

petergphillips:
not going to do this for now

PR #2490, opened 2025-08-08
SDIT-1997 Add row locks to POST api/offender-dates/{bookingId}

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/repository/BookingRepository.java
```diff
@@ -318,6 +318,26 @@ public List<ScheduledEvent> getBookingActivities(final Collection<Long> bookingI
         return activities.stream().map(ScheduledEventDto::toScheduledEvent).collect(Collectors.toList());
     }
 
+    public void lockReleaseDetails(final Long bookingId) {
```
karenmillermoj, line 321: this is a copy of lockAttendance immediately below

src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/OffenderDatesResourceTest.kt
```diff
@@ -0,0 +1,307 @@
+package uk.gov.justice.hmpps.prison.api.resource.impl
```
karenmillermoj, line 1: Tests have not changed, just refactored to kotlin

PR #2514, opened 2025-08-18
SDIT-2943 Improve performance of get location appointments by removing TRUNC

karenmillermoj:


src/main/java/uk/gov/justice/hmpps/prison/repository/sql/ScheduleRepositorySql.kt
```diff
@@ -201,8 +201,7 @@ enum class ScheduleRepositorySql(val sql: String) {
         WHERE OIS.TO_INTERNAL_LOCATION_ID = :locationId
         AND OIS.EVENT_TYPE = 'APP'
         AND OIS.EVENT_STATUS <> 'CANC'
-        AND OIS.EVENT_DATE >= TRUNC(COALESCE(:fromDate, OIS.EVENT_DATE))
-        AND TRUNC(OIS.EVENT_DATE) <= COALESCE(:toDate, OIS.EVENT_DATE)
+        AND OIS.EVENT_DATE between COALESCE(:fromDate, OIS.EVENT_DATE) AND COALESCE(:toDate, OIS.EVENT_DATE)
```
karenmillermoj, line 204: event_date does not have any time portions set so **trunc**ing is unnecessary.  This was causing the index to be ignored resulting in a lengthier query

PR #2528, opened 2025-09-09
CDPS-1574: Full person endpoint for use with prisoner profile

Ewan-Donovan:
Took a simpler approach because I was losing time to it.

https://dsdmoj.atlassian.net/browse/CDPS-1574

related pr: https://github.com/ministryofjustice/hmpps-person-integration-api/pull/87


src/main/java/uk/gov/justice/hmpps/prison/api/model/FullPersonDto.kt
```diff
@@ -0,0 +1,31 @@
+package uk.gov.justice.hmpps.prison.api.model
+
+import com.fasterxml.jackson.annotation.JsonInclude
+import io.swagger.v3.oas.annotations.media.Schema
+
+@Schema(description = "Person - DTO for use in returning all person data for use in the Core Person Record proxy")
```
petergphillips, line 6: having a class that references Person is really confusing in Prison API.  A person in NOMIS terms means a contact, it **never** means a prisoner

petergphillips: would also be good explaining why there is a core person endpoint in this service e.g.
> This endpoint is a temporary endpoint providing prisoner details in a similar format to what the Core Person Record service is expected to provide.  It is only to be used by Prisoner Profile.

PR #2541, opened 2025-09-26
SDIT-3049 Add military flag and previous prison to prisoner-search

steverendell:
Fix transfer via TAP and finalise


petergphillips (approved):
LGTM

PR #2661, opened 2026-05-12
SDIT-3684 Tidy after cucumber test removal

karenmillermoj:


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/AuthTokenHelper.kt
```diff
@@ -1,4 +1,4 @@
-package uk.gov.justice.hmpps.prison.executablespecification.steps
+package uk.gov.justice.hmpps.prison.api.resource.impl
```
karenmillermoj, line 1: Moved package as only those in impl package are now using this file

petergphillips (approved):
LGTM

PR #2662, opened 2026-05-13
IR-1613 Refactor prisoner role handling logic

Mjwillis:
### Description of Changes

- Replaced enum-based role handling with string-based validation.
- Added checks to ensure roles exist before assignment.
- Updated related models and adjusted tests to match the new logic.


petergphillips (approved):
LGTM

PR #2670, opened 2026-05-18
MO-1966 Move to single source of truth for recall status.

dave-virden:
Removes an old and out of date list maintained in the RecallCalc and uses SentenceTypeRecallType instead.


src/test/java/uk/gov/justice/hmpps/prison/api/model/RecallCalcTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+import java.util.Collections;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class RecallCalcTest {
```
petergphillips: all new code should be written in kotlin

petergphillips (approved):
LGTM

PR #2677, opened 2026-05-18
SDIT-3796: 🔥 Remove unused GET /api/prisoners/prisoner-numbers/active

petergphillips:
- **SDIT-3796: 🔥 Remove unused GET /api/prisoners/prisoner-numbers/active**


src/main/java/uk/gov/justice/hmpps/prison/api/resource/PrisonerResource.kt
```diff
@@ -141,47 +135,6 @@ class PrisonerResource(
     )
   }
 
-  @GetMapping("/prisoner-numbers/active")
```
petergphillips, line 144: only used by prisoner search and moved to nomis-prisoner-api anyway

karenmillermoj (approved):
lgtm

PR #2692, opened 2026-05-25
SDIT-3685: ♻️ Convert tests to kotlin

petergphillips:


src/test/java/uk/gov/justice/hmpps/prison/api/resource/impl/CorePersonCommunicationNeedsResourceTest.kt
```diff
@@ -16,15 +16,15 @@ class CorePersonCommunicationNeedsResourceTest : ResourceTest() {
     @Test
     fun `returns 401 without an auth token`() {
       webTestClient.get()
-        .uri("/api/offenders/A1234AA/core-person-record/communication-needs")
+        .uri("/api/offenders/A1234AB/core-person-record/communication-needs")
```
petergphillips, line 19: this test was conflicting with a repository one so switched to a different booking

mikehalmamoj (approved):

PR #2707, opened 2026-07-01
MAPB-580 Remove `/agencies/{agencyId}/locations/groups` from Prison API

Mjwillis:


src/main/java/uk/gov/justice/hmpps/prison/repository/LocationRepository.java
```diff
@@ -37,37 +33,17 @@ public Optional<Location> findLocation(final long locationId, final StatusFilter
     }
 
 
-    public Optional<Location> findLocation(final long locationId, final String username) {
+    public void findLocation(final long locationId, final String username) {
```
Mjwillis, line 36: Not sure we really need this anymore but kept for now and remove warning from file as result never used

mikehalmamoj (approved):

PR #2719, opened 2026-07-31
SDIT-4084: ⚡️ Remove bookingId as never used for religion history

petergphillips:


src/main/java/uk/gov/justice/hmpps/prison/repository/jpa/repository/OffenderBeliefRepository.kt
```diff
@@ -12,13 +12,12 @@ interface OffenderBeliefRepository : CrudRepository<OffenderBelief, Long> {
     """
     SELECT
     belief
-    FROM OffenderBelief belief
+    FROM OffenderBelief belief join fetch belief.beliefCode
```
petergphillips, line 15: I've added the same performance improvement from NOMIS Prisoner API here to avoid extra selects for profile codes

mikehalmamoj (approved):

PR #2723, opened 2026-08-24
MAPA-316 Correct the cell swap comment about location type

Mjwillis:
**Comment-only. No behaviour change.**

Follow-up to #2722. The comment I added there says a cell swap location *"is a WING rather than a CELL"*. That's true of the test seed data (`R__3_2__AGENCY_INTERNAL_LOCATIONS.sql:110`) but not of the live estate.

Sweeping all 127 prisons in production — via `GET /api/locations/code/{prisonId}-CSWAP`, which uses the same `findOneByDescription` lookup the cell move performs — these locations turn out to be recorded as **WING, HOLD and AREA**.

The gate has always been correct for all of them, because `isCellSwap()` never looked at `locationType`:

```java
public boolean isCellSwap() {
    return !isCertifiedFlag() && isActive() && parentLocation == null
            && locationCode != null && locationCode.equals("CSWAP");
}
```

So nothing changes except the comment — but it's the kind that invites someone to "tighten" the check to `locationType.equals("WING")` later and break a hundred prisons. It now says so explicitly, in `MovementUpdateService` and in the test helper that builds the fixture.

`./gradlew ktlintCheck` and `MovementUpdateServiceTest` (25 tests) pass.

MAPA-316

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/service/MovementUpdateService.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/service/MovementUpdateService.java
@@ -54,9 +54,11 @@ public class MovementUpdateService {
             return transformToCellSwapResult(offenderBooking);
         }
 
-        // A cell swap location is checked first and on its own: it is a WING rather than a CELL and is
-        // deliberately uncapped, so it satisfies neither of the checks below. Short-circuiting here is
-        // what lets this endpoint serve the cell swap that move-to-cell-swap was added for.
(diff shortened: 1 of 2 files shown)


Mjwillis:
Superseded by #2724, which removes the `move-to-cell-swap` endpoint entirely and rewrites this same comment block in the process. The correction here — that `isCellSwap()` ignores `locationType`, and that these locations are recorded as WING, HOLD and AREA across the estate — is carried over verbatim, along with the `cellSwapDestination()` KDoc fix.

Closing so there is one PR to review rather than two that conflict.

PR #2728, opened 2026-09-02
SDIT-3840 Allow sentences to be disabled on release.

andymarke:
This keeps existing behaviour as is; but allows a client to send through a flag for a release to deactivate sentences like NOMIS does.

This currently is would only be used for creating test data given Restricted Patients does use the default on no change

### DIFF
--- a/src/main/java/uk/gov/justice/hmpps/prison/api/model/RequestToReleasePrisoner.java
+++ b/src/main/java/uk/gov/justice/hmpps/prison/api/model/RequestToReleasePrisoner.java
@@ -38,6 +38,10 @@ public class RequestToReleasePrisoner {
     @Default
     private String toLocationCode = "OUT";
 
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/justice/hmpps/prison/api/model/RequestToReleasePrisoner.java
```diff
@@ -38,6 +38,10 @@ public class RequestToReleasePrisoner {
     @Default
     private String toLocationCode = "OUT";
 
+    @Schema(description = "Whether to deactivate sentences upon release, default is false", example = "false")
+    @Default
+    private boolean deactivateSentences = false;
+
```
mikehalmamoj, line 44: Is this Latin?

mikehalmamoj (approved):

